PR #17, opened 2017-11-07
Add doc-view

jojojames:
@Ambrevar Looking for some feedback before I push this.

1. C-j/C-k for next/previous page seems reasonable. I think gj/gk are useful too.
   Do we want to bind those also?

2. C-u, this replaces the universal argument. It should probably be wrapped in the evil variable wants-c-u (or whatever the name is) but what happens if they change that value after they've set this.
   It'd be nice for it to automatically update. Maybe an advice instead? That seems heavy.
   
3. Opinions on RET for image-next-line?

4. +, 0, - for scaling?

5. I used W for fit width which is a movement from evil but since it's doc-view, that won't be used anyway.

6. H and P are kinda similar to W.

7. The Ctrl-* keys are just taken from the emacs version of the binds.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -45,6 +45,7 @@
     debugger
     diff-mode
     dired
+    doc-view
     edebug
     elfeed
     elisps-refs
--- a/evil-doc-view.el
+++ b/evil-doc-view.el
@@ -36,6 +36,8 @@
   (evil-define-key 'motion doc-view-mode-map
     (kbd "C-j") 'doc-view-next-page
     (kbd "C-k") 'doc-view-previous-page
+    (kbd "gj") 'doc-view-next-page
+    (kbd "gk") 'doc-view-previous-page
     (kbd "C-d") 'forward-page
     (kbd "j") 'doc-view-next-line-or-next-page
     (kbd "k") 'doc-view-previous-line-or-previous-page
@@ -44,6 +46,7 @@
     (kbd "J") 'doc-view-goto-page
     (kbd "RET") 'image-next-line
     (kbd "+") 'doc-view-enlarge
+    (kdb "=") 'doc-view-enlarge
     (kbd "0") 'doc-view-scale-reset
     (kbd "-") 'doc-view-shrink
     (kbd "W") 'doc-view-fit-width-to-window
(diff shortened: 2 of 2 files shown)


Ambrevar:
1.
	> C-j/C-k for next/previous page seems reasonable. I think gj/gk are useful too.

	The question of `gj`/`gk` should go to the rationale.
I personally don't like it much because you can't "keep pressed" to run a sequence of the bindings.
I find the many other bindings more efficient: `C-j`/`C-k`, `[`/`]`, other parentheses, maybe `M-j`/`M-k`.
I'd suggest the following hierarchy, from the most precise / most useful to the less:

		j/k > C-j/C-k > [/] > gj/gk

	If we need more:

		j/k > C-j/C-k > M-j/M-k > [/] > {/} > (/) > gj/gk



2.
	> C-u, this replaces the universal argument. It should probably be wrapped in the evil variable wants-c-u (or whatever the name is) but what happens if they change that value after they've set this.
It'd be nice for it to automatically update. Maybe an advice instead? That seems heavy.

	I've faced the same issue with `image` and I haven't taken the time to think of a proper solution.
	This an important issue.

3.
	> Opinions on RET for image-next-line?

	`evil-image` has `image-toggle-animation`.  We can discuss this.
	`evil-pdf` has `image-next-line`.

4.
	> +, 0, - for scaling?

	+/-/0 is obvious.
	Question: should `=` be bound to? On QWERTY, `+` is shifted and `=` is not.
	On the other hand, `=` could also mean 100% zoom like `0`.

5.
	> I used W for fit width which is a movement from evil but since it's doc-view, that won't be used anyway.
	Same in `evil-pdf` and `evil-image`.

6.
	> H and P are kinda similar to W.

	Same in `evil-pdf` and `evil-image`.

7.
	> The Ctrl-* keys are just taken from the emacs version of the binds.
	Bindings that are too specialized can be left to their Emacs state version, I think it's fine.
	Not sure about docview, though, I'll have a lot.

For the moment, the most important is to ensure consistency between those 3 modes:

- image / image+
- docview
- pdftools

For now there seems to be a conflict between image mode and pdftools, see
https://github.com/politza/pdf-tools/issues/324.

pdftools is also a big one.

jojojames:
@Ambrevar Thanks. I added a few things, just made "=" do what the base emacs keymap did.

We can take care of the consistency elsewhere.

As for gj/gk, I think we want to add it at least. It's a common vim command that's also not really ever used by Emacs. If we can bind to it, we should, as it's kind of a 'free keybind'.

Some vim users might not want to chord ever, so even c-j/c-k might be more than gj/gk even if c-j/c-k can be rapidly pressed in sequence.

Issue #27, opened 2017-11-10
Rationale: Jump (go to definition) and display other-window


Ambrevar:
Several modes (occur, pdf, etc.) allow to jump in other window or to display in other window (that is, pop up buffer but don't move the cursor).

In many cases, `RET` should jump to definition, i.e. act just like ~gd~.

Then what about `S-RET` to jump in other window, and `M-RET` to display in other window?

Also `o` or `C-o` have been used in some places I believe.

jojojames:
I like o and C-o when it's available.

S-RET/M-RET also works for me.

Ambrevar:
Actually `C-o` is an important global binding for `evil-jump-backward`.  I don't think we should override it, just like `'` and `\``.

If not `C-o`, then what about `O`?

jojojames:
Sure, sounds good.

jojojames:
For the rationale, 'Jumping' and 'Go to Definition' are separate concepts.

> Existing bindings:

    mu4e has j in Emacs, J in Evil.
    C-o is possibly used for “display other window”.

mu4e is a good example. J makes perfect sense but gd doesn't. 

Any kind of jump should probably exist on 'J' and go to definition is 'gd'.

Ambrevar:
Acknowledged.  What other "jumps" can you think of?

jojojames:
I'll have to get back to you on that as I think it's a relatively common idiom but not many come to mind.

mu4e and notmuch at the very least.

Ambrevar:
Evil also has its own jumps with `C-i` and `C-o`.
Maybe that could be re-used instead of `J`?

Ambrevar:
Instead of "go to definition" which is too specific or inappropriate phrasing in some contexts, I suggest "follow".

Thus there are two binding sets we should add to the rationale: "following" and "jumping".

- "Following" follows / displays (possibly in other window) the thing at point.

- "Jumping" is unrelated to the thing at point.

jojojames:
> Evil also has its own jumps with C-i and C-o.
Maybe that could be re-used instead of J?

Lets not do that, those 'jumps' from evil are named accordingly from VIM.

The 'J' style jumps are Emacs style jumps.

> Instead of "go to definition" which is too specific or inappropriate phrasing in some contexts, I suggest "follow".

[...]

PR #42, opened 2017-11-23
Make sort / follow / select bindings consistent with the rationale

### LINKED ISSUE #20: Rationale: Sorting vs. Filtering
Filtering (a.k.a. searching / narrowing) is an action that is common to many modes.
Sorting less so however.

It seems that many modes in Emacs state use "s" or "S" for sorting.
If we use the same in Evil states, then what key should we reserve for filtering?

Filtering bindings are inconsistent in Emacs state.  We should not override `/` unless the new binding completely supersedes the functionality of `evil-search-forward`.

Ambrevar:
This should fix #20 and #27.
Since I don't use all the modes impacted by this change, this requires some extra feedback.

The biggest change is that "o" and "C-o" do not "follow to other window" anymore.  Instead, "o" is for sorting and "C-o" is back to its global binding, `evil-jump-backward`.

### DIFF
--- a/evil-arc-mode.el
+++ b/evil-arc-mode.el
@@ -45,10 +45,12 @@
     "gg" 'beginning-of-buffer
     "G" 'end-of-buffer
 
-    ;; goto
-    (kbd "<return>") 'archive-view
-    "o" 'archive-extract
-    "O" 'archive-extract-other-window
+    ;; open
+    (kbd "<return>") 'archive-extract
+    (kbd "S-<return>") 'archive-extract-other-window
+    (kbd "M-<return>") 'archive-view
+    "gd" 'archive-extract
+    "gD" 'archive-extract-other-window
 
     "a" 'archive-alternate-display
     "d" 'archive-flag-deleted
--- a/evil-dired.el
+++ b/evil-dired.el
@@ -129,11 +129,12 @@
     "x" 'dired-do-flagged-delete
     "y" 'dired-show-file-type
     "+" 'dired-create-directory
-    ;; goto
-    "o" 'dired-find-file-other-window
-    (kbd "C-o") 'dired-display-file
+    ;; open
+    (kbd "S-<return>") 'dired-find-file-other-window
+    (kbd "M-<return>") 'dired-display-file
+    "gD" 'dired-find-file-other-window
     ;; sort
-    "s" 'dired-sort-toggle-or-edit
+    "o" 'dired-sort-toggle-or-edit
     ;; moving
     "<" 'dired-prev-dirline
     ">" 'dired-next-dirline
--- a/evil-man.el
+++ b/evil-man.el
@@ -48,8 +48,8 @@
 
     ;; goto
     "gm" 'man
-    "gd" 'Man-goto-section
-    "gR" 'Man-follow-manual-reference
+    "gd" 'Man-goto-section ; TODO: "gd" does not match the rationale of "go to definition". Change?
+    "gR" 'Man-follow-manual-reference ; TODO: Make this match Info-follow-reference?
     "gs" 'Man-goto-see-also-section
 
     ;; refresh
--- a/evil-occur.el
+++ b/evil-occur.el
@@ -43,10 +43,12 @@
     [mouse-2] 'occur-mode-mouse-goto
     (kbd "C-c C-c") 'occur-mode-goto-occurrence
 
-    ;; goto
-    (kbd "C-m") 'occur-mode-goto-occurrence
-    "o" 'occur-mode-goto-occurrence-other-window
-    (kbd "C-o") 'occur-mode-display-occurrence
+    ;; open
+    (kbd "<return>") 'occur-mode-goto-occurrence
+    (kbd "S-<return>") 'occur-mode-goto-occurrence-other-window
+    (kbd "M-<return>") 'occur-mode-display-occurrence
+    "gd" 'occur-mode-goto-occurrence
+    "gD" 'occur-mode-goto-occurrence-other-window
 
     "gj" 'occur-next
     "gk" 'occur-prev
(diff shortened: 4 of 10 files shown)


evil-arc-mode.el
```diff
@@ -45,10 +45,10 @@
     "gg" 'beginning-of-buffer
     "G" 'end-of-buffer
 
-    ;; goto
-    (kbd "<return>") 'archive-view
-    "o" 'archive-extract
-    "O" 'archive-extract-other-window
+    ;; follow
+    (kbd "<return>") 'archive-extract
+    (kbd "S-<return>") 'archive-extract-other-window
```
jojojames: I'm thinking, since there was nothing using 'o' and 'O' for this mode, we should keep the bindings there and also add the ones for s-return and m-return, etc.

One thing that's come to mind is that these s/m-RETURN bindings don't work in terminal. Ideally we'd have our set of bindings support terminal + GUI emacs. So I think we need another alternative to use for the terminal case.

evil-cider.el
```diff
@@ -49,7 +49,7 @@
       (kbd "M-p") #'cider-test-previous-result
       (kbd "M-n") #'cider-test-next-result
 
-      ;; goto
+      ;; follow
```
jojojames: Mentioned in the other thread, but I'm not a fan of the word 'follow'.

evil-dired.el
```diff
@@ -131,11 +131,11 @@
     "x" 'dired-do-flagged-delete
     "y" 'dired-show-file-type
     "+" 'dired-create-directory
-    ;; goto
-    "o" 'dired-find-file-other-window
-    (kbd "C-o") 'dired-display-file
+    ;; follow
+    (kbd "S-<return>") 'dired-find-file-other-window
```
jojojames: Mentioned above, but it'd be good to have an idea how we want to support terminal usage.

evil-man.el
```diff
@@ -48,7 +48,7 @@
 
     ;; goto
     "gm" 'man
-    "gd" 'Man-goto-section
+    "gd" 'Man-goto-section ; TODO: This is not a "go to definition".
```
jojojames: This seems like a stray TODO?

evil-occur.el
```diff
@@ -43,10 +43,10 @@
     [mouse-2] 'occur-mode-mouse-goto
     (kbd "C-c C-c") 'occur-mode-goto-occurrence
 
-    ;; goto
-    (kbd "C-m") 'occur-mode-goto-occurrence
-    "o" 'occur-mode-goto-occurrence-other-window
-    (kbd "C-o") 'occur-mode-display-occurrence
+    ;; follow
+    (kbd "<return>") 'occur-mode-goto-occurrence
+    (kbd "S-<return>") 'occur-mode-goto-occurrence-other-window
```
jojojames: Same here, story for terminal usage.

(Last comment on this even though there are multiple files which would want one.)

evil-profiler.el
```diff
@@ -55,8 +54,8 @@
     "i" 'profiler-report-describe-entry
     "=" 'profiler-report-compare-profile
 
-    ;; goto
-    "gd" 'profiler-report-find-entry
```
jojojames: Was it necessary to remove 'gd' here?

evil-arc-mode.el
Ambrevar: Yes, terminals are limited.  It should be noted that not all Emacs' default bindings work on terminals.
One that comes to mind is C-M-% (`query-replace-regexp`) on QWERTY layouts.

My suggestion: let's support terminals as much as possible, but let's not make the rationale crazy just for the sake of it.
In this case: if `o`/`O` are available, then let's bind them.

evil-dired.el
Ambrevar: In this case, `O` is already taken, but maybe that won't be the case once the dired bindings have been worked out.

evil-man.el
Ambrevar: Bad phrasing, but I meant here that we should change the binding: this has nothing to do with "go to definition".

Should we define a standard binding for commands "listing stuff/sections"?

evil-occur.el
Ambrevar: Can be set to `o`/`O`, there is no conflict.

evil-profiler.el
Ambrevar: Not per se, but I'm trying to make sense of the rationale.
Should "follow" and "go to definition" be the same concepts and bindings? If not, then this profiler command is a "follow" and should be bound to `RET` and not `gd`.

Let's discuss this in #27.

evil-arc-mode.el
jojojames: I agree with you but, to be fair, that command C-M-% is relatively rare to do. I'm guessing the commands we're going to be replacing are somewhat more common (o -> [S-C-M] return) which makes it more of an issue.

jojojames: Maybe we should use 'go' and 'gO' also?

jojojames:
Feel free to merge this in whenever you're comfortable with it. Although there's a few things that probably need to be discussed, we can continue iterating on and improving it after.

evil-arc-mode.el
Ambrevar: That would be unnecessary if we use `gd` and `gD`, no?

Ambrevar:
I've done some changes.  Have a look and merge if you're happy.

jojojames:
Looks good to me, I think you'll have to rebase your branch before it can be merged in.

Ambrevar:
Just for linearity or do you want me to squash to commits into only one?

jojojames:
Linearity. Squash rebase is good too.

evil-arc-mode.el
jojojames: I think we should use go and gO to mirror what S-Return and M-Return would be (for terminal use).

Again, I don't necessarily believe gd == RET conceptually.

gd/gD -- go to definition related 

ret/s-ret/m-ret - related to opening things

Adding go/gO would work.

Along with that, go could be used in programming modes to 'go to other', e.g. test or header/impl files.

PR #50, opened 2017-12-06
Make evil-collection-mode-list customizable.

dieggsy:
Would be nice if the user could use the custom interface to determine what gets enabled and what doesn't.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -34,7 +34,11 @@
 ;;      (evil-minibuffer-init)
 
 ;;; Code:
-(defvar evil-collection-mode-list
+(defgroup evil-collection nil
+  "A set of keybindings for Evil mode"
+  :group 'evil)
+
+(defcustom evil-collection-mode-list
   `(ag
     alchemist
     anaconda-mode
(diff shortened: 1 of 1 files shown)


dieggsy:
Well, looks like this needs a bit more work for the custom interface to display the list properly

dieggsy:
Force pushed some changes, though `sexp` is probably still too broad

evil-collection.el
```diff
@@ -34,7 +34,11 @@
 ;;      (evil-minibuffer-init)
 
 ;;; Code:
-(defvar evil-collection-mode-list
+(defgroup evil-collection nil
+  "A set of keybindings for Evil mode"
+  :group 'evil)
```
jojojames: Group is probably 'evil-collection or 'evilcoll (after I make the namespace changes).

jojojames:
This looks reasonable otherwise. Once the :group variable is changed, I can take a look after merging.

dieggsy:
The group is correct for the `defcustom` hovever the `:group` in a defgroup should not be itself - I think this leads to the group containing itself

noctuid:
Yeah, `:group` is correct for the `defgroup`. The `:prefix` would be `evilcoll-` or  whatever ends up being used.

jojojames:
Cool, sounds good. :)

I'll merge this and test it later.

PR #54, opened 2017-12-07
Fix updated prefix in minibuffer-setup-hook

dieggsy:
Heh, this broke my emacs.


jojojames:
Thanks, I knew I broke some stuff. :(

PR #59, opened 2017-12-10
Add more ivy-minibuffer-map bindings

dieggsy:
Also consolidate binding location. I think all ivy bindings should go in `evil-collection-ivy`. The insert bindings are because some of evil's bindings override ivy's. (Although, maybe we shouldn't override C-r in insert mode, even in the minibuffer... ?)

### DIFF
--- a/evil-collection-ivy.el
+++ b/evil-collection-ivy.el
@@ -75,7 +75,23 @@
     "gc" 'ivy-occur-toggle-calling
 
     ;; quit
-    "q" 'quit-window))
+    "q" 'quit-window)
+
+  (defvar evil-collection-setup-minibuffer)
+  (when evil-collection-setup-minibuffer
+    (evil-define-key 'normal ivy-minibuffer-map
+      (kbd "<escape>") 'abort-recursive-edit
+      (kbd "<return>") 'exit-minibuffer
+      (kbd "C-m") 'ivy-done
+      "j" 'ivy-next-line
+      "k" 'ivy-previous-line)
+
+    (evil-define-key 'insert ivy-minibuffer-map
+      [backspace] 'ivy-backward-delete-char
+      (kbd "C-r") 'ivy-reverse-i-search
+      (kbd "C-n") 'ivy-next-line
+      (kbd "C-p") 'ivy-previous-line)))
+
 
 (provide 'evil-collection-ivy)
 ;;; evil-collection-ivy.el ends here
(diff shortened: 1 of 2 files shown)


fbergroth:
Also see https://github.com/jojojames/evil-collection/pull/55#issuecomment-350210652

dieggsy:
I added a couple more bindings, though I'm not sure a check to see if the minibuffer setup is enabled is necessary - this will only take effect if it's enabled either way.

jojojames:
I think we should gate those keybinds with the check for evil-collection-setup-minibuffer since it's non-obvious that's what it's for.

dieggsy:
@jojojames ok, added.

jojojames:
Might need to defvar it.

In toplevel form:

evil-collection-ivy.el:80:9:Error: reference to free variable ‘evil-collection-setup-minibuffer’

make: *** [compile] Error 1

jojojames:
As @fbergroth mentioned, it looks like there's quite a few keys in those keymaps. It'd be good to add them at some point (doesn't have to be in this request).

I don't use evil-minibuffer so I don't have much of an opinion on those keybindings.

dieggsy:
@jojojames I actually had considered this when _not_ putting it in a check for `evil-minibuffer`, but thought that concern was premature and the var would be defined anyway. Guess not. I think it still feels cleaner to me to just not do this check, but I can add the defvar if you feel strongly about that.

dieggsy:
Regrading the "quite a few keys", I haven't (yet) noticed others clashing - do we want to just define all of them anyway (even if they work as is?)

jojojames:
@dieggsy For the check, I believe the warning is spurious (as in realistically, that variable is defined by the time it hits this codepath but the byte compiler doesn't know that). I think the gate is good. Everytime I come back to this issue, I have to do a mental switch to think about minibuffer + evil.

So it helps, at least for me. ;)

@dieggsy Up to you or if @fbergroth wants to tackle them at some point. (re: defining those extra keys)

dieggsy:
@jojojames If we think they _all_ need to be redefined, I'm probably up for that at some point in the future, but I'm also comfortable just adding the ones that don't work.

It does _sorta_ feel safer to redefine them all in insert mode...

dieggsy:
Added the defvar

jojojames:
I think it's better to redefine all of them (at some point), but I'm also a fan of incremental improvement. So do what works!

I'll merge this for now and if you decide to add more later, we can do it in a separate pr.

dieggsy:
Sounds good to me.

PR #61, opened 2017-12-18
Add key translation functionality

noctuid:
This commit isn't ready to be merged. Any feedback would be appreciated. Some notes:

- in the readme, I'll add detailed explanation and an example of defining `evil-collection-user-setup-function` with some example swaps
- `cl-lib` is added as an explicit dependency (already used by evil)
- I added a macro wrapper for swapping keys: `evil-collection-swap-key`
- I've removed the general.el-specific functionality; the general.el versions have tests; I haven't tested the evil-collection versions yet but will do so before finalizing the pull request

I think that a `evil-collection-mode-map-(p|a)list` should be added (would a plist or alist be preferred?). It would pair each mode with a list of keymap names for that mode (I can add this). This would require more work to maintain, but I think that it's necessary given that a significant number of these modes have multiple keymaps (e.g. elfeed).

I also think that more may need to be done to make it easy to limit the scope of translations. For example, `hjkl` swaps are only desirable in keymaps that rebind `hjkl` to the mode-specific navigation keys. For this case, it's probably not an issue, since if `hjkl` are all unbound in the keymap, all new definitions will just be nil (those positions will still be bound to the default `evil-next-line`, `evil-previous-line`, etc.). I think the main case where this could be an issue is for keys that can have different meanings (e.g. if `o` happened to be used for something else in modes where there is no sorting and the user only wanted to translate a key to `o` in modes where it is used for sorting). These situations would require users to create their own blacklists/whitelists. I'm not sure about how much of an issue this could be, whether it would make sense to have `evil-collection` provide information about the types of keybindings made in each mode, or whether there is some other solution. I think this can be addressed with a later pull request if necessary.

### DIFF
--- a/evil-collection-helm.el
+++ b/evil-collection-helm.el
@@ -48,6 +48,14 @@
 
 (declare-function helm-window "helm-lib")
 
+(defconst evil-collection-helm-maps '(help-map
+                                      help-find-files-map
+                                      helm-read-file-map
+                                      helm-generic-files-map
+                                      helm-buffer-map
+                                      helm-moccur-map
+                                      helm-grep-map))
+
 ;; From https://github.com/emacs-helm/helm/issues/362.
 ;; Also see https://emacs.stackexchange.com/questions/17058/change-cursor-type-in-helm-header-line#17097.
 ;; TODO: With Evil, the cursor type is not right in the header line and the evil
@@ -112,6 +120,7 @@
                    :background (face-background 'default)))
            header-line-format)
           (when update (force-mode-line-update)))))))
+
 
 (defun evil-collection-helm-setup ()
   "Set up `evil' bindings for `helm'."
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -8,7 +8,7 @@
 ;; Pierre Neidhardt <ambrevar@gmail.com>
 ;; URL: https://github.com/emacs-evil/evil-collection
 ;; Version: 0.0.1
-;; Package-Requires: ((emacs "25.1") (evil "1.2.13"))
+;; Package-Requires: ((emacs "25.1") (cl-lib "0.5") (evil "1.2.13"))
 ;; Keywords: evil, tools
 
 ;; This program is free software; you can redistribute it and/or modify
@@ -32,7 +32,7 @@
 ;; This is so because many users find it confusing.
 
 ;;; Code:
-
+(require 'cl-lib)
 (require 'evil)
 
 (defvar evil-want-integration)
@@ -158,6 +158,98 @@ this confusing. It will be included if
   :type '(repeat (choice symbol sexp))
   :group 'evil-collection)
 
+(defvar evil-collection-setup-hook nil
+  "Hook run by `evil-collection-init' for each mode that is evilified.
+This hook runs after all setup (including keybindings) for a mode has already
+taken place. The arguments passed to functions for this hook are the name of the
+mode and a list of keymap names (i.e. symbols, not actual keymaps) customized by
+Evil Collection for that mode. More arguments may be added in the future, so
+functions added to this hook should include a \"&rest _rest\" for forward
+compatibility.")
+
+(defun evil-collection--translate-key (state keymap-symbol
+                                             translations
+                                             destructive)
+  "Helper function for `evil-collection-translate-key'.
+In the keymap corresponding to STATE and KEYMAP-SYMBOL, make the key
+TRANSLATIONS. When DESTRUCTIVE is non-nil, make the TRANSLATIONS destructively
+without creating/referencing a backup keymap."
+  (let* ((backup-keymap-symbol (intern (format "evil-collection-%s%s-backup-map"
+                                               keymap-symbol
+                                               (if state
+                                                   (format "-%s-state" state)
+                                                 ""))))
+         (keymap (symbol-value keymap-symbol))
+         (lookup-keymap (if (and (not destructive)
+                                 (boundp backup-keymap-symbol))
+                            (symbol-value backup-keymap-symbol)
+                          (copy-keymap
+                           (if state
+                               (evil-get-auxiliary-keymap keymap state t t)
+                             keymap))))
+         (maps (cl-loop for (key replacement) on translations by 'cddr
+                        ;; :destructive can be in TRANSLATIONS
+                        unless (keywordp key)
+                        collect key
+                        and collect (when replacement
+                                      (lookup-key lookup-keymap replacement)))))
+    (unless (or destructive
+                (boundp backup-keymap-symbol))
+      (set backup-keymap-symbol lookup-keymap))
+    (apply #'evil-define-key* state keymap maps)))
+
+;;;###autoload
+(cl-defun evil-collection-translate-key (states keymaps
+                                                &rest translations
+                                                &key destructive
+                                                &allow-other-keys)
+  "Translate keys in the keymap(s) corresponding to STATES and KEYMAPS.
+STATES should be the name of an evil state, a list of states, or nil. KEYMAPS
+should be a symbol corresponding to the keymap to make the translations in or a
+list of keymap symbols. Like `evil-define-key', when a keymap does not exist,
+the keybindings will be deferred until the keymap is defined, so
+`with-eval-after-load' is not neccessary. TRANSLATIONS corresponds to a list of
+key replacement pairs. For example, specifying \"a\" \"b\" will bind \"a\" to
+\"b\"'s definition in the keymap. Specifying nil as a replacement will unbind a
+key. If DESTRUCTIVE is nil, a backup of the keymap will be stored on the initial
+invocation, and future invocations will always look up keys in the backup
+keymap. When no TRANSLATIONS are given, this function will only create the
+backup keymap without making any translations. On the other hand, if DESTRUCTIVE
+is non-nil, the keymap will be destructively altered without creating a backup.
+For example, calling this function multiple times with \"a\" \"b\" \"b\" \"a\"
+would continue to swap and unswap the definitions of these keys. This means that
+when DESTRUCTIVE is non-nil, all related swaps/cycles should be done in the same
+invocation."
+  (declare (indent defun))
+  (unless (listp keymaps)
+    (setq keymaps (list keymaps)))
+  (unless (and (listp states)
+               (not (null states)))
+    (setq states (list states)))
+  (dolist (keymap-symbol keymaps)
+    (dolist (state states)
+      (evil-delay `(and (boundp ',keymap-symbol)
+                        (keymapp ,keymap-symbol))
+          `(evil-collection--translate-key ',state ',keymap-symbol
+                                           ',translations ,destructive)
+        'after-load-functions t nil
+        (symbol-name (cl-gensym (format "evil-collection-translate-key-in-%s"
+                                        keymap-symbol)))))))
+
+;;;###autoload
+(defmacro evil-collection-swap-key (states keymaps &rest args)
+  "Wrapper around `evil-collection-translate-key' for swapping keys.
+STATES, KEYMAPS, and ARGS are passed to `evil-collection-translate-key'. ARGS
+should consist of key swaps (e.g. \"a\" \"b\" is equivalent to \"a\" \"b\" \"b\"
+\"a\" with `evil-collection-translate-key') and optionally keyword arguments for
+`evil-collection-translate-key'."
+  (declare (indent defun))
+  (setq args (cl-loop for (key replacement) on args by 'cddr
+                      collect key and collect replacement
+                      and unless (keywordp key)
+                      collect replacement and collect key))
+  `(evil-collection-translate-key ,states ,keymaps ,@args))
+
 ;;;###autoload
 (defun evil-collection-init ()
   "Register the Evil bindings for all modes in `evil-collection-mode-list'.
@@ -178,7 +270,14 @@ instance:
       (dolist (req reqs)
         (with-eval-after-load req
           (require (intern (concat "evil-collection-" (symbol-name m))))
-          (funcall (intern (concat "evil-collection-" (symbol-name m) "-setup"))))))))
+          (funcall (intern (concat "evil-collection-" (symbol-name m)
+                                   "-setup")))
+          (let ((mode-keymaps
+                 (ignore-errors
+                   (symbol-value
+                    (intern (format "evil-collection-%s-maps" m))))))
+            (run-hook-with-args 'evil-collection-setup-hook
+                                m mode-keymaps)))))))
 
 (defvar evil-collection-delete-operators '(evil-delete
                                            evil-cp-delete
--- a/readme.org
+++ b/readme.org
@@ -384,6 +384,122 @@ definition".
     - ~f11~ Step Into
     - ~S-f11~ Step Out
 
+** Key Translation
+~evil-collection-translate-key~ allows binding a key to the definition of
+another key in the same keymap (comparable to how vim's keybindings work). Its
+arguments are the ~states~ and ~keymaps~ to bind/look up the key(s) in followed
+optionally by keyword arguments (currently only ~:destructive~) and
+key/replacement pairs. ~states~ should be nil for non-evil keymaps, and both
+~states~ and ~keymaps~ can be a single symbol or a list of symbols.
+
+This function can be useful for making key swaps/cycles en masse. For example,
+someone who uses an alternate keyboard layout may want to retain the ~hjkl~
+positions for directional movement in dired, the calendar, etc.
+
+Here's an example for Colemak of making swaps in a single keymap:
+#+begin_src emacs-lisp
+(evil-collection-translate-key nil 'evil-motion-state-map
+  ;; colemak hnei is qwerty hjkl
+  "n" "j"
+  "e" "k"
+  "i" "l"
+  ;; add back nei
+  "j" "e"
+  "k" "n"
+  "l" "i")
+#+end_src
+
+Here's an example of using ~evil-collection-setup-hook~ to cycle the keys for
+all modes in ~evil-collection-mode-list~:
+#+begin_src emacs-lisp
+(defun my-hjkl-rotation (_mode mode-keymaps &rest _rest)
+  (evil-collection-translate-key 'normal mode-keymaps
+    "n" "j"
+    "e" "k"
+    "i" "l"
+    "j" "e"
+    "k" "n"
+    "l" "i"))
+
+;; called after evil-collection makes its keybindings
+(add-hook 'evil-collection-setup-hook #'my-hjkl-rotation)
+
+(evil-collection-init)
+#+end_src
+
+A more common use case of ~evil-collection-translate-key~ would be for keeping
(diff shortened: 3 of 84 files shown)


evil-collection.el
```diff
@@ -112,6 +114,62 @@ this confusing. It will be included if
...
+  (declare (indent defun))
+  (let* ((keymap (if state
+                     (evil-get-auxiliary-keymap
+                      (symbol-value keymap-name) state t t)
+                   (symbol-value keymap-name)))
+         (backup-keymap-name (intern (format
+                                      "evil-collection-%S-%S-state-backup-map"
+                                      keymap-name
+                                      (or state 'no))))
+         (lookup-keymap (if (and (not destructive)
```
jojojames: I see we're playing around with destructive.

When would you recommend one or the other? (e.g. swapping keys with destructive set to nil and vice versa)

evil-collection.el
```diff
@@ -112,6 +114,62 @@ this confusing. It will be included if
...
+STATE, KEYMAP-NAME, and ARGS are passed to `evil-collection-translate-key'. ARGS
+should consist of key swaps (e.g. \"a\" \"b\" is equivalent to \"a\" \"b\" \"b\"
+\"a\" with `evil-collection-translate-key') and optionally keyword arguments for
+`evil-collection-translate-key'."
+  (declare (indent defun))
+  (setq args (cl-loop for (key replacement) on args by 'cddr
+                      collect key and collect replacement
+                      and unless (keywordp key)
+                      collect replacement and collect key))
+  `(evil-collection-translate-key ,state ,keymap-name ,@args))
```
jojojames: Nice.

jojojames:
>    in the readme, I'll add detailed explanation and an example of defining evil-collection-user-setup-function with some example swaps

+1 

A tangential but related topic is rebinding SPC so users can use that as their leader/hydra key.

 Right now, I use general/evil-intercept-maps so I can get access to SPC. I imagine that'd be a good one to use as an example.

>    cl-lib is added as an explicit dependency (already used by evil)

Fine by me. Just curious why we need to depend on it ourselves too if evil already does?

>    I added a macro wrapper for swapping keys: evil-collection-swap-key

Nice, I like that the calling semantics feel similar to evil-define-key.

>    I've removed the general.el-specific functionality; the general.el versions have tests; I haven't tested the evil-collection versions yet but will do so before finalizing the pull request

Do you intend to port over any of the tests?

> I think that a evil-collection-mode-map-(p|a)list should be added (would a plist or alist be preferred?). It would pair each mode with a list of keymap names for that mode (I can add this). This would require more work to maintain, but I think that it's necessary given that a significant number of these modes have multiple keymaps (e.g. elfeed).

Who owns managing this alist/plist? More concretely, which files will manage the alist/plist itself? I'm shooting in the dark, but I'd guess this alist will be managed by either the user or per mode (in each mode file). I think this sounds like a good idea, just not sure what you're imagining.

I think if it's a more user oriented feature, a plist sounds better? If it's more internally managed, alist would probably work. I don't any a strong opinion on plist vs alist in this case either way.

>  I think the main case where this could be an issue is for keys that can have different meanings (e.g. if o happened to be used for something else in modes where there is no sorting and the user only wanted to translate a key to o in modes where it is used for sorting). These situations would require users to create their own blacklists/whitelists.

I do think that occurence will (and already do) happen somewhat often. I'm not sure if it'll be a problem as you said though.

evil-collection.el
```diff
@@ -132,7 +190,9 @@ instance:
       (dolist (req reqs)
         (with-eval-after-load req
           (require (intern (concat "evil-collection-" (symbol-name m))))
-          (funcall (intern (concat "evil-collection-" (symbol-name m) "-setup"))))))))
+          (funcall (intern (concat "evil-collection-" (symbol-name m) "-setup")))
+          (when (fboundp 'evil-collection-user-setup)
+            (funcall #'evil-collection-user-setup m)))))))
```
jojojames: I was thinking this is nice if the user wants to define all their translations in one place, but I'm not sure if *all* users like that kind of configuration.

I could imagine myself putting random swap code in different 'mode' files that I have strewn across my .emacs.d.

I wonder if we want to do something where the user can both define a global function that will run with evil-collection-user-setup as well as a more precise function they can define like evil-collection-specific-mode-user-setup.

The code may just check for one first and run it (or even just run both if both are available).

evil-collection.el
```diff
@@ -112,6 +114,62 @@ this confusing. It will be included if
...
+non-nil, the keymap will be destructively altered without a backup being
+created. If DESTRUCTIVE is nil, a backup of the keymap will be stored on the
+initial invocation, and future invocations will always look up keys in the
+backup keymap. On the other hand, if DESTRUCTIVE is non-nil, calling this
+function multiple times with \"a\" \"b\" \"b\" \"a\", for example, would
+continue to swap and unswap the definitions of these keys. This means that when
+DESTRUCTIVE is non-nil, all related swaps/cycles should be done in the same
+invocation."
+  (declare (indent defun))
+  (let* ((keymap (if state
```
jojojames: I might be wrong here, but taking a look at the state variable, we're constraining it to only passing in one 'state variable right?

If so, I'm curious if we want to support more states at once, kinda similar to evil-define-key.

(evil-define-key '(normal visual) keymap key command)

noctuid:
> A tangential but related topic is rebinding SPC so users can use that as their leader/hydra key.
> Right now, I use general/evil-intercept-maps so I can get access to SPC. I imagine that'd be a good one to use as an example.

I'll add an example of binding `SPC` in an intercept map and using the translation function to put the default `SPC` keybindings on another key.

> Fine by me. Just curious why we need to depend on it ourselves too if evil already does?

It's not necessary. I just added it to be explicit since `cl-loop` is used in the file (and evil isn't required). Is the lack of `(require 'evil)` intentional? There is now a warning because `evil-get-auxiliary-map` is used, so I'm wondering whether I should require evil or just add a `declare-function` (since it's not really necessary to require evil immediately when requiring evil-collection).

> Do you intend to port over any of the tests?

I can if you want but don't think the duplication is necessary.

> Who owns managing this alist/plist? More concretely, which files will manage the alist/plist itself? I'm shooting in the dark, but I'd guess this alist will be managed by either the user or per mode (in each mode file). I think this sounds like a good idea, just not sure what you're imagining.

It would just be defined below `evil-collection-mode-list` in `evil-collection.el` but wouldn't be customizable. It would be useful in the user setup function for getting the actual keymap names to make the translations in. The keymap list could just be given as a second argument to the setup function regardless of the implementation. Alternatively each file could define a consistently named variable (e.g. `evil-collection-<mode>-maps`). 

> When would you recommend one or the other? (e.g. swapping keys with destructive set to nil and vice versa)

The default should usually be used, especially in the user's init. `:destructive t` could potentially be useful for experimentation/interactive evaluation. The limitation of the default behavior is that you can't translate a key to another key that was defined after the first `evil-collection-translate-key`.

> The code may just check for one first and run it (or even just run both if both are available).

I'll add a call to `evil-collection-<mode>-user-setup`, if it exists, as well.

> If so, I'm curious if we want to support more states at once, kinda similar to evil-define-key.

There's definitely a use case for allowing multiple keymaps. I'm not sure how often multiple states would be useful, but I'll add that as well.

jojojames:
> I'll add an example of binding SPC in an intercept map and using the translation function to put the default SPC keybindings on another key.

Cool.

> It's not necessary. I just added it to be explicit since cl-loop is used in the file (and evil isn't required). Is the lack of (require 'evil) intentional? There is now a warning because evil-get-auxiliary-map is used, so I'm wondering whether I should require evil or just add a declare-function (since it's not really necessary to require evil immediately when requiring evil-collection).

I think the require was just forgotten. Feel free to require it. @Ambrevar, chime if the lack of a require was intentional.

> I can if you want but don't think the duplication is necessary.

Up to you.

> It would just be defined below evil-collection-mode-list in evil-collection.el but wouldn't be customizable.
> Alternatively, each file could define a consistently named variable (e.g. evil-collection-<mode>-maps).

I don't have a strong preference here but I'm leaning towards the alternate unless the first choice sways you more.

>  It would be useful in the user setup function for getting the actual keymap names to make the translations in.
> The keymap list could just be given as a second argument to the setup function regardless of the implementation.

Sounds good.

> I'll add a call to evil-collection-<mode>-user-setup, if it exists, as well.

Cool.

> There's definitely a use case for allowing multiple keymaps. I'm not sure how often multiple states would be useful, but I'll add that as well.

Yeah, I think whatever matches evil-define-key's calling convention most consistently is great.

evil-collection.el
Ambrevar: In terms of naming, something like `load-hook` or `setup-hook` would be more conventional, wouldn't it?

I'd expect users to look for a "hook" somewhere, but not necessarily for "user-setup".

We could also use the `run-hooks` function.

evil-collection.el
Ambrevar: I think supporting several states would be quite useful, although not mandatory.

Ambrevar:
> I think the require was just forgotten. Feel free to require it. @Ambrevar, chime if the lack of a require was intentional.

Indeed, `(require 'evil)` was forgotten I suppose.

noctuid:
It probably would make more sense to use load hooks for each individual mode, but thinking about it again, I don't think this would be all this useful. It's already easy enough to add mode-specific configuration afterwards without hooks.

The convenient thing about the user setup function is that it would repeatedly/automatically be called with the mode/keymaps as arguments. This isn't all that similar to most hooks, but I think you're right that the naming would be more expected. I'll change it to `evil-collection-setup-hook` and use `run-hook-with-args`.

jojojames:
> It probably would make more sense to use load hooks for each individual mode, but thinking about it again, I don't think this would be all this useful. It's already easy enough to add mode-specific configuration afterwards without hooks.

I don't have a strong opinion either way.

noctuid:
I've pushed the changes and rebased. However, trying to use the setup hook will currently cause failures with certain modes (e.g. edebug and cider) because some of the keymaps do not exist when the translation function runs. Do you think it would make sense to split these into multiple files (e.g. add an edebug-x file like there is an image+ one instead of using `with-eval-after-load`)? I'll probably just use `evil-delay` so that it won't matter if the keymap isn't loaded, but that can make things uglier. As a side note, why do some files require the mode with `noerror` and declare the keymap variables (e.g. rtags)?

Also, I still need to add the `SPC` example to the readme.

jojojames:
> However, trying to use the setup hook will currently cause failures with certain modes (e.g. edebug and cider) because some of the keymaps do not exist when the translation function runs.

> Do you think it would make sense to split these into multiple files (e.g. add an edebug-x file like there is an image+ one instead of using with-eval-after-load)?

For something like edebug and edebug-x, it's not unreasonable to split the files but I took a look at cider and all the with-eval-after-loads are for the same package.

But then it brings up another point which is why there's with-eval-after-loads on those. I would think evil-define-key handles it in a deferred state.
In that case, it might have be better to remove the with-evals in cider *and* split the edebug files.

If those are the only ones, it's probably fine to handle it case by case. But it does seem like evil-delay is the right approach (to match evil-define-key semantics).

Let me know if it'd be helpful for me to split/extract those files myself so we can keep this PR focused.

> I'll probably just use evil-delay so that it won't matter if the keymap isn't loaded, but that can make things uglier. As a side note, why do some files require the mode with noerror and declare the keymap variables (e.g. rtags)?

+1, unless it makes the code much more complicated?

I think it was mostly to make the byte compiler happen. Semantically, we shouldn't need to require anything.

With the exception of the multi modes that trigger the same evil-collection-package.

> Also, I still need to add the SPC example to the readme.

+1 !

noctuid:
> +1, unless it makes the code much more complicated?

It shouldn't be too bad, and I think it's the expected behavior given that this is what `evil-define-key` does.

> Let me know if it'd be helpful for me to split/extract those files myself so we can keep this PR focused.

The split can be done later if wanted, but the unnecessary `with-eval-after-load`s definitely should be removed (so that those `evil-define-key`s happen before the translations).

> I think it was mostly to make the byte compiler happen. Semantically, we shouldn't need to require anything.

I understand the require statement, but then aren't the `defvar`s redundant unless the require statement fails? When would it be okay for the `require` to fail?

jojojames:
> The split can be done later if wanted, but the unnecessary with-eval-after-loads definitely should be removed (so that those evil-define-keys happen before the translations).

Cool, I'll look into removing the evals in cider.

> I understand the require statement, but then aren't the defvars redundant unless the require statement fails? When would it be okay for the require to fail?

I think, for something like the rtags defvars, those are unnecessary since evil-define-key will defer them anyways. Vars like in the typescript mode will require them in order to make the byte compiler happy.

```
(require 'typescript-mode nil t)
(defvar typescript-indent-level)
```

noctuid:
I've updated the pull request, but it's still not ready to be merged. For one, I'd forgotten that I'd overridden an evil function a while ago to make it possible to make as an auxiliary keymap an intercept map, meaning the example in the readme won't work. I made a pull request to evil that will fix that if accepted.

Let me know if the changes to the readme are clear though.

> Vars like in the typescript mode will require them in order to make the byte compiler happy.

I guess the idea is to prevent compiler warnings even if the user doesn't have the package installed.

jojojames:
> Let me know if the changes to the readme are clear though.

Looks good to me.

Ambrevar:
@noctuid: You wrote:

> It's already easy enough to add mode-specific configuration afterwards without hooks.

How?

noctuid:
I fixed a few documentation issues and rebased. I added a note in the example that an evil version from today or later is required (is that sufficient?).

> How?

It would be fairly simple (but not ideal) to just advise specific setup functions with `:after`. The hook added in this commit is also more convenient since it's automatically called after *every* setup function (meaning it can also easily be used for general configuration/setup not related to single, specific modes).

noctuid:
I still need to add keymap lists for new modes added since I first made the commit (forgot about that).

noctuid:
Alright, the hook should now work for the new modes and not error for files that don't make any keybindings.

jojojames:

I'll merge unless @Ambrevar or @noctuid thinks it's not ready?

evil-collection-helm.el
```diff
@@ -113,6 +121,7 @@
            header-line-format)
           (when update (force-mode-line-update)))))))
 
+
```
Ambrevar: Why an extra newline?  Not that it really matters...

evil-collection-vdiff.el
```diff
@@ -30,6 +30,9 @@
 (require 'vdiff nil t)
 (require 'evil)
 
+;; TODO: maybe way to specify usage of `evil-define-minor-mode-key'
+;; (defconst evil-collection-vdiff-maps)
```
Ambrevar: Why is it commented?  I don't understand the above comment.

evil-collection.el
```diff
@@ -158,6 +158,94 @@ this confusing. It will be included if
   :type '(repeat (choice symbol sexp))
   :group 'evil-collection)
 
+(defvar evil-collection-setup-hook nil
+  "Hook run by `evil-collection-init' for each mode that is evilified.
+All functions will be called with the name of the mode and a list of keymap
+names corresponding to the keymaps that evil-collection made keybindings in for
```
Ambrevar: Isn't "that evil-collection made keybindings in for that mode" weirdly phrased?

evil-collection.el
```diff
@@ -158,6 +158,94 @@ this confusing. It will be included if
...
+that mode.")
+
+(defun evil-collection--translate-key
+    (state keymap-symbol translations destructive)
```
Ambrevar: Is it good Elisp style to start the argument list on the next line?

readme.org
```diff
@@ -384,6 +384,122 @@ definition".
     - ~f11~ Step Into
     - ~S-f11~ Step Out
 
+** Key Translation
+~evil-collection-translate-key~ allows binding a key to the definition of
+another key in the same keymap (comparable to how vim's keybindings work). Its
```
Ambrevar: vim -> Vim

readme.org
```diff
@@ -384,6 +384,122 @@ definition".
...
+arguments are the ~states~ and ~keymaps~ to bind/look up the key(s) in followed
```
Ambrevar: Spurious "the" before ~states~?

readme.org
```diff
@@ -384,6 +384,122 @@ definition".
...
+By default, the first invocation of ~evil-collection-translate-key~ will make a
+backup of the keymap. Each subsequent invocation will look up keys in the backup
+instead of the original. This means that a call to
+~evil-collection-translate-key~ will always have the same behavior even if
+evaluated multiple times. When ~:destructive t~ is specified, keys are looked up
+in the keymap as it is currently. This means that a call to
+~evil-collection-translate-key~ that swapped two keys would continue to
+swap/unswap them with each call. Therefore when ~:destructive t~ is used, all
+cycles/swaps must be done within a single call to
+~evil-collection-translate-key~. To make a comparison to vim keybindings,
```
Ambrevar: vim -> Vim

Ambrevar:
Beside a few nits, looks good to me too!  Thank you for this excellent work!

noctuid:
I don't think that there is any recommended style for splitting the arglist. Personally I prefer moving it down entirely when arguments can't be grouped, but in this case, I've changed it to be split in same way as the main translation function is which probably makes more sense.

Let me know if the docstring for `evil-collection-setup-hook` makes more sense now.

As for the TODO, `evil-define-minor-mode-key` doesn't actually work with named keymaps, so some new way to handle it would need to be added. For example, something instead of just a symbol to mark that it's not a keymap but a minor mode (e.g. instead of the symbol ='vdiff-mode-map=, which isn't altered, the entry might be, for example, '(minor-mode . vdiff-mode)). The translation function would then need to detect that this isn't an actual keymap symbol and use `evil-get-minor-mode-keymap` instead of `evil-get-auxiliary-keymap` and `evil-define-minor-mod-key` instead of `evil-define-key*`. Do you think that this behavior is worth adding or have a better suggestion for how to handle this?

evil-collection.el
```diff
@@ -158,6 +158,98 @@ this confusing. It will be included if
   :type '(repeat (choice symbol sexp))
   :group 'evil-collection)
 
+(defvar evil-collection-setup-hook nil
+  "Hook run by `evil-collection-init' for each mode that is evilified.
+This hook runs after all setup (including keybindings) for a mode has already
+taken place. The arguments passed to functions for this hook are the name of the
+mode and a list of keymap names (i.e. symbols not actual keymaps) corresponding
```
Ambrevar: Comma after "symbols"?

evil-collection.el
```diff
@@ -158,6 +158,98 @@ this confusing. It will be included if
...
+to the keymaps that evil-collection binds keys in for that mode. More arguments
```
Ambrevar: I still find this sentence a tad convoluted.
What about:
> and a list of keymap names (i.e. symbols, not actual keymaps) customized by Evil Collection for that mode.

?

Ambrevar:
Regarding the TODO: It sounds reasonable but you know much more than me in that area.
If you plan to work on it later, could you replace the comment with your last explanation?

noctuid:
I updated the docstring. I just removed the TODO for now. If someone wants this functionality in the future, I can further consider the best way to do it.

PR #62, opened 2017-12-24
Add quit bindings to evil-collection ibuffer

jgkamat:
Per [the readme](https://github.com/jojojames/evil-collection#quitting-q-zq-zz), `q` is used for quitting when macros are not useful. Binding `q` to bury buffer is a bad decision since it goes against this philosophy, and there's already a default binding for bury buffer - `b`. Since `b` dosen't make too much sense in ibuffer mode, it makes sense to use this key in evil mode too. 

Let me know if you think differently, while I'm open to binding other keys to bury-buffer, I think `q` should be reserved for quitting to keep the experience seamless.

### DIFF
--- a/evil-collection-ibuffer.el
+++ b/evil-collection-ibuffer.el
@@ -79,7 +79,7 @@
     "`" 'ibuffer-switch-format
     "-" 'ibuffer-add-to-tmp-hide
     "+" 'ibuffer-add-to-tmp-show
-    "q" 'ibuffer-bury-buffer
+    "b" 'ibuffer-bury-buffer
     (kbd ",") 'ibuffer-toggle-sorting-mode
     (kbd "o i") 'ibuffer-invert-sorting
     (kbd "o a") 'ibuffer-do-sort-by-alphabetic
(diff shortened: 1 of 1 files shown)


jojojames:
Yup, you're right, I was going to change the q to quit-window or something (as you did). I think burying buffer can be set to 'x'? What do you think?

jgkamat:
I actually really like the `x` binding, but that seems used by ibuffer-do-kill-on-deletion-marks (which I think is a reasonable binding for it as well). Do you have any other ideas? `gb` is an easy one, but that's a bit inconvenient.

jojojames:
Hmnn, that's probably why I haven't changed it yet. What about 'X'? I think that is the shell pipe command that's also on another key.

bling:
what's wrong with keeping `b` as bury?

jgkamat:
Yes, `X` is taken as well, I think that `b` might be the best option unless we need `b` for 'back' (which I don't think we do, but I could be missing something).

jojojames:
I think it's more intuititive for 'b' to go 'back'.

We should try and keep navigation commands as much as possible.

@jgkamat 'X' is taken but the command is also available on other keys so it's not the end of the world to replace it.

jgkamat:
@jojojames, agreed, that is true. I've just added a commit which rebinds `X` to bury-buffer and unbinds the previous binding.

bling:
FWIW, i think keeping `b` for bury better here because none of the other navigation commands (like, w, e, B, E) do navigation.

jojojames:
> FWIW, i think keeping b for bury better here because none of the other navigation commands (like, w, e, B, E) do navigation.

You're right (in that the navigation commands don't do navigation) but that's something I wanted to change.

I had modified ibuffer relatively quickly when removing the previous evilified-state on them. Ideally, most/all the navigation bindings should be useable (especially if it can be triggered from another key).

I think w,e,b,E,B, etc and maybe l (I think that may be on redisplay right now) should be updated.

Most/all bindings in evil-collection are not set in stone so we should continue iterating on them. :)

jojojames:
Going to merge this but if you guys had more comments on ibuffer directionally (navigation-first or ibuffer-commands first), @bling, @jgkamat, feel free to open another issue or PR!

PR #65, opened 2017-12-28
anaconda-mode-view-mode-map doesn't exist; use anaconda-view-mode-map instead

edwargix:
I don't know whether `evil-collection` is meant for the stable packages in melpa. If it is, then reject this PR because the current setup is correct. However the latest anaconda-mode on melpa renamed `anaconda-mode-view-mode-map` to `anaconda-view-mode-map`.

### DIFF
--- a/evil-collection-anaconda-mode.el
+++ b/evil-collection-anaconda-mode.el
@@ -34,7 +34,10 @@
   ;; Bindings don't seem to be set the first time.
   (add-hook 'anaconda-mode-hook #'evil-normalize-keymaps)
 
-  (evil-define-key 'normal anaconda-view-mode-map
+  ;; latest anaconda uses `anaconda-view-mode-map'; anaconda stable
+  ;; uses `anaconda-mode-view-mode-map'
+  (evil-define-key 'normal (or anaconda-view-mode-map
+			       anaconda-mode-view-mode-map)
     "gj" 'next-error-no-select
     "gk" 'previous-error-no-select
     (kbd "C-j") 'next-error-no-select
--- a/evil-collection-anaconda-mode.el
+++ b/evil-collection-anaconda-mode.el
@@ -34,7 +34,7 @@
   ;; Bindings don't seem to be set the first time.
   (add-hook 'anaconda-mode-hook #'evil-normalize-keymaps)
 
-  (evil-define-key 'normal anaconda-mode-view-mode-map
+  (evil-define-key 'normal anaconda-view-mode-map
     "gj" 'next-error-no-select
     "gk" 'previous-error-no-select
     (kbd "C-j") 'next-error-no-select
--- a/evil-collection-anaconda-mode.el
+++ b/evil-collection-anaconda-mode.el
@@ -36,8 +36,9 @@
 
   ;; latest anaconda uses `anaconda-view-mode-map'; anaconda stable
   ;; uses `anaconda-mode-view-mode-map'
-  (evil-define-key 'normal (or anaconda-view-mode-map
-			       anaconda-mode-view-mode-map)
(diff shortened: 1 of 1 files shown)


jojojames:
I don't mind just tracking latest melpa but it'd be nice to have compatibility with melpa stable too if it's not too much work. (i.e. We don't support melpa-stable as a policy but should work with it if it's easy to do so.)

Any ideas on what we should do to get both working?

Maybe

```
(when (boundp 'anaconda-view-mode-map)
    (defvaralias anaconda-mode-view-mode-map anaconda-mode-view-mode-map))
```

I've never used defvaralias so not sure if that'll work.

evil-define-key also binds those keymaps later if they're not already loaded so that could be tricky there too. I think there's a cleaner solution to this. @Ambrevar, thoughts?

We can merge this and just track latest melpa if it'll make things easier for everyone too.

Ambrevar:
If used in only one place, `(or anaconda-mode-view-mode-map anaconda-view-mode-map)` should be enough, right?

With a comment on top, of course :)

fbergroth:
https://github.com/proofit404/anaconda-mode/commit/e72e9beeb8c80acfee4d85748464d1c5147946ad I think you should update `next-error-no-select` with `anaconda-view-mode-next-definition` as well (same for `previous-error-no-select`).

noctuid:
This is going to cause an error for the version where the first keymap doesn't exist. You could just have both and one `evil-define-key` would just never make the keybindings.

Ambrevar:
What is going to cause an error?  My suggestion?  Why?

noctuid:
The variable is never defined (`Symbol's value as a variable is void:...`).

jojojames:
@noctuid Really dumb question,

> You could just have both and one evil-define-key would just never make the keybindings.

But could you give a quick example?

noctuid:
```
(evil-define-key 'normal anaconda-mode-view-mode-map
....)
(evil-define-key 'normal anaconda-view-mode-map
....)
```
One would just be deferred forever since the map would never exist. Though since anaconda-mode is required, it would probably just be better to add a slightly more complicated check (e.g. `(if (boundp 'anaconda-view-mode-map) anaconda-view-mode-map...`).

jojojames:
Ah, yeah, I was afraid that was what you meant.

We might want to do that for now @edwargix and we can think about a better way to handle these types of compatibility issues separately.

For:

> proofit404/anaconda-mode@e72e9be I think you should update next-error-no-select with anaconda-view-mode-next-definition as well (same for previous-error-no-select).

I'm guessing defalias would work? You don't have to do ^ @edwargix but it'd be great if you did too. :)

edwargix:
I tested the most recent commit with the stable anaconda package

edwargix:
@jojojames not sure what you mean by using `defalias`. Could you give an example? I think @fbergroth meant replacing `next-error-no-select` with `anaconda-view-mode-next-definition` (which was written specifically for an `anaconda-view-mode` buffer).

jojojames:
> I tested the most recent commit with the both the latest and the stable anaconda packages

Cool. It'll work for this PR, I'll merge it in. I think @noctuid's version seems a little more bulletproof. It seems that anaconda's mode maps are already loaded by the time it hits this piece of code (e.g. the mode map won't need to leverage the delay aspect of evil-define-key).

> Could you give an example? I think @fbergroth meant replacing next-error-no-select with anaconda-view-mode-next-definition (which was written specifically for an anaconda-view-mode buffer), and the same for previous.

I didn't look at the specific commit yet but my thought was that that function (anaconda-view-mode-next-definition) was a newly introduced function that wouldn't be available in the stable version of anaconda-mdoe.

PR #68, opened 2018-01-06
Key bindings for free-keys

talwrii:
https://github.com/Fuco1/free-keys

### DIFF
--- a/evil-collection-free-keys.el
+++ b/evil-collection-free-keys.el
@@ -1,0 +1,42 @@
+;;; evil-collection-free-keys.el --- Evil bindings for free-keys -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017 Tal Wrii
+
+;; Author: Tal Wrii <talwrii@gmail.com>
+;; Maintainer: James Nguyen <talwrii@gmail.com>
+;; URL: https://github.com/jojojames/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, free-keys, tools
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published
+;; by the Free Software Foundation; either version 3, or (at your
+;; option) any later version.
+;;
+;; This file is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; For a full copy of the GNU General Public License
+;; see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Evil bindings for free-keys.
+;; https://github.com/Fuco1/free-keys
+
+;;; Code:
+(require 'free-keys)
+(require 'evil)
+
+(defun evil-collection-free-keys-setup ()
+  "Set up `evil' bindings for `free-keys'."
+  (evil-set-initial-state 'free-keys-mode 'motion)
+  (evil-define-key 'motion free-keys-mode-map
+    "b" 'free-keys-change-buffer
+    "p" 'free-keys-set-prefix
+    "q" 'quit-window))
+
+(provide 'evil-collection-free-keys)
+;;; evil-collection-free-keys ends here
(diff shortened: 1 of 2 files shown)


evil-collection-free-keys.el
```diff
@@ -0,0 +1,42 @@
+;;; evil-collection-free-keys.el --- Evil bindings for free-keys -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017 Tal Wrii
+
+;; Author: Tal Wrii <talwrii@gmail.com>
+;; Maintainer: James Nguyen <talwrii@gmail.com>
```
jojojames: Might want to add @Ambrevar here too.

evil-collection-free-keys.el
```diff
@@ -0,0 +1,42 @@
...
+;;
+;; For a full copy of the GNU General Public License
+;; see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Evil bindings for free-keys.
+;; https://github.com/Fuco1/free-keys
+
+;;; Code:
+(require 'free-keys)
```
jojojames: This can just be (require 'free-keys nil t).

evil-collection-free-keys.el
```diff
@@ -0,0 +1,42 @@
...
+(require 'evil)
+
+(defun evil-collection-free-keys-setup ()
+  "Set up `evil' bindings for `free-keys'."
+  (evil-set-initial-state 'free-keys-mode 'motion)
```
jojojames: This should be 'normal state.

evil-collection-free-keys.el
```diff
@@ -0,0 +1,42 @@
...
+  (evil-define-key 'motion free-keys-mode-map
+    "b" 'free-keys-change-buffer
```
jojojames: Hmnn, I really like 'b' as a navigation movement. How do you feel about using 'c' instead?

evil-collection-free-keys.el
```diff
@@ -0,0 +1,42 @@
+;;; evil-collection-free-keys.el --- Evil bindings for free-keys -*- lexical-binding: t -*-
```
jojojames: Look into adding the free-keys entry to the evil-collection-mode-list here.

https://github.com/jojojames/evil-collection/blob/master/evil-collection.el

jojojames:
This looks good other than those minor changes.

evil-collection-free-keys.el
talwrii: This mode is a "readonly mode" i.e. you don't edit the buffer. Does it still want to be normal?

evil-collection-free-keys.el
talwrii: So, the mode hard codes these keybindings into documentation in the header-line.
I don't know what a work around for this would look like.

https://github.com/Fuco1/free-keys/blob/edfd69dc369b2647447b7c28c7c1163b1ddf45b4/free-keys.el#L213

evil-collection-free-keys.el
talwrii: Done.

evil-collection-free-keys.el
talwrii: Done

evil-collection-free-keys.el
talwrii: Done.

talwrii:
I've made most of the changes you suggested.

A couple of queries:

`free-keys` is a "readonly mode" - it displays you information that is machine generated and overwritten by various commands. Should I still use normal mode?

The `b` keybinding is hard coded in the header-line:

https://github.com/Fuco1/free-keys/blob/edfd69dc369b2647447b7c28c7c1163b1ddf45b4/free-keys.el#L213

What is a good thing to do?

evil-collection-free-keys.el
jojojames: Yeah, see https://github.com/jojojames/evil-collection/issues/21

evil-collection-free-keys.el
jojojames: You can add a hook to free-keys-mode-hook (or whatever the hook's name is) to change the header line locally.

jojojames:
^ Responded in comments.

jojojames:
@talwrii This discussion went a little stale so I just went ahead and added free-keys. :)

Thanks!

Issue #49, opened 2017-12-06
Support evil-search-module set to evil-search


dieggsy:
Here's a (perhaps a bit unclear) similiar issue opened and resolved in evil-magit: https://github.com/emacs-evil/evil-magit/issues/33

tl;dr, when `evil-search-module` is set to `evil-search` (either before evil loads or with custom), different search commands are used, namely `evil-ex-search*` instead of `evil-search*`.

jojojames:
Thanks.

Now the question is whether to hardcode the same logic into every file that needs to explicitly bind / or do *something* else.

https://github.com/emacs-evil/evil-magit/commit/a24186be7cc2cdab24b56f6dcc4665eeb8349c1a

noctuid:
I think all the logic should be in one place. A couple of ways of doing this off the top of my head

- Create variables (e.g. `evil-search-forward`) that hold menu items to conditionally act as one command or the other based on the current value of `evil-search-module`
- Make no changes to the default keybindings; instead, provide a function that takes a keymap as an arg and calls `substitute-key-definition` for all relevant definitions like `evil-select-search-module` does; let the user put it in their setup function

dieggsy:
@noctuid for the second option, couldn't this function be called in elisp when doing `evil-*-setup` ? It could just check `evil-search-module` and make substitutions conditionally.

noctuid:
Yeah, it could do it the same way as evil does it and just automatically check (but still provide the function if the user wants to switch later). I kind of like the the idea of keybindings checking `evil-search-module` each time, but I doubt people toggle `evil-search-module`, so it probably wouldn't be useful.

dieggsy:
In fact, `evil-search-module` needs to be set before evil loads, so toggling it would probably be a pain?

noctuid:
You can use `evil-select-search-module` to change it afterwards. `evil-collection` could do the same thing.

jojojames:
@noctuid I don't have a preference either way here. Which of these is more preferable to you?

@Ambrevar Feel free to chime in too.

Ambrevar:
Hmmm... Stupid question, but what's the difference between the two commands?  I see the point is not moved to the same position during the search, but there must be something else...

If I get the issue correctly (sorry, little time, I just skimmed it), this is the same as for the `C-u` question (in doc-view, image-mode, etc.).
We certainly need a general approach to this, and if it's convincing enough, maybe upstream (Evil) would consider it.

[...]

PR #74, opened 2018-01-12
Support evil-search-module set to evil-search (#49)

jojojames:
@dieggsy @Ambrevar @noctuid

Would appreciate a quick look if you got time.

### DIFF
--- a/evil-collection-ag.el
+++ b/evil-collection-ag.el
@@ -29,6 +29,7 @@
 ;;; Code:
 (require 'ag nil t)
 (require 'evil)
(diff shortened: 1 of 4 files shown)


noctuid:
Looks good to me. It might be more convenient to add a macro for menu-item creation (you could just add an `evil-collection` equivalent of `general-predicate-dispatch`).

jojojames:
@noctuid Thanks, I'll go ahead and merge.

As for the macro, I'll keep that in mind. I think it'll be good to add once we see more of these menu-items crop up.

PR #83, opened 2018-01-27
which-key and Buffer-menu

maxsandova:
The which key works just fine.  For some reason I can't get the Buffer-menu to work, even after commenting `(evil-add-hjkl-bindings Buffer-menu-mode-map 'motion)`

### DIFF
--- a/evil-collection-buffer-menu.el
+++ b/evil-collection-buffer-menu.el
@@ -1,0 +1,48 @@
+;;; evil-collection-buffer-menu.el --- Evil bindings for buffer-menu -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017 Pierre Neidhardt
+
+;; Author: Maximiliano Sandoval <msandova@protonmail.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <ambrevar@gmail.com>
+;; URL: https://github.com/jojojames/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, buffer-menu, tools
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published
+;; by the Free Software Foundation; either version 3, or (at your
+;; option) any later version.
+;;
+;; This file is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+;;
+;; For a full copy of the GNU General Public License
+;; see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Evil bindings for `Buffer-mode'
+
+;;; Code:
+
+(require 'evil)
+(require 'buff-menu nil t) ;;Necessary?
+
+;; This function does not excecute itself and I don't know why
+;; therefore ZZ and ZQ are not working, j and k works thanks to evil-integration
+(defun evil-collection-buffer-menu-setup ()
+  "Set up `evil' bindings for `Buffer-menu'."
+
+  (evil-define-key '(normal motion) Buffer-menu-mode-map
+    "ZQ" 'evil-quit
+    "ZZ" 'quit-window
+    "j"  'evil-next-line
+    "k"  'evil-previous-line
+    "h"  'evil-backward-char
+    "l"  'evil-forward-char))
+
+(provide 'evil-collection-buffer-menu)
+;;; evil-collection-buffer-menu.el ends here
--- a/evil-collection-integration.el
+++ b/evil-collection-integration.el
@@ -100,7 +100,7 @@
 
 ;;; Buffer-menu
 
-(evil-add-hjkl-bindings Buffer-menu-mode-map 'motion)
+(evil-add-hjkl-bindings Buffer-menu-mode-map '(normal motion)) ;;temporary?
 
 ;; dictionary.el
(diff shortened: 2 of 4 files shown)


jojojames:
It's probably better to split these up into two separate PRs.

evil-collection-integration.el
```diff
@@ -100,7 +100,7 @@
 
 ;;; Buffer-menu
 
-(evil-add-hjkl-bindings Buffer-menu-mode-map 'motion)
+(evil-add-hjkl-bindings Buffer-menu-mode-map '(normal motion)) ;;temporary?
```
jojojames: This line should be removed completely.

maxsandova: I did that with no success, it still does not bind anything

maxsandova:
I managed to get it working, I don't understand what is the problem, see evil-collection.

evil-collection-buffer-menu.el
```diff
@@ -0,0 +1,45 @@
...
+;; Evil bindings for `Buffer-menu'
+
+;;; Code:
+
+(require 'evil)
+
+(defun evil-collection-buffer-menu-setup ()
+  "Set up `evil' bindings for `Buffer-menu'."
+  
+  (evil-define-key '(motion normal) Buffer-menu-mode-map
```
jojojames: The full keymap should be bound. This is only a subset.

Also, you should remove the 'motion', it's inconsistent with the rest (most) of the bindings which start at 'normal'.

evil-collection-buffer-menu.el
```diff
@@ -0,0 +1,45 @@
...
+    "ZQ" 'evil-quit
+    "ZZ" 'quit-window
+    "j"  'evil-next-line
```
jojojames: You shouldn't have to bind hjkl.

evil-collection-integration.el
```diff
@@ -98,9 +98,9 @@
   (when (fboundp 'evil-repeat-abort)
     (evil-repeat-abort)))
 
-;;; Buffer-menu
+;;; Buffer-mode
 
-(evil-add-hjkl-bindings Buffer-menu-mode-map 'motion)
+;; (evil-add-hjkl-bindings Buffer-menu-mode-map 'motion)
```
jojojames: These lines should be completely removed.

evil-collection-reftex.el
```diff
@@ -34,11 +34,11 @@
 
 ;; original code can be found in reftex-ref.el
 (setq reftex-select-label-prompt
-  "Select: [RET]select [j]next [k]previous [gr]escan [go]context [q]uit [?]help")
+  "Select: [RET]select [j]next [k]previous [gr]escan [go]context [q]uit [g?]help")
```
jojojames: It'd be better for these to go in separate PRs.

evil-collection.el
```diff
@@ -167,5 +168,8 @@ instance:
           (require (intern (concat "evil-collection-" (symbol-name m))))
           (funcall (intern (concat "evil-collection-" (symbol-name m) "-setup"))))))))
 
+(require 'evil-collection-buffer-menu)
+(funcall 'evil-collection-buffer-menu-setup)
```
jojojames: We shouldn't be doing this here. This item should go in as part of the original mode list.

evil-collection-which-key.el
```diff
@@ -0,0 +1,48 @@
...
+;;; Code:
+
+(require 'evil)
+(require 'which-key nil t)
+
+;; which key is coded so that the promp properly shows j and k as the bindings
+(defun evil-collection-which-key-setup ()
+  "Set up `evil' bindings for `which-key'."
+  
+  (evil-define-key nil which-key-C-h-map
```
jojojames: Which-key is fine to add, as a separate PR.

jojojames: I'm not sure why 'nil' is needed here but you could also probably use 'define-key' when the precedence order is different than normal.

jojojames:
Thanks! As mentioned, these changes should go in separate PRs usually.

If you manage to get all the issues sorted out, I could technically just rebase and merge all the changes together but it's probably better to use separate branches in the future.

jojojames:
No problem, everyones learning. :)

I'll fix up some of the things you added myself so you can skip them (like tweaking reftex).

evil-collection.el
maxsandova: It did not work for some reason

jojojames: You can just add a print statement in the setup function to be called to see if it's being triggered.

maxsandova: I used the code
```elisp
(dolist (mode evil-collection-mode-list)
  (let ((m mode)
	(reqs (list mode)))
    (when (listp mode)
      (setq m (car mode)
	    reqs (cdr mode)))
    (dolist (req reqs)
      (with-eval-after-load req
	(print (intern (concat "evil-collection-" (symbol-name m))))
	))))
```
The result in the *Messages* buffer was
```
evil-collection-bookmark
evil-collection-comint
evil-collection-company
evil-collection-compile
evil-collection-custom
evil-collection-debug
evil-collection-diff-mode
evil-collection-dired
evil-collection-elisp-mode
evil-collection-help
evil-collection-image
evil-collection-info
evil-collection-ivy
evil-collection-occur
evil-collection-package-menu
evil-collection-xref
nil
```
A lot of things are missing, in particular buffer-mode. I don't know what to make of it. I also put a print in the function `evil-collection-buffer-mode` and It did not activate, the function is not loaded at all.

jojojames: Try using this file -> https://github.com/jojojames/evil-collection/commit/602d24f07dad7fefe72293bda63a400842b04f62

It's because buff-menu is not provided.

PR #87, opened 2018-01-28
Help mode

maxsandova:
Added an alternative to "RET"

### DIFF
--- a/evil-collection-help.el
+++ b/evil-collection-help.el
@@ -51,6 +51,7 @@
     ;; "\C-j" 'help-go-forward
     ;; "\C-k" 'help-go-back
 
+    "go" 'push-button
     "g?" 'describe-mode
     "gr" 'revert-buffer
     "<" 'help-go-back
--- a/evil-collection-help.el
+++ b/evil-collection-help.el
@@ -51,8 +51,11 @@
     ;; "\C-j" 'help-go-forward
     ;; "\C-k" 'help-go-back
 
-    "go" 'push-button ;; this should open in same window
-    "gO" 'push-button ;; this should open in another window
+    ;; The following bindings don't do what they are supposed to. "go" should open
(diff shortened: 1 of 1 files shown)


jojojames:
Can you squash these commits into one?

You can squash and then force push your branch and it'll appear as one commit when all is said and done.

evil-collection-help.el
```diff
@@ -51,6 +51,8 @@
     ;; "\C-j" 'help-go-forward
     ;; "\C-k" 'help-go-back
 
+    "go" 'push-button ;; this should open in same window
+    "gO" 'push-button ;; this should open in another window
```
jojojames: Are these comments true? These both trigger 'push-button'.

maxsandova: thats why I put should.  Neither do what they should `push-button` sometimes opens in the same window sometimes in other-window context-wise.

jojojames: OK, it might be better to clarify the comments here.

maxsandova: Ok, do you have a suggestion on implementation? I don't see the point on writing two functions here, I would just bind both as in the commit.

jojojames: You can just update the comments here to reflect that they 'should' open in same/another window *but* don't in all cases.

maxsandova: Ok, that I will do.

jojojames:
Thanks.

maxsandova:
Thanks to you, without realizing it, you teached me git, cask, code etiquette among other things.

jojojames:
@maxsandova Thanks, it was all you. :)

Last nit for your edification, you might want to take a look at https://github.com/erlang/otp/wiki/writing-good-commit-messages

It's not the *only* way to do things, but it's *one* nice way to do it.

PR #88, opened 2018-01-31
Neotree mode support

maxsandova:
Added support for neotree-mode. 
* Some keys require revision (e.g. "s,S") some others
like "H,K,L" could be used more. 
* Many keybings do not coincide with
the ones used in spacemacs. 
* `neotree-change-root` could use a better
alternative.

### DIFF
--- a/evil-collection-neotree.el
+++ b/evil-collection-neotree.el
@@ -29,12 +29,7 @@
 ;;; Code:
 
 (require 'evil)
-(require 'neotree)
-
-;; TODO: Compare with spacemacs keybindings.
-;; https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bspacemacs/spacemacs-ui-visual/packages.el
-
-;; Maybe there are better uses of H, K, L,
+(require 'neotree nil t)
 
 (defun evil-collection-neotree-setup ()
   "Set up `evil' bindings for `neotree'."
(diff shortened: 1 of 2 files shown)


maxsandova:
In the Readme it says that neotree uses "gr", thats not true. Neotree uses "g" to refresh and "gr" is binded only in spacemacs to refresh.

evil-collection-neotree.el
```diff
@@ -0,0 +1,100 @@
...
+;; For a full copy of the GNU General Public License
+;; see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Evil bindings for `neotree'
+
+;;; Code:
+
+(require 'evil)
+(require 'neotree)
```
jojojames: This should be (require 'neotree nil t).

Try running make with cask to check for compile errors.

evil-collection-neotree.el
```diff
@@ -0,0 +1,100 @@
...
+
+;; TODO: Compare with spacemacs keybindings.
+;; https://github.com/syl20bnr/spacemacs/blob/bd7ef98e4c35fd87538dd2a81356cc83f5fd02f3/layers/%2Bspacemacs/spacemacs-ui-visual/packages.el
```
jojojames: It might be good to take inspiration from spacemacs here though it's not a requirement.

evil-collection-neotree.el
```diff
@@ -0,0 +1,100 @@
...
+    "z"		(neotree-make-executor
+		 :dir-fn  'neo-open-dir)
+    "gd"	(neotree-make-executor
+		 :dir-fn 'neo-open-dired)
+    "gD"	(neotree-make-executor
+		 :dir-fn 'neo-open-dired)
+    "go"	(neotree-make-executor
+		 :file-fn 'neo-open-file
+		 :dir-fn  'neo-open-dir)
+    "gO"	'neotree-quick-look
```
jojojames: Can you re-align these so that there's no alignment? (e.g. no variable spaces between key and function.

jojojames:
I don't use neotree so the eventual bindings, you can determine. It generally looks OK other than those comments.

evil-collection-neotree.el
maxsandova: Already changed it. I am waiting to figure if they keybings are good enough and I will commit.

evil-collection-neotree.el
maxsandova: Do you mean?
```lisp
"go"	        (neotree-make-executor
		 :file-fn 'neo-open-file
		 :dir-fn 'neo-open-dir)
"gO"	'neotree-quick-look
```

jojojames:
There should be only one space between the key and the function name.

E.g.

(evil-define-key blah blah
 (kbd "key") 'function)

not:

(evil-define-key blah blah
 (kbd "key")                                  'function)

PR #89, opened 2018-02-01
Remove debugs in evil-collection-diff-mode.el

edkolev:
I often see the `Evil Diff: enter motion state` message while using magit and vdiff. Seems to me that these messages were meant for debugging purposes, hence this PR.

Thank you!

### DIFF
--- a/evil-collection-diff-mode.el
+++ b/evil-collection-diff-mode.el
@@ -42,10 +42,8 @@
   "Make read-only in motion state, writable in normal state."
   (if buffer-read-only
       (progn
-        (evil-motion-state)
-        (message "Evil Diff: enter motion state"))
-    (evil-normal-state)
-    (message "Evil Diff: enter normal state")))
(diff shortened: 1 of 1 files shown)


jojojames:
Thanks, I'll merge this right now and we can look into adding debug-messages later if needed.

edkolev:
Great, thanks for the fast response!

edkolev:
@jojojames just FYI, I now see that `diff-hl`'s `diff-hl-flydiff-mode` is causing the spam of `Evil Diff: enter motion state` messages - every time I pause writing, I get the message. Mabe diff-hl makes heavy use of `diff-mode`

jojojames:
I don't have a problem with disabling those messages if you have a PR.

edkolev:
Hey, no need, this very PR already removed all the debug messages that I've run into. I was just clarifying what triggered them.

PR #90, opened 2018-02-01
Epa

maxsandova:
Support for epa

* Added a preliminary set of keybingins.
* The more troubling one is "v" which is bind to `epa-verify-file` in
  this mode there isn't any reason to use `visual-state` so it may be Ok.

### DIFF
--- a/evil-collection-epa.el
+++ b/evil-collection-epa.el
@@ -1,0 +1,65 @@
+;;; evil-collection-epa.el --- Evil bindings for epa-mode -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017 James Nguyen
+
+;; Author: Maximiliano Sandoval <msandova@protonmail.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <ambrevar@gmail.com>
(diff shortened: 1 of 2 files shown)


evil-collection-epa.el
```diff
@@ -0,0 +1,65 @@
...
+    "u"	 'epa-unmark-key
+    "gr" 'revert-buffer
+    "q"	 'epa-exit-buffer
+    "E"	 'epa-decrypt-file
+    "d"	 'epa-delete-keys	;; Originally this was binded to "r".
+    "ZZ" 'quit-window
+    "ZQ" 'evil-quit
+    
+    ;; Unchanged keybindings.
+    "v"	'epa-verify-file	;; Conflicts with visual.
```
jojojames: How about capital V? and we leave v for visual select?

evil-collection-epa.el
```diff
@@ -0,0 +1,65 @@
...
+       				;; There isn't any reason to use visual anyways.
+    "s"	'epa-sign-file
+    "e"	'epa-encrypt-file
+    "i"	'epa-import-keys
```
jojojames: The alignment of these key->functions are off (like the other PR).

Issue #70, opened 2018-01-08
Suggestion: add keybindings for vdiff


edkolev:
Hi, and thanks for this package!

I would like to suggest adding the following [vdiff](https://github.com/justbur/emacs-vdiff) keybindings.

```elisp
  (evil-define-minor-mode-key 'normal 'vdiff-mode
    "]c" 'vdiff-next-hunk
    "[c" 'vdiff-previous-hunk
    "go" 'vdiff-receive-changes
    "gp" 'vdiff-send-changes)
```

The `go` and `gp` are intended to mimic vim's `do` and `dp`. I'm not aware of a sane universal way to bind `do` and `dp` without causing possible breakages. The best option I know is provided by general.el, but I don't think it can be used in evil-collection.

I'll submit a PR if you think this is a good addition.

noctuid:
I think the best approach here to mimic `do` and `dp` here would probably be to bind `o` and `p` in operator state, which is how some packages like evil-surround do it. That works for evil-surround since it can be used with any operator, but in this case it should only work for `d`. Advising `evil-delete` or binding `d` to a command that temporarily bound `o` and `p` in `evil-operator-shortcut-map` would probably work. Something similar to this:

```elisp
(evil-define-operator evil-collection-vdiff-delete
  (beg end type register yank-handler)
  "..."
  (interactive "<R><x><y>")
  (define-key evil-operator-shortcut-map "o" #'vdiff-receive-changes)
  (define-key evil-operator-shortcut-map "p" #'vdiff-receive-changes)
  (evil-delete beg end type register yank-handler)
  (define-key evil-operator-shortcut-map "o" nil)
  (define-key evil-operator-shortcut-map "p" nil))
```

I tried using `let` so that `evil-operator-shortcut-map` is only temporarily altered instead of explicitly unbinding the keys, but that doesn't seem to work.

edkolev:
Some time ago I played around with this. Here s the final semi-working snippet 
https://github.com/justbur/emacs-vdiff/issues/14#issuecomment-283597316

I'll try to come up with a solid implementation.

noctuid:
I think using a menu item is preferable since the user can still bind `o`/`p` by default/for different operators.

jojojames:
I'm not familiar with vimdiff but this seems like a good addition to have.

```
    "]c" 'vdiff-next-hunk
    "[c" 'vdiff-previous-hunk
```

[...]

PR #91, opened 2018-02-15
Add vdiff.el support

edkolev:
Refs #70

Please let me know if I can improve the PR.

Maybe the `evil-collection-delete-operators` var should be in evil-collection.el

### DIFF
--- a/evil-collection-vdiff.el
+++ b/evil-collection-vdiff.el
@@ -1,0 +1,58 @@
+;;; evil-collection-vdiff.el --- Evil bindings for vdiff -*- lexical-binding: t -*-
+
+;; Copyright (C) 2017 Pierre Neidhardt
+
+;; Author: Evgeni Kolev <evgenysw@gmail.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <ambrevar@gmail.com>
+;; URL: https://github.com/jojojames/evil-collection
+;; Version: 0.0.1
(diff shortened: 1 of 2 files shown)


edkolev:
> Maybe the evil-collection-delete-operators var should be in evil-collection.el

I just moved it there - CI (package-lint in particular) complained.

jojojames:
I don't use vdiff myself so I can merge this and if someone has more tweaks/feedback, we can iterate on it as it comes.

As for the delete operators, I'll check to see if I can move it to vdiff when I have some time. It's not bad where it is though.

noctuid:
Looks good to me. As for the `evil-collection-delete-operators`, it could potentially be removed in the future if these packages bound `[remap evil-delete]` enstead of `d` itself. This is what lispyville does now. If there were any new alternate delete operators added like this, it would be possible to have them work without adding them to `evil-collection-delete-operators`. For example:

```elisp
(when (or (eq evil-this-operator (key-binding [remap evil-delete]))
          (memq evil-this-operator evil-collection-delete-operators))
  #'vdiff-receive-changes)
```
At this point, this is probably overkill, but it may be worth considering in the future to prevent having to keep track of every delete operator.

jojojames:
^ Sounds good, I'll give it a shot.

@edkolev 

Should this be 'evil' instead in evil-collection-delete-operators?

```
-                                           vil-sp-delete
+                                           evil-sp-delete
```

edkolev:
@jojojames thanks, I just submitted a fix.

Issue #19, opened 2017-11-07
Rationale: Marking


Ambrevar:
The rationale has now some idea about marking.

It's not clear yet if we should ignore `u` or `U` to unmark, since most modes supporting marking do not support "undoing actions".

It's not clear either whether the marking bindings should be togglers and whether they should work on regions.  I like the idea of regions though.

Somelauw:
A simple solution is that in normal mode:

- `m` toggles mark
- `u` undoes

And in visual mode

- `m` marks
- `u` unmarks

jojojames:
It'll probably be good to include the current rationale in this thread as the readme will probably evolve indepedent of this thread.

I think 'u' for unmarking fits pretty well. I'm not sure there's a mode where you have both an unmark and an undo action.

jojojames:
```
    (define-key map (kbd "* *") 'ibuffer-unmark-all)
    (define-key map (kbd "* c") 'ibuffer-change-marks)
    (define-key map (kbd "U") 'ibuffer-unmark-all-marks)
    (define-key map (kbd "* M") 'ibuffer-mark-by-mode)
    (define-key map (kbd "* m") 'ibuffer-mark-modified-buffers)
    (define-key map (kbd "* u") 'ibuffer-mark-unsaved-buffers)
    (define-key map (kbd "* s") 'ibuffer-mark-special-buffers)
    (define-key map (kbd "* r") 'ibuffer-mark-read-only-buffers)
    (define-key map (kbd "* /") 'ibuffer-mark-dired-buffers)
    (define-key map (kbd "* e") 'ibuffer-mark-dissociated-buffers)
    (define-key map (kbd "* h") 'ibuffer-mark-help-buffers)
    (define-key map (kbd "* z") 'ibuffer-mark-compressed-file-buffers)
```

```
    (define-key map "**" 'dired-mark-executables)
    (define-key map "*/" 'dired-mark-directories)
    (define-key map "*@" 'dired-mark-symlinks)
    (define-key map "*%" 'dired-mark-files-regexp)
    (define-key map "*c" 'dired-change-marks)
    (define-key map "*s" 'dired-mark-subdir-files)
    (define-key map "*m" 'dired-mark)
    (define-key map "*u" 'dired-unmark)
    (define-key map "*?" 'dired-unmark-all-files)
    (define-key map "*!" 'dired-unmark-all-marks)
    (define-key map "U" 'dired-unmark-all-marks)
[...]

PR #96, opened 2018-02-24
Normalize mark bindings

Ambrevar:
Now that we've covered a decent number of modes, I've skimmed through them to see which bindings prevailed for marking.
It turns out that their was little deviation, so I've adjusted the few modes that showed inconsistencies.

(I might have missed a few, though.)

I've updated the rationale accordingly.  Let me know what you think.

### DIFF
--- a/evil-collection-bookmark.el
+++ b/evil-collection-bookmark.el
@@ -45,8 +45,6 @@
     "1" 'bookmark-bmenu-1-window
     "x" 'bookmark-bmenu-execute-deletions
     "d" 'bookmark-bmenu-delete
-    "u" 'bookmark-bmenu-unmark
-    "m" 'bookmark-bmenu-mark
     "/" 'bookmark-bmenu-search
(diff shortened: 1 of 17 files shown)


Ambrevar:
Regarding @jojojames last suggestion in #19, the main question that remains is whether we should replace `x` for `#` to execute action on mark.

maxsandova:
I think that `#` it is way out of reach and most of people who come from vanilla emacs are accustomed to use `x`.

Ambrevar:
On a QWERTY keyboard, `#` is fairly easy to press with `Shift-3`.

jojojames:
Might be good to check the CI and see what can fix it up.

Looks good otherwise!

For 'execute mark', I think leaving it on 'x' is probably OK unless 'x' is taken by some deletion operation. We can probably backfill to '#' if so.

Ambrevar:
CI should be fine now.

Ambrevar:
Merged.

PR #104, opened 2018-03-18
Add daemons.el support

jgkamat:
This PR adds support for [daemons.el](https://github.com/cbowdon/daemons.el).

This is a very simple binding, simply binding all the default binds to evil as well. I don't think it's a problem because all the bindings correspond to editing commands (which don't make sense in daemon mode).

Let me know if anything seems wrong! :)

### DIFF
--- a/evil-collection-daemons.el
+++ b/evil-collection-daemons.el
@@ -39,7 +39,11 @@
     "r" #'daemons-reload-at-point
     "R" #'daemons-restart-at-point
 
-    "q" #'quit-window)
+    "gr" #'revert-buffer
+
+    "q" #'quit-window
+    "ZZ" #'quit-window
+    "ZQ" #'quit-window)
 
   ;; Functions are available in daemons-output-mode-map as well
   (evil-define-key '(normal visual) daemons-output-mode-map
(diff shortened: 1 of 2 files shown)


evil-collection-daemons.el
```diff
@@ -0,0 +1,58 @@
...
+
+  ;; Functions are available in daemons-output-mode-map as well
+  (evil-define-key '(normal visual) daemons-output-mode-map
+    (kbd "RET") #'daemons-status-at-point
+    "s" #'daemons-start-at-point
+    "S" #'daemons-stop-at-point
+    "r" #'daemons-reload-at-point
+    "R" #'daemons-restart-at-point
+
+    "q" #'quit-window)
```
Ambrevar: Also bind "ZZ" and "ZQ" to quit-window.

Ambrevar:
Looks pretty good to me.  Thanks for letting me know about daemons.el!

jgkamat:
The `j/k` behavior is handled by evil automatically (since we are in normal mode, `evil-next-line` seems to work fine). 

`gr` needs a binding, I'll add that in the next commit, thanks :)

jojojames:
Looks good! Thanks!

Issue #106, opened 2018-03-24
Bindings for custom "insert modes" where regular insert-state doesn't make sense


terlar:
Some modes where insert-state doesn't make sense, but have some kind of mode that can be triggered to do edits.

# Examples

### ivy-occur-grep-mode
- `i` would run `ivy-wgrep-change-to-wgrep-mode`
### wgrep-mode
- `ZQ` would run `wgrep-abort-changes`
- `ZZ` would run `wgrep-finish-edit`

### dired-mode
- `i` would run `dired-toggle-read-only` (which triggers `wdired-mode`)
- `I` would run `dired-maybe-insert-subdir` (since this is currently bound to `i`)
### wdired-mode
- `ZQ` would run `wdired-abort-changes`
- `ZZ` would run `wdired-finish-edit`

I am not sure if the binding should be `i` or something else, what do you think? Does this sound like a good idea? There is probably more states where regular insert state wouldn't make sense but we have some kind of mode that would change into an editable state.

Ambrevar:
Right now we have `C-x C-q`, which is definitely very emacs-y, and your suggestion makes sense in my opinion.  At first glance, I'd agree with adding this to the rationale.
@jojojames ?

jojojames:
One thing I like about C-x C-q is that you can do it for both entering and existing that state.

If we go with i, we'll need a second binding to leave it (ZQ/ZZ as mentioned above are some ways to exit). 

Maybe we should add ESC in Normal state to fully exit wdired?

Normal State:
i -> enter wdired-like mode
ESC -> exit wdired-like mode
ZZ -> Abort Changes
ZZ -> Finish wdired

Ambrevar:
`C-x C-q` would still work anyways, we don't have to unbind it.  We can make everyone happy! :)

dieggsy:
I think for wdired mode it would make sense to do something like what @noctuid suggests in https://github.com/emacs-evil/evil-collection/issues/107.

noctuid:
Agreed, same for wgrep.

jojojames:
One thing that makes me hesitant about that (:q/:wq) is that wdired/w-ag/w-ivy is that their presentation is more similar to another 'state' and not a new mode/window.

[...]

PR #111, opened 2018-03-31
Wdired+wgrep

jojojames:
Would appreciate a quick look.

https://github.com/emacs-evil/evil-collection/issues/106

@Ambrevar @noctuid @terlar

### DIFF
--- a/evil-collection-wdired.el
+++ b/evil-collection-wdired.el
@@ -1,0 +1,46 @@
+;;; evil-collection-wdired.el --- Bindings for `wdired' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018 James Nguyen
+
+;; Author: James Nguyen <james@jojojames.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <ambrevar@gmail.com>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, emacs, tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


terlar:
Looks good! Thanks for the update.

One binding that I also have added to wgrep mode that might make sense is the `d` bound to `wgrep-mark-deletion`.

evil-collection-wdired.el
```diff
@@ -0,0 +1,46 @@
...
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;; Bindings for `wdired'
```
Ambrevar: Only two `;` here.

evil-collection-wgrep.el
```diff
@@ -0,0 +1,48 @@
...
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;; Bindings for `wgrep'
```
Ambrevar: Same.

evil-collection-wdired.el
Ambrevar: + missing period.

Ambrevar:
Looks good to me.  I think something similar can be done for Helm buffers.  I'll look into it later.

jojojames:
> One binding that I also have added to wgrep mode that might make sense is the d bound to wgrep-mark-deletion.

Hmnn, wouldn't you want to keep evil's normal d?

jojojames:
I'll merge this now but we can continue talking about ^ if you are interested. @terlar

terlar:
That is correct, probably not a good idea to change that behavior, I was thinking it didn't make sense to use `d`, but together with the text-objects it does. However the binding `dd` would make sense to run `wgrep-mark-deletion` as that is currently a noop in this mode (complains about buffer being read-only). I am not sure how to bind these motion keys though. What do you think? This would do what people expect.

PR #113, opened 2018-03-31
Support git-timemachine

wpcarro:
Supports evil bindings for git-timemachine. On my machine, git-timemachine
starts with evil in normal mode, which is problematic when trying to access the
keybindings "n" and "p", which navigate to the next and previous revisions.
Additionally, normal mode eclispses "q", which exits the mode.

I tried using `(evil-set-initial-state 'git-timemachine-mode 'motion)`, but that
didn't work. I assume this is because `git-timemachine` is a minor-mode. To work
around this, I used `add-hook` to ensure motion mode was the initial state.

Once motion mode is the initial state, "p" and "q" become available.
Unfortunately, "n" is still unavailable. To get around this, I used a
buffer-local binding in the local motion state to map "n" appropriately. One
known shortcoming of this approach is that there is no cleanup done after
exiting the mode. Therefore, "n" is still mapped in the local motion map after
`git-timemachine` is quit. This doesn't feel great.

Any suggestions are eagerly welcomed. Forgive any crude techniques that I used
to get this functioning. I just wanted to broach the discussion with some of the
other maintainers to get some insights and hopefully augment my implementation
as needed.

Additionally, if "n" and "p" are not the preferred Evil idioms, please let me
know, and I'll adopt whatever is recommended. It is convenient that "n" and "p"
are set my `git-timemachine` by default, since they also are the Vim way of
searching next and previous items...

### DIFF
--- a/evil-collection-git-timemachine.el
+++ b/evil-collection-git-timemachine.el
@@ -1,8 +1,24 @@
 ;;; evil-collection-git-timemachine.el --- Bindings for `git-timemachine' -*- lexical-binding: t -*-
+
 ;; Author: William Carroll <wpcarro@gmail.com>
+;; URL: https://github.com/emacs-evil/evil-collection
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 ;;; Commentary:
-;; Evil keybindings for `git-timemachine' that conform to the principles outlines in evil-collection
+;; Evil keybindings for `git-timemachine' that conform to the principles
+;; outlines in evil-collection
 
 ;;; Code:
 (require 'evil)
--- a/evil-collection-git-timemachine.el
+++ b/evil-collection-git-timemachine.el
@@ -1,0 +1,22 @@
+;;; evil-collection-git-timemachine.el --- Bindings for `git-timemachine' -*- lexical-binding: t -*-
+;; Author: William Carroll <wpcarro@gmail.com>
+
+;;; Commentary:
+;; Evil keybindings for `git-timemachine' that conform to the principles outlines in evil-collection
+
+;;; Code:
+(require 'evil)
+(require 'git-timemachine)
+
+(defun ecgt--setup-bindings ()
+  "Setup `git-timemachine' since `evil-set-initial-state' is unavailable to
+minor modes."
+  (evil-motion-state nil)
+  (evil-local-set-key 'motion "n" #'git-timemachine-show-next-revision))
+
+(defun evil-collection-git-timemachine-setup ()
+  "Setup `evil' keybindings for `git-timemachine'."
+  (add-hook 'git-timemachine-mode-hook #'ecgt--setup-bindings))
+
+(provide 'evil-collection-git-timemachine)
+;;; evil-collection-git-timemachine.el ends here
--- a/evil-collection-git-timemachine.el
+++ b/evil-collection-git-timemachine.el
@@ -22,17 +22,22 @@
 
 ;;; Code:
 (require 'evil)
-(require 'git-timemachine)
+(require 'git-timemachine nil t)
 
-(defun ecgt--setup-bindings ()
-  "Setup `git-timemachine' since `evil-set-initial-state' is unavailable to
-minor modes."
-  (evil-motion-state nil)
-  (evil-local-set-key 'motion "n" #'git-timemachine-show-next-revision))
+(defvar git-timemachine-mode-map)
+(defconst evil-collection-git-timemachine-map '(git-timemachine-mode-map))
 
 (defun evil-collection-git-timemachine-setup ()
   "Setup `evil' keybindings for `git-timemachine'."
-  (add-hook 'git-timemachine-mode-hook #'ecgt--setup-bindings))
+  (evil-define-minor-mode-key 'normal 'git-timemachine-mode
+    "\C-k" #'git-timemachine-show-previous-revision
+    "\C-j" #'git-timemachine-show-next-revision
+    "q"    #'git-timemachine-quit
+    "gtg"  #'git-timemachine-show-nth-revision
+    "gtt"  #'git-timemachine-show-revision-fuzzy
+    "gty"  #'git-timemachine-kill-abbreviated-revision
+    "gtY"  #'git-timemachine-kill-revision
+    "gtb"  #'git-timemachine-blame))
 
 (provide 'evil-collection-git-timemachine)
 ;;; evil-collection-git-timemachine.el ends here


evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,22 @@
...
+;; Author: William Carroll <wpcarro@gmail.com>
+
+;;; Commentary:
+;; Evil keybindings for `git-timemachine' that conform to the principles outlines in evil-collection
+
+;;; Code:
+(require 'evil)
+(require 'git-timemachine)
+
+(defun ecgt--setup-bindings ()
```
wpcarro: Here `ecgt` stands for `evil-collection-git-timemachine`. I was unsure of the idioms that `evil-collection` preferred for namespacing, but I did want to ensure that the `--` was present to encode the intention of privacy.

jojojames:
For reference, this looks to be the minor-mode declaration for git-timemachine.

```
(define-minor-mode git-timemachine-mode
 "Git Timemachine, feel the wings of history."
 :init-value nil
 :lighter " Timemachine"
 :keymap
 '(("p" . git-timemachine-show-previous-revision)
   ("n" . git-timemachine-show-next-revision)
   ("g" . git-timemachine-show-nth-revision)
   ("t" . git-timemachine-show-revision-fuzzy)
   ("q" . git-timemachine-quit)
   ("w" . git-timemachine-kill-abbreviated-revision)
   ("W" . git-timemachine-kill-revision)
   ("b" . git-timemachine-blame))
 :group 'git-timemachine)
```

jojojames:
You'll want to bind all those keys explicitly in an 'evil-define-key' block set to normal state.

So a partial snippet of that will look like:
```
(evil-define-key 'normal git-timemachine-mode-map
   "q" 'git-timemachine-quit
   "etc" 'etc)
```

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,22 @@
+;;; evil-collection-git-timemachine.el --- Bindings for `git-timemachine' -*- lexical-binding: t -*-
```
jojojames: You can use evil-collection-free-keys (or some other file) as an example/template.

You'll want the license/copyright/maintainer/etc as the header.

Of course, leave yourself as the author. :)

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,22 @@
...
+;; Author: William Carroll <wpcarro@gmail.com>
+
+;;; Commentary:
+;; Evil keybindings for `git-timemachine' that conform to the principles outlines in evil-collection
+
+;;; Code:
+(require 'evil)
+(require 'git-timemachine)
```
jojojames: This should be

```
(require 'git-timemachine nil t)
```

for compilation purposes.

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,22 @@
...
+
+(defun ecgt--setup-bindings ()
+  "Setup `git-timemachine' since `evil-set-initial-state' is unavailable to
+minor modes."
+  (evil-motion-state nil)
```
jojojames: Instead of this, set the initial state in the setup function.

```
(evil-set-initial-state 'free-keys-mode 'normal)
```

^ free-keys example.

We prefer normal state to motion.

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,22 @@
...
+  (evil-local-set-key 'motion "n" #'git-timemachine-show-next-revision))
```
jojojames: We should use evil-define-key here instead (normal state). I posted a snippet somewhere in this PR.

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,22 @@
...
+
+(defun evil-collection-git-timemachine-setup ()
+  "Setup `evil' keybindings for `git-timemachine'."
+  (add-hook 'git-timemachine-mode-hook #'ecgt--setup-bindings))
```
jojojames: Can just have the binding code in here without a hook.

I think sometimes evil doesn't normalize keymaps so you can add a hook to git-timemachine-mode-hook to call evil-normalize-keymaps.

evil-collection-git-timemachine.el
jojojames: You can use the full namespace instead which becomes

evil-collection-git-timemachine--setup-bindings but as mentioned in another comment, you can just remove this function and set up the bindings in the original setup function.

jojojames:
You'll also need these two snippets.

```
(defvar free-keys-mode-map)

(defconst evil-collection-free-keys-maps '(free-keys-mode-map))

```

free-keys example^

jojojames:
> Additionally, if "n" and "p" are not the preferred Evil idioms, please let me
know, and I'll adopt whatever is recommended. It is convenient that "n" and "p"
are set my git-timemachine by default, since they also are the Vim way of
searching next and previous items...

I don't really have strong opinions here since I don't use the mode but maybe @Ambrevar might have more commentary.

We could probably add those n/p commands to C-j/C-k and ]/[ though.

evil-collection-git-timemachine.el
jojojames: Hmnn, totally glossed over the fact you said you tried this already.

jojojames: Hmnn, it's probably not needed either way. The mode will probably start up in normal state so you don't have to do anything special other than bind the keys in normal state.

evil-collection-git-timemachine.el
wpcarro: I'm not sure what normalizing keymaps does, but I'll be sure to look more into it!

evil-collection-git-timemachine.el
wpcarro: @jojojames I'll take your word for it! I'm less familiar with these requirements. I found prior art in other modules in this repo that use `(require 'blah)`, but I presume what you're proposing is more defensive.

jojojames: Yeah those are when the package is built-in.

wpcarro:
@jojojames Thanks so much for the feedback. It was very helpful. I went through everything, read the docs for the portions I didn't understand, and updated my implementation per your requests. The latest two commits should explain each change.

Thanks so much for the help!

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,40 @@
...
+;;; Code:
+(require 'evil)
+(require 'git-timemachine nil t)
+
+(defun evil-collection-git-timemachine-setup ()
+  "Setup `evil' keybindings for `git-timemachine'."
+  (evil-define-minor-mode-key 'normal 'git-timemachine-mode
+    "p" #'git-timemachine-show-previous-revision
+    "n" #'git-timemachine-show-next-revision
+    "g" #'git-timemachine-show-nth-revision
```
jojojames: Hmnn, since I don't really use git-timemachine, I'm not sure if 'g' is appropriate to take. I think one would one to use the 'g'* functions.

evil-collection-git-timemachine.el
```diff
@@ -0,0 +1,40 @@
...
+    "t" #'git-timemachine-show-revision-fuzzy
+    "q" #'git-timemachine-quit
+    "w" #'git-timemachine-kill-abbreviated-revision
```
jojojames: These 'w' keys seem a little emacs-y. Would x/X be appropriate for use?

ambihelical:
I've used git-timemachine quite a bit with evil, but through a hydra.  I can say without question you will sometimes want to search through the buffer.  So any bindings used with search should not be stolen. n is particularly bad, t is less bad.

jojojames:
@ambihelical Open to suggestions on keybindings. Nows a good time. :)

jojojames:
That makes sense.

@wpcarro 

Which keys look appropriate to you?

From the readme:

```
The following keys are free when insert state does not make sense in the current mode:

    a, A, i, I
    c, C, r, R, s, S
    d, D, x, X
    o, O
    p, P
    =, <, >
    J
    ~

    Any of those keys can be set to be a prefix key.

```

wpcarro:
@jojojames I'm AFK on vacation until this coming Monday. I'll be plugged back in then if you don't mind waiting!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jojojames:
@wpcarro Sounds good to me, take your time and enjoy your vacation!

jackkamm:
Was just about to submit a PR for git-timemachine when I saw this! My PR was going to use "g" and "z" as prefixes but I like the idea of using a single free prefix for discoverability. However I can't think of any strong reason to prefer a particular prefix. Maybe "i" because it's easy to reach and is the 2nd letter in "git" and "time"? Or "J" because git-timemachine involves "jumping" between revisions?

wpcarro:
@jackkamm I like the idea of using `C-n` and `C-p`! As for the other bindings, I still don't have an appealing answer.

Does anyone know what `evil-collection` thoughts are on Hydras? Do we want to avoid the dependency?

jackkamm:
How about this idea: `gt` prefix for the other bindings (aside from `C-n`/`C-p`/`q` for next/prev/quit). Looks free and is mnemonic.

jojojames:
gt looks good to me.

Yes, lets avoid hydras as that would be too opinionated.

wpcarro:
@jojojames @jackkamm thanks for the feedback. Gonna give `gt` a try. Standby for the PR updates.

wpcarro:
@jackkamm forgive me, but my Emacs keybinding knowledge always needs a little bit of help. I use `general.el` in my personal configuration. It looks like `C-p` is invoking `evil-paste`. What's the best way to override this? Or better yet, what's the best way to ensure that the KBDs in the `evil-define-minor-mode-key` have precedence over the evil commands?

Here's what I have:

```elisp
(defun evil-collection-git-timemachine-setup ()
  "Setup `evil' keybindings for `git-timemachine'."
  (evil-define-minor-mode-key 'normal 'git-timemachine-mode
    "C-p" #'git-timemachine-show-previous-revision
    "C-n" #'git-timemachine-show-next-revision
    "q"   #'git-timemachine-quit
    "gtg" #'git-timemachine-show-nth-revision
    "gtt" #'git-timemachine-show-revision-fuzzy
    "gtx" #'git-timemachine-kill-abbreviated-revision
    "gtX" #'git-timemachine-kill-revision
    "gtb" #'git-timemachine-blame))
```

I'd also love some suggestions for DRYing up the `gt` prefix stuff. It's working fine, but in `general` I can do something like:

```elisp
(general-define-key
  :states 'normal
  :prefix "gt"
  "g" #'git-timemachine-quit)
```

jackkamm:
You need to add to the setup function
```
(add-hook 'git-timemachine-mode-hook #'evil-normalize-keymaps)
```
Also I suggest "gty/Y" instead of "gtx/X", as `'git-timemachine-kill-revision` doesn't actually kill anything despite the name.

re: prefixes, it's possible to define a separate keymap with `(make-sparse-keymap)` and bind it to `gt`. But I would stick with your current approach, it seems simpler to me.

wpcarro:
@jackkamm even with `\C-p`, `evil-paste` is still attempted.

jackkamm:
Putting this in `evil-collection-git-timemachine.el` works for me, does it work for you?
```
(require 'evil)
(require 'git-timemachine)

(defconst evil-collection-git-timemachine-maps '(git-timemachine-mode-map))

(defun evil-collection-git-timemachine-setup ()
  "Setup `evil' keybindings for `git-timemachine'."
  (evil-define-minor-mode-key 'normal 'git-timemachine-mode
    "\C-p" #'git-timemachine-show-previous-revision
    "\C-n" #'git-timemachine-show-next-revision
    "q"   #'git-timemachine-quit
    "gtg" #'git-timemachine-show-nth-revision
    "gtt" #'git-timemachine-show-revision-fuzzy
    "gtx" #'git-timemachine-kill-abbreviated-revision
    "gtX" #'git-timemachine-kill-revision
    "gtb" #'git-timemachine-blame))

(provide 'evil-collection-git-timemachine)
```

wpcarro:
Hmm it's now working for me too... that's odd but somewhat relieving!

Pushing now. Merge at your leisure!

jojojames:
Hmnn, looking at the completed version, it seems the entire keymap is on a prefix.

I was thinking most of the keys could probably exist on some *key* given it's not too useful in this mode.

Here's the original keymap.

```
(define-minor-mode git-timemachine-mode
 "Git Timemachine, feel the wings of history."
 :init-value nil
 :lighter " Timemachine"
 :keymap
 '(("p" . git-timemachine-show-previous-revision) ;; ] / gj / c-j / m-j
   ("n" . git-timemachine-show-next-revision) ;; [ / gk / c-k / m-k
   ("g" . git-timemachine-show-nth-revision) ;; gn or gtn
   ("t" . git-timemachine-show-revision-fuzzy) ;; r
   ("q" . git-timemachine-quit) ;; q
   ("w" . git-timemachine-kill-abbreviated-revision) ;; x
   ("W" . git-timemachine-kill-revision) ;; X
   ("b" . git-timemachine-blame)) ;; B
 :group 'git-timemachine)
```

I've put some suggestions as comments.

We could probably still add all of these to a prefix too (so we'd have gt*).
I'm just guessing putting all the keys on the prefix just means no one is going to press those keys.

Since I don't actually use git-timemachine, I'll defer to @wpcarro @jackkamm @ambihelical on your final opinion.

jackkamm:
@jojojames I'm fine with your suggestion, except for "B", since this overrides a motion. Maybe "gb" instead?

jojojames:
>  "gb" is a possible alternative that's free.

Works for me.

>  (or maybe we can just bind to all of them?)

Yup, we can do that.

wpcarro:
@jojojames I appreciate the feedback! The functions from `git-timemachine` that I use the most frequently are not prefixed. These are `C-p`, `C-n`, and `q`.

I personally am fine with the other functions sitting behind the `gt` prefix. When combined with `which-key` and the mneumonic of `(g)it(t)imemachine`, I think it's reasonable.

I'm heeding @jackkamm's advice of swapping `y` for `x`, since this is really a yanking operation in Vim lingo. And unfortunately we will want `y` and `Y` to be free in case someone is just trying to copy code, which I do in `git-timemachine`.

Do you mind calling out specific bindings that you find problematic?

jojojames:
Could C-n/C-p be C-j/C-k to be more like the rest of evil-collection?

wpcarro:
@jojojames I have no problems changing it to that if that's what the `evil-collection` philosophy is! I changed it to `C-n`, `C-p` after @jackkamm shared the suggestion of using those bindings to navigate history.

Is there a compelling reason we should deviate from the guide?

wpcarro:
Doing `C-p` -> `C-k` and `C-n` -> `C-j`

jojojames:
Thanks, good stuff!

jackkamm:
@jojojames looks like git-timemachine was not added to `evil-collection-mode-list`, can you add it? Thanks and sorry to bother.

jojojames:
Sure, what's the 'provide for git-timemachine?

jackkamm:
@jojojames thanks! It's `git-timemachine`

PR #117, opened 2018-04-14
Optionally provide modes to evil-collection-init

noctuid:
This makes manually configuring things one mode at a time easier since `evil-collection-init` still does all the work.


Ambrevar:
Looks good to me.  @jojojames ?

PR #119, opened 2018-04-21
Add integration for lsp-ui-imenu

Dickby:


Dickby:
This pull request just adds bindingf for `lsp-ui-imenu--prev-kind` and `lsp-ui-imenu-next-kind`

jojojames:
My guess is your fork is not rebased against 'upstream' (emacs-evil/evil-collection) so your pull request contains a diff of that difference to upstream.

Easiest way might be to remove your fork, re-fork evil-collection, add that change again and then PR.

More correct change is to add a remote to upstream repo (magit can do this easily), rebase against upstream/master to your origin/master, add the commit and PR.

PR #123, opened 2018-05-09
Add support for dd in wgrep buffers

terlar:
This seems to be the only way to define `dd` without losing the functionality of
the rest of the `d`-prefixed keys. This should allow regular `d` operations as
well as `dd` to remove the target line.

This is just a copy of the `evil-delete` operator from [evil source](https://github.com/emacs-evil/evil/blob/c3317f8d1bd60405a2c3aad372da126651a74fbe/evil-commands.el#L1348-L1374). And
builds upon what I mentioned earlier in:

https://github.com/emacs-evil/evil-collection/pull/111

### DIFF
--- a/evil-collection-wgrep.el
+++ b/evil-collection-wgrep.el
@@ -34,12 +34,44 @@
 
 (defconst evil-collection-wgrep-maps '(wgrep-mode-map))
 
+(evil-define-operator evil-collection-wgrep-delete (beg end type register yank-handler)
+  "Delete text from BEG to END with TYPE.
+Save in REGISTER or in the kill-ring with YANK-HANDLER."
+  (interactive "<R><x><y>")
+  (unless register
+    (let ((text (filter-buffer-substring beg end)))
+      (unless (string-match-p "\n" text)
+        ;; set the small delete register
+        (evil-set-register ?- text))))
+  (let ((evil-was-yanked-without-register nil))
+    (evil-yank beg end type register yank-handler))
+  (cond
+   ((eq type 'block)
+    (evil-apply-on-block #'delete-region beg end nil))
+   ((eq type 'line)
+    (if (and (= end (point-max))
+             (or (= beg end)
+                 (/= (char-before end) ?\n))
+             (/= beg (point-min))
+             (=  (char-before beg) ?\n))
+        (delete-region (1- beg) end)
+      (dotimes (_ (count-lines beg end))
+        (wgrep-mark-deletion)
+        (evil-next-line))))
+   (t
+    (delete-region beg end)))
+  ;; place cursor on beginning of line
+  (when (and (called-interactively-p 'any)
+             (eq type 'line))
+    (evil-first-non-blank)))
+
 (defun evil-collection-wgrep-setup ()
   "Set up `evil' bindings for `wgrep'."
   (evil-define-key nil wgrep-mode-map
     [remap evil-write] 'wgrep-finish-edit)
 
   (evil-define-key 'normal wgrep-mode-map
+    "d" 'evil-collection-wgrep-delete
     "ZQ" 'wgrep-abort-changes
     "ZZ" 'wgrep-finish-edit
     (kbd "<escape>") 'wgrep-exit))


Ambrevar:
Thanks for your contribution!

Note that you don't need to redefine an operator for that sort of things.  Look at `evil-collection-vdiff-setup`, it does exactly what you want.  Can you adapt your pull request using that method?  It's shorter, more maintainable, more flexible.
Let me know if you need more help.

terlar:
Thank you for your feedback, it is not clear to me exactly how that works. Would be really nice to have a solution that works without copying an internal evil method.

Is this `menu-item` with filter part some kind of hack to allow this operator only to work for `d`?

I tried defining this, but no luck:

```
(evil-define-minor-mode-key 'operator 'wgrep-mode
      "d" '(menu-item
            ""
            nil
            :filter (lambda (&optional _)
                      (when (memq evil-this-operator
                                  evil-collection-delete-operators)
                        #'wgrep-mark-deletion))))
```

How is this working? Any hints are welcome, thanks.

Ambrevar:
It does not work because wgrep-mode does not exist: wgrep is not a minor mode.

The following seems to work:
```
(evil-define-key 'operator wgrep-mode-map
    "u" '(menu-item
          ""
          nil
          :filter (lambda (&optional _)
                    (when (memq evil-this-operator
                                evil-collection-delete-operators)
                      (setq evil-inhibit-operator t)
                      #'wgrep-mark-deletion))))
```
It does not if I replace "u" with "d" however... Not sure why.

I don't know about the internals of Evil, so I can't explain about `menu-item`.  Maybe @jojojames / @noctuid ?

noctuid:
`dd`, `yy`, etc. are all implemented specially. For example, if you define `d` in the global operator state to anything, it will work with `y` but not work with or break `d`. I think the `menu-item` approach is (probably) not suitable here for that reason. `general-key-dispatch` works for this case, but like I've said before, I don't think it's a good idea to use its approach in a package. Redefining `evil-delete` works but is not ideal. I've never encountered this situation personally but I may have encountered a package with an alternative workaround (not sure). I'd have to think about whether there is a better solution.

jojojames:
Curious if there's anything we can do to move this forward? I'm not too sure what the next step is. @terlar

noctuid:
Well it's possible to do, but the solution isn't nice. If you think this is useful enough to warrant it, it might make sense to add support for doing something like this in evil directly. `evil-operator-range` is responsible for the special behavior of `dd` and other operators. It binds the key that the operator is bound to to `evil-line` in `evil-operator-shortcut-map`. One workaround would be to advise `evil-operator-range`, so that it binds the key to a menu item that will dispatch to the correct definition. This is not at all an ideal solution and wouldn't compose very well if you wanted to do this in more than one mode (you'd have to have one shared menu item for all modes/keybindings). Another possible solution that doesn't require advice/overriding `define-key` would be to give a map precedence over `evil-operator-shortcut-map`. That could be done like this:

```elisp
(evil-put-property
 'evil-state-properties
 'operator
 :enable 
 (cons 'some-map (evil-get-property evil-state-properties 'operator :enable)))
```

You could then bind the menu item in that keymap. This keymap will always be checked (regardless of whether it would otherwise be active), so you'd have to make `some-map` local and bind keys in it using `wgrep-setup-hook` to prevent the need to chek wgrep mode and fix the one universal menu item problem. That's not too terrible, but it's way more hacky than just using `evil-define-key` with an extended menu item.

noctuid:
Here's a basic example of what I meant (doesn't work as expected):
```elisp
;; create map
(defvar-local some-mode-delete-map nil)
(put 'some-mode-delete-map 'permanent-local t)

;; enable map in operator state
(evil-put-property
 'evil-state-properties
 'operator
 :enable
 (cons 'some-mode-delete-map
       (evil-get-property evil-state-properties 'operator :enable)))

;; bind key to menu item
(defun some-mode-dd-setup ()
  (setq some-mode-delete-map (make-sparse-keymap))
  (define-key some-mode-delete-map
    "d" '(menu-item
          ""
          nil
          :filter (lambda (&optional _)
                    (when (memq evil-this-operator
                                evil-collection-delete-operators)
                      ;; NOTE: this breaks other d<key> functionality,
                      ;; but without it, get unwanted deletions
                      ;; (setq evil-inhibit-operator t)
                      #'new-dd-command)))))
(add-hook 'some-mode-hook #'some-mode-dd-setup)

;; for testing in scratch buffer with random command:
(setq some-mode-delete-map (make-sparse-keymap))
(define-key some-mode-delete-map
  "d" '(menu-item
        ""
        nil
        :filter (lambda (&optional _)
                  (when (memq evil-this-operator
                              evil-collection-delete-operators)
                    #'info))))
```

Creating a new state for this purpose would be more ideal (I didn't originally realize this worked), but it has the same issue with `evil-inhibit-operator`. If this isn't considered a hack/abuse of states, and there is a workaround to the `evil-inhibit-operator` issue, it might be worth making a PR to evil adding a new state for this purpose. It may be an easy fix; I haven't looked at it yet.
```elisp
(evil-define-state operator-shortcut
  "Operator shortcut state with global keymap for user use.")

(evil-put-property
 'evil-state-properties
 'operator
 :enable
 (cons 'operator-shortcut
       (evil-get-property evil-state-properties 'operator :enable)))

;; example
(evil-define-key 'operator-shortcut emacs-lisp-mode-map
  "d" '(menu-item
        ""
        nil
        :filter (lambda (&optional _)
                  (when (memq evil-this-operator
                              evil-collection-delete-operators)
                    ;; same issue
                    ;; (setq evil-inhibit-operator t)
                    #'info))))
```

noctuid:
It looks like the menu item code is run when first entering operator state. I didn't look into why this happens (probably happens during `(evil-change-state 'operator)`; happens for enabled maps/states other than `operator` state itself it seems). Here's a basic workaround that you can test further (only inhibits operator state if the definition is actually used):
```elisp
(evil-define-state...
...
(defmacro my-inhibit-operator (command)
  "Return a command that inhibits evil operator code."
  `(lambda ()
     (interactive)
     (setq evil-inhibit-operator t)
     (call-interactively ,command)))


(evil-define-key 'operator-shortcut emacs-lisp-mode-map
  "d" '(menu-item
        ""
        nil
        :filter (lambda (&optional _)
                  (when (memq evil-this-operator
                              evil-collection-delete-operators)
                    (my-inhibit-operator #'info)))))
```
This works as expected for me in all the basic cases I tried.

PR #124, opened 2018-05-09
Use evil-ex-search when requested in info mode

### LINKED ISSUE #122: Discussion of isearch in info-mode
In info mode, `n/p` is bound to the isearch next/previous commands [currently](https://github.com/emacs-evil/evil-collection/blob/master/evil-collection-info.el#L74). 

However, `/` is not bound to evil-ex-search, and this means that normal evil searches do not have a working next/previous.

I wanted to bring up a larger point though, I personally don't like isearch and prefer to use the default evil ex search behavior everywhere (I have it customized and I want to keep those customizations without duplicated them for isearch). 

Would it sound OK to add a `evil-collection-want-ex-search` to turn off isearch bindings? Does anyone else have thoughts on better solutions?

Is there a reason for bidning to isearch at all, when evil default search works fine? In this case, it might search across pages, but I personally find that more confusing. :P

jgkamat:
I also added "folding" to the emms binds, although, it's not a perfect match (there's no "collapse one tree" command, and "expand-one-level" run repeatedly breaks the tree). For some reason, isearch in emms dosen't seem to search all folds for me, but I don't mind too much, so I'm not touching that for now.

Let me know if anything seems wrong. 

Closes #122

### DIFF
--- a/evil-collection-evil-search.el
+++ b/evil-collection-evil-search.el
@@ -31,6 +31,8 @@
 ;;; Code:
 (require 'evil-search)
 
+(defvar evil-collection-evil-search-enabled (eq evil-search-module 'evil-search))
+
 (defvar evil-collection-evil-search-forward
   '(menu-item "" nil :filter (lambda (&optional _)
                                (if (eq evil-search-module 'evil-search)
--- a/evil-collection-evil-search.el
+++ b/evil-collection-evil-search.el
@@ -31,7 +31,8 @@
 ;;; Code:
 (require 'evil-search)
 
-(defvar evil-collection-evil-search-enabled (eq evil-search-module 'evil-search))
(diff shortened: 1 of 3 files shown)


evil-collection-evil-search.el
```diff
@@ -31,6 +31,8 @@
 ;;; Code:
 (require 'evil-search)
 
+(defvar evil-collection-evil-search-enabled (eq evil-search-module 'evil-search))
```
Ambrevar: Hmm, I think this should rather be a function to make it dynamic.  Otherwise, if evil-search-module changes it won't be reflected after definition of `evil-collection-evil-search-enabled`.

Ambrevar:
Looks good, please convert the defvar to a function and I'll merge.

PR #128, opened 2018-05-19
Force isearch in pdf mode

jgkamat:
For PDF files, the default evil search dosen't seem to work at all (for me), so I think it's better to just force isearch on here so people have something.

I also made gg/G scroll to the absolute start/end of the document by scrolling to the top/bottom of the image after navigating to the correct page. This fixes scrolling down a couple lines and then pressing `gg`


Ambrevar:
Thanks!

PR #131, opened 2018-05-29
Abort company search with single <escape>

wasamasa:
This originates from my init file. There are two keymaps in company where one can abort, but by default with "\e\e\e" instead of a single hit. Curiously enough in Evil a single hit works out of the box for regular completion candidates, but in the search it doesn't.  This PR fixes it for the latter.


jojojames:
Looks good!

Issue #129, opened 2018-05-26
Move ediff to evil-collection


Emiller88:
Just wondering why evil-ediff hasn't been added here. 

If no one sees any issue I'll open a PR.

jojojames:
Related to https://github.com/emacs-evil/evil-collection/issues/28

We can revisit this topic if there's enough interest. As of right now, I'm ok with leaving the packages separate but there's a couple options to bundle them together (requiring evil-ediff when installing evil-collection, etc).

Not sure how others feel about this though.

Emiller88:
That was what I was looking for was that there was already a conversation. I was just interested why there was an evil-collection-magit, and an evil-magit.

I probably don't understand the history of the emacs-evil org.

jojojames:
Yeah, I think it'd be pretty cool if there was just one package users could import (evil-collection) and all their evil-integration woes would be handled. In that manner, probably requiring (adding those packages as dependencies might be interesting). I haven't really thought too much of the downsides though.

I'd be curious what @Ambrevar and @noctuid thinks.

noctuid:
Because `evil-collection` intends to provide consistent keybindings across modes, I think it makes sense to have all modes supported, at least eventually. In some cases, it might make sense to ask other package authors to merge their packages into evil-collection. Alternatively, evil-collection could just depend on them like you suggest and potentially reconcile any inconsistent keybindings. For packages that only make keybindings and don't do any other form of integration, this would be less necessary.

One thing to consider is whether other packages make sense to integrate here, at least in their entirety (specifically I'm thinking about evil-org). Ediff and mu4e support definitely makes sense. I don't think it would make sense to integrate evil-lispy or lispyville at all. Finally, org agenda keybindings definitely make sense, but org mode keybindings don't fit with the general theme (keybindings in non-editing modes where a lot of letter keys can be rebound), so only some of evil-org's functionality is relevant to evil-collection.

Ambrevar:
I don't see much of a problem with evil-org: after all, many other modes supported by Evil Collection do not bear the same "consistency" because of their inherent structure that is too different (Helm, diff, etc.).

I don't know about evil-lispy, but lispyville seems to be a tad unfinished as of now.  We've been discussing binding themes and much of the bindings seem to be extremely opinionated.  Much more than the rest of Evil Collection.

At last, evil-mu4e is also unfinished in my opinion but much more doable.  I've been working on polishing it and I think I'll send a patch soon.

[...]

Issue #21, opened 2017-11-08
Customizable initial/primary state


noctuid:
I prefer to never actually enter motion state and only use the keymap for actual motions.

Ambrevar:
Could you be more specific?  What is your prefered state?  Use case?  Which modes?

Evil motion state is meant for modes where text insertion is not possible.

noctuid:
I always use normal state everywhere. My global keybindings are all made in normal state. A couple of the default keybindings in normal state also make sense when text insertion isn't possible (e.g. yanking and macro recording). For me, using motion state just makes things unnecessarily complicated because now I have to bind things that I would normally bind in just normal state in both normal in motion state. That or I'd have to bind all these keys in motion state and unbind some of the defaults in normal state.

jojojames:
Is there a list of tradeoffs for motion vs normal state?

I'd prefer not only to make this customizable but actually default to normal state if there weren't any obvious downsides.

It makes the transition easier for modes that can technically be in motion state but can be toggled into an editing state. Occur/Ag/Rg/Dired/etc are some that are like that.

Ambrevar:
The main downside of using normal state for read-only modes is that _all the insertion keys_ (and that's lots of them) have to be unbound/rebound.

As for the global question, off the top of my head there is some inheritance at work, but I have to look into this.
Otherwise, if your key is truly global (i.e. also for insertion), use `global-set-key`, and if your key should be global in both normal and motion state, it's trivial to do so with `evil-global-set-key`.

I'm against switching to normal state because that's not how Evil was designed.  The motion state is here for a purpose.

I'll investigate further.

jojojames:
I'm not sure I follow why they have to be unbound.

For the case where there's a key "under" evil's normal state key, we would've bounded it in our package anyway.

For the case where there's no key "under" evil's normal state key, there's no loss other than the user *maybe* pressing 'i' in a readonly buffer and getting some sort of message.

> I'm against switching to normal state because that's not how Evil was designed. The motion state is here for a purpose.

Lets evaluate based on the tradeoffs of both approaches. I digress, I believe I read somewhere from the previous evil maintainer that evil-define-key itself wasn't a great design and using something like buffer-hooks and evil local keymaps is probably the better approach.

noctuid:
> The main downside of using normal state for read-only modes is that all the insertion keys (and that's lots of them) have to be unbound/rebound.

[...]

PR #136, opened 2018-06-01
Ediff: Merge from evil-ediff.el

justbur:
Adds evil-collection-ediff.el

Ref #129 

Please check to make sure I followed the conventions of the repo properly.

### DIFF
--- a/evil-collection-ediff.el
+++ b/evil-collection-ediff.el
@@ -1,0 +1,205 @@
+;;; evil-collection-ediff.el --- Evil bindings for ediff -*- lexical-binding: t -*-
+;; Copyright (C) 2015 Justin Burkett
+
+;; Author: Justin Burkett <justin@burkett.cc>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <ambrevar@gmail.com>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Homepage: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1") (evil "1.2.3"))
+;; Keywords: evil, ediff, tools
+
+;; This file is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published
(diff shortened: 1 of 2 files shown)


jojojames:
This looks good other than potentially the motion vs normal state default.

Ok to merge for me. @Ambrevar ?

Ambrevar:
Should we change to normal state now or after the merge?
Switching to normal state can be tricky sometimes since many keys don't fallback to their Emacs-state equivalent the same way motion-state does.

justbur:
What is the issue with motion state? I think motion state makes sense in buffers that are not intended to be edited, like the ediff control buffer.

jojojames:
https://github.com/emacs-evil/evil-collection/issues/21

I think we reached a consensus to use normal state by default.

justbur:
I just skimmed that thread. If normal state is the default, then that's what it should be. I don't think it will have a big effect here.

noctuid:
Since that discussion was pretty long, it may make sense to summarize the reasoning somewhere. To summarize my points from that discussion, motion state is:
- confusing to many users (new users and users coming from vim; e.g. many users don't know when they should use the motion keymap and just bind their keys in normal, visual, and motion state, don't know about motion state and don't understand why their normal state keybindings don't work in buffers that start in motion state, etc.)
- poorly named (doesn't only contain motions)
- an incomplete solution to a minor issue that unnecessarily complicates things

I personally haven't had issues with accidental keypresses, but motion doesn't solve this issue for all non-editable buffers anyway (e.g. motions don't make sense in the ediff control buffer). A more complete solution would be to add even more states, which would be even more convoluted. I think using normal state as the default state everywhere and just rebinding keys to `ignore` where it makes sense is a simpler solution.

Ambrevar:
Let's merge (and change to normal state later)?

justbur:
I just made the change to normal state and rebased

Ambrevar:
Thanks!

justbur:
@bling `q` quits for me. I can't reproduce what you're seeing

bling:
@justbur mystery solved.  i also had the old `evil-ediff` package loaded.  thanks.

PR #137, opened 2018-06-02
add a count to pdf-tool's next-line-or-next-page

concavegit:
For some reason the count arguments for `pdf-view-next-line-or-next-page` and `pdf-view-previous-line-or-previous-page` only work for `<down>` and `<up>`, but not `j` and `k`. As in, `2 <down>` scrolls down by 2 pages as expected, but `2 j` scrolls down by 2 pages.

After a bit of experimentation, the best solution so far is wrapping the functions in a `dotimes`. After that fix, and both worked for all of the mentioned keys.

### DIFF
--- a/evil-collection-pdf.el
+++ b/evil-collection-pdf.el
@@ -46,6 +46,22 @@
 (defvar pdf-outline-buffer-mode-map)
 (defvar pdf-occur-buffer-mode-map)
 
+(defun evil-collection-pdf-view-next-line-or-next-page (&optional count)
+  "'evil' wrapper include a count argument to `pdf-view-next-line-or-next-page'"
+  (interactive "P")
+  (if count
+      (dotimes (_ count nil)
+	(pdf-view-next-line-or-next-page 1))
+    (pdf-view-next-line-or-next-page 1)))
+
+(defun evil-collection-pdf-view-previous-line-or-previous-page (&optional count)
+  "'evil' wrapper include a count argument to `pdf-view-previous-line-or-previous-page'"
+  (interactive "P")
+  (if count
+      (dotimes (_ count nil)
+	(pdf-view-previous-line-or-previous-page 1))
+    (pdf-view-previous-line-or-previous-page 1)))
+
 (defun evil-collection-pdf-view-goto-page (&optional page)
   "`evil' wrapper around `pdf-view-last-page'."
   (interactive "P")
@@ -69,8 +85,8 @@
   (evil-define-key 'normal pdf-view-mode-map
     ;; motion
     (kbd "<return>") 'image-next-line
-    "j" 'pdf-view-next-line-or-next-page
-    "k" 'pdf-view-previous-line-or-previous-page
+    "j" 'evil-collection-pdf-view-next-line-or-next-page
+    "k" 'evil-collection-pdf-view-previous-line-or-previous-page
     (kbd "SPC") 'pdf-view-scroll-up-or-next-page
     (kbd "S-SPC") 'pdf-view-scroll-down-or-previous-page
     (kbd "<delete>") 'pdf-view-scroll-down-or-previous-page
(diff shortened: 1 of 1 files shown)


jojojames:
I wonder why it doesn't work for j and k. It might be good to rootcause this first. 

Otherwise, I'll leave to @Ambrevar .

concavegit:
I think the issue is in `evil`, as `M-2 n` (the emacs binding for `pdf-view-next-line-or-next-page` with a number argument) works without modification. I am unconfident about my ability to find the root cause at this point, so I will probably leave it here.

jojojames:
Thanks, this is an improvement over the previous so I'll merge it for now. If we can find the root cause at some point in the future, we can better fix it then.

I think this is merged along with your other change so I'll close this one.

jojojames:
Actually, not sure if this got merged in or not. Can you double check?

concavegit:
I do not think it is merge, as there is no `(defun evil-collection-pdf-view-next-line-or-next-page ..` in the master copy.

concavegit:
Would it help for me to put a TODO comment by the defuns to indicate a search for a root cause?

jojojames:
Yes, that would be awesome.

I was going to merge it now but I'll wait for your todo comment commit.

concavegit:
Added the TODO with a link to this thread.

jojojames:
Thanks!

Ambrevar:
Sorry for being late on this.  I actually could not reproduce the issue.
I'll try again but it might have to do with @concavegit's config.  @jojojames, can you reproduce?

Ambrevar:
@concavegit: I just noticed that you've used tabs.  Please stick to spaces only :)

Ambrevar:
Also in terms of code it would have been simpler to remove the `if` and set count to 1 by default:
```
(setq count (or count 1))
```

That said, let's try to hunt down the root cause first! ;)

concavegit:
It is worth noting that it looks like the changes are already in master, so I cannot reproduce without changing the lines back and recompiling. That being said, here is my config: https://gitlab.com/concavegit/dotfiles/blob/master/emacs/.emacs.d/init.el. I will try your above init.el as well.

concavegit:
Cutting away irrelevant stuff, here is a smaller config which maintains the problem on my machine:

```elisp
(require 'package)
(package-initialize)

(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
(add-to-list 'package-archives '("org" . "http://orgmode.org/elpa/"))

(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

(require 'use-package)

(setq use-package-always-ensure t)

(use-package evil
  :init
  (setq evil-want-integration nil)
  :config
  (evil-mode 1))

(use-package evil-collection :after evil :config (evil-collection-init))
(use-package pdf-tools
  :mode ("\\.pdf$" . pdf-view-mode)
  :config
  (pdf-tools-install))
```
Make sure the lines in `evil-collection-pdf.el` with `evil-collection-pdf-view-(next|prev)..` are changed back to `pdf-view-(next|prev)..` and recompile the directory.

concavegit:
I will try this in an ubuntu vm to make sure its not anything else being funky.

Ambrevar:
A VM is a little overkill :p Simply load with `emacs -Q -l my-barren-init.el` and you should be good.
Also what's your `M-x emacs-version`?

concavegit:
26.1, though I have needed something similar in my config since at least 25, I think

Ambrevar:
Can you try on a fresh ELPA cache?
(Either set `package-user-dir` to a different path before package.el initialization or rename your old folder.)

Ambrevar:
Also can you try without use-package?

concavegit:
Just making sure: is your local copy of evil-collection pre-this merge?

Ambrevar:
I use a git checkout and tried both before and after this merge: can't reproduce.

Your example above does not count as a clean checkout: you should reinstall those 3 packages with your current Emacs to make sure there is not issue with byte-code mismatch.

concavegit:
Alright, I have deleted my `~/.emacs.d/elpa` and have reinstalled pdf-tools and evil, while loading commit 050af5d for evil-collection and am getting the same problem. Should we just leave this here as a mystery and leave this here for now? I think at this point if there is a resolution it is something far too simple to see at this point.

Ambrevar:
Thanks for trying all this.  Maybe the mistake is on my end.
What do @jojojames and @noctuid think?  Can you reproduce?

concavegit:
No problem! To be clear, I would not be suprised at all if the mistake was mine the whole time.

concavegit:
I am just suprised that we have not figured out why I am at fault yet :smile:

jojojames:
@Ambrevar Sorry, I thought it was reproducible since there were no comments!

Feel free to do as you see fit.

PR #143, opened 2018-06-05
Fix #135: Whitelist, blacklist and binding record

### LINKED ISSUE #135: New Feature: Reserved keys
It would be nice if I could provide a list of keys in states up front that wouldn't be overwritten by this package. This is most useful if you have a non-standard leader key. I know I can correct the bindings after loading the package, but then I have to go and find every conflict, which gets annoying. 

I'm imagining writing a custom `evil-collection-evil-define-key` that filters the keys before making bindings. I could write one.

justbur:
I implemented a `evil-collection-define-key` function to support whitelists and blacklists. I also added the ability to record and print key bindings made by this package. The print function is a little rudimentary, and this should be tested before it is merged, but it seems to work fine for me so far.

### DIFF
--- a/evil-collection-ag.el
+++ b/evil-collection-ag.el
@@ -35,7 +35,7 @@
 
 (defun evil-collection-ag-setup ()
   "Set up `evil' bindings for `ag'."
-  (evil-define-key '(normal visual) ag-mode-map
+  (evil-collection-define-key '(normal visual) 'ag 'ag-mode-map
     "k" 'evil-previous-line
     "h" 'evil-backward-char
 
--- a/evil-collection-ag.el
+++ b/evil-collection-ag.el
@@ -35,7 +35,7 @@
 
 (defun evil-collection-ag-setup ()
   "Set up `evil' bindings for `ag'."
-  (evil-collection-define-key '(normal visual) 'ag 'ag-mode-map
+  (evil-collection-define-key '(normal visual) 'ag-mode-map
     "k" 'evil-previous-line
     "h" 'evil-backward-char
 
--- a/evil-collection-company.el
+++ b/evil-collection-company.el
@@ -54,24 +54,26 @@ be set through custom or before evil-collection loads."
 
 (defun evil-collection-company-setup ()
   "Set up `evil' bindings for `company'."
-  (define-key company-active-map (kbd "C-n") 'company-select-next-or-abort)
-  (define-key company-active-map (kbd "C-p") 'company-select-previous-or-abort)
-  (define-key company-active-map (kbd "C-j") 'company-select-next-or-abort)
-  (define-key company-active-map (kbd "C-k") 'company-select-previous-or-abort)
-  (define-key company-active-map (kbd "M-j") 'company-select-next)
-  (define-key company-active-map (kbd "M-k") 'company-select-previous)
+  (evil-collection-define-key nil 'company-active-map
+    (kbd "C-n") 'company-select-next-or-abort
+    (kbd "C-p") 'company-select-previous-or-abort
+    (kbd "C-j") 'company-select-next-or-abort
+    (kbd "C-k") 'company-select-previous-or-abort
+    (kbd "M-j") 'company-select-next
+    (kbd "M-k") 'company-select-previous)
 
   (when evil-want-C-u-scroll
-    (define-key company-active-map (kbd "C-u") 'company-previous-page))
+    (evil-collection-define-key nil 'company-active-map (kbd "C-u") 'company-previous-page))
 
   (when evil-want-C-d-scroll
-    (define-key company-active-map (kbd "C-d") 'company-next-page))
+    (evil-collection-define-key nil 'company-active-map (kbd "C-d") 'company-next-page))
 
-  (define-key company-search-map (kbd "C-j") 'company-select-next-or-abort)
-  (define-key company-search-map (kbd "C-k") 'company-select-previous-or-abort)
-  (define-key company-search-map (kbd "M-j") 'company-select-next)
-  (define-key company-search-map (kbd "M-k") 'company-select-previous)
-  (define-key company-search-map (kbd "<escape>") 'company-search-abort)
+  (evil-collection-define-key nil 'company-search-map
+    (kbd "C-j") 'company-select-next-or-abort
+    (kbd "C-k") 'company-select-previous-or-abort
+    (kbd "M-j") 'company-select-next
+    (kbd "M-k") 'company-select-previous
+    (kbd "<escape>") 'company-search-abort)
 
   ;; Sets up YCMD like behavior.
   (when evil-collection-company-use-tng (company-tng-configure-default)))
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -182,9 +182,8 @@ This is a list of strings that are suitable for input to `kbd'."
 (defvar evil-collection--bindings-record (make-hash-table :test 'eq)
   "Record of bindings currently made by Evil Collection. This is
 a hash-table with the package symbol as a key.  The associated
-values are the package's bindings which are stored as a
-hash-table with key being the key to be bound and value as the
-binding.")
+values are the package's bindings which are stored as a list of
+the form ((STATE KEY BINDING)).")
 
 (defvar evil-collection-setup-hook nil
   "Hook run by `evil-collection-init' for each mode that is evilified.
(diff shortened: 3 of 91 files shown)


evil-collection.el
```diff
@@ -166,6 +166,24 @@ this confusing. It will be included if
   :type '(repeat (choice symbol sexp))
   :group 'evil-collection)
 
+(defcustom evil-collection-key-whitelist '()
+  "List of keys that may be used by evil-collection.
```
Ambrevar: Spell it "Evil Collection"?

evil-collection.el
```diff
@@ -166,6 +166,24 @@ this confusing. It will be included if
...
+This is a list of strings that are suitable for input to
+`kbd'. If there are no keys in the list, the whitelist will be ignored."
```
Ambrevar: GNU style is to put double-spacing between sentences.

evil-collection.el
```diff
@@ -166,6 +166,24 @@ this confusing. It will be included if
...
+  :type '(repeat string)
+  :group 'evil-collection)
+
+(defcustom evil-collection-key-blacklist '()
+  "List of keys that may not be used by evil-collection.
+This is a list of strings that are suitable for input to `kbd'."
+  :type '(repeat string)
+  :group 'evil-collection)
+
+(defvar evil-collection-bindings-record '()
```
Ambrevar: Name `evil-collection--bindings-record` because it should be private?

evil-collection.el
```diff
@@ -175,6 +193,55 @@ Evil Collection for that mode. More arguments may be added in the future, so
...
+  (let* ((whitelist (mapcar 'kbd evil-collection-key-whitelist))
+         (blacklist (mapcar 'kbd evil-collection-key-blacklist))
+         (record (cdr-safe (assoc package evil-collection-bindings-record))))
+    (while bindings
+      (let ((key (pop bindings))
+            (def (pop bindings)))
+        (when (or (and whitelist (member key whitelist))
+                  (not (member key blacklist)))
+          (push (cons key def) record)
+          (eval-after-load package
```
Ambrevar: Why not using `with-eval-after-load`?

evil-collection.el
```diff
@@ -175,6 +193,55 @@ Evil Collection for that mode. More arguments may be added in the future, so
...
+            `(condition-case err
+                 (evil-define-key* ',state ,map-sym ,key ',def)
+               (error
+                (message "evil-collection: Bad binding %s"
+                         '(evil-define-key* ',state ,map-sym ,key ',def))))))))
+    (setq record (nreverse record))
+    (if (assoc package evil-collection-bindings-record)
+        (setcdr (assoc package evil-collection-bindings-record) record)
+      (push (cons package record) evil-collection-bindings-record))
```
Ambrevar: What about using a hashmap instead of an alist?  It would simplify this `if` as well as the code above a little bit.  I think it's also more readable.

evil-collection.el
```diff
@@ -175,6 +193,55 @@ Evil Collection for that mode. More arguments may be added in the future, so
...
+| Key | Definition |
+|-----|------------|
+")
+        (dolist (binding (cdr package))
+          ;; Don't print nil definitions
+          (when (cdr binding)
+            ;; FIXME: Handle keys with | in them
+            (insert (format "| %s | %S |\n"
+                            (key-description (car binding)) (cdr binding)))))
+        (org-table-align)))))
```
Ambrevar: What about a tabulated-list instead of Org mode?

Ambrevar:
Excellent job!  Thank you so much for this!
(I haven't tested it yet though :p)

evil-collection.el
justbur: You mean refer to the package as Evil Collection in all docstrings?

evil-collection.el
justbur: ok

evil-collection.el
justbur: I can. I was in the habit of using the older function internally.

evil-collection.el
justbur: I can do that. I figured since I iterate over the list later, a list makes a little more sense, but a hashmap is fine too.

evil-collection.el
justbur: I thought about that. I was thinking of this and related functions as being akin to automatically generated readmes, hence org mode. I think there's also better familiarity with org-mode in general, but I don't have a strong opinion here.

evil-collection.el
justbur: ok, but two spaces are not being used consistently in the current package.

Ambrevar: We should fix this.

evil-collection.el
Ambrevar: I think so, the README uses "Evil Collection" as a title.

evil-collection-ag.el
```diff
@@ -35,7 +35,7 @@
 
 (defun evil-collection-ag-setup ()
   "Set up `evil' bindings for `ag'."
-  (evil-define-key '(normal visual) ag-mode-map
+  (evil-collection-define-key '(normal visual) 'ag 'ag-mode-map
```
jojojames: Very minor but I think it's a little regrettable for the extra 'package' parameter when most keymaps (I know some don't) are 1:1 with the package name. (i.e. ag-mode-map -> ag).

Not a blocker or anything, just thought I'd comment.

jojojames:
This looks good to me. More eyeballs on it would be good too.

evil-collection.el
justbur: I actually find it a little more straightforward to use `eval-after-load` here, because `with-eval-after-load` is a macro and doesn't eval its arguments.

justbur:
@Ambrevar I made the changes in fedbea8

evil-collection-ag.el
justbur: That's true for many packages, but certainly not all. See ivy, helm, pdf-tools, etc.

Another thought I had was to have a mapping somewhere between keymaps and conataining packages, which might be useful information to keep track of. 

Before you were using `evil-define-key` which uses some tricks to defer loading bindings until a map is defined (without needing to know the package), but I don't think it's a good idea to use this functionality for evil-collection. It can be tricky to debug when something goes wrong, and I think it's a good idea to be explicit about which keymap is where in case something changes upstream and it goes uncaught because `evil-define-key` hides what's going on in a sense.

justbur:
ok, I removed the package arg to `evil-collection-define-key` and just re-implemented the `evil-define-key` functionality. I realized it's probably not going to be used for many packages anyway, since you are deferring the bindings until after the package is loaded.

Dropping the package arg means that when the bindings are printed they will be organized according to keymap instead of package, but that might be better anyway. 

See 439b9ff

evil-collection-company.el
```diff
@@ -54,24 +54,26 @@ be set through custom or before evil-collection loads."
 
 (defun evil-collection-company-setup ()
   "Set up `evil' bindings for `company'."
-  (define-key company-active-map (kbd "C-n") 'company-select-next-or-abort)
-  (define-key company-active-map (kbd "C-p") 'company-select-previous-or-abort)
-  (define-key company-active-map (kbd "C-j") 'company-select-next-or-abort)
-  (define-key company-active-map (kbd "C-k") 'company-select-previous-or-abort)
-  (define-key company-active-map (kbd "M-j") 'company-select-next)
-  (define-key company-active-map (kbd "M-k") 'company-select-previous)
+  (evil-collection-define-key nil 'company-active-map
```
jojojames: I'm curious if this keeps the same keymap precedence before. I think I used define-key because other keys were still taking over, over evil keymaps.

jojojames:
This looks good for me as far as I can tell. @Ambrevar ?

Ambrevar:
Question: is it necessary to use a keymap symbol instead of the map itself?
If not, I think we should use the map for consistency with other Emacs functions.
If it is, then can you document why?

justbur:
> Question: is it necessary to use a keymap symbol instead of the map itself?
If not, I think we should use the map for consistency with other Emacs functions.
If it is, then can you document why?

`define-key` and `evil-define-key` are not consistent with one another, because the first is a function and the second is a macro. Using an unquoted keymap in a function assumes that the keymap is not void (otherwise you get an error). `evil-define-key` is able to use an unquoted keymap exactly because it's a macro. That's how `evil-define-key` can accept an undefined variable for the keymap. 

I replicated the `evil-define-key` functionality but used a function instead of a macro, so the keymap needs to be quoted to avoid errors due to undefined variables. Using a quoted symbol also allows me to get the name of the symbol for use in reporting keybindings. 

I don't think macros should be used unless it's necessary, and macros can get especially tricky when they're called inside functions (the behavior can change depending on whether the file is byte compiled or not, for example). I noticed that the general package also uses functions with quoted keymaps by the way.

evil-collection-company.el
justbur: More specifically, `evil-define-key*` with `nil` as the state is equivalent to using a bunch of `define-key` calls like you had here. `evil-collection-define-key` adds the new filtering and reporting functionality on top.

justbur:
I rebased onto master and made some more improvements in `evil-collection-describe-all-bindings`

Ambrevar:
All good to me.  If we are all good here we can merge.

fbergroth:
This is really cool. Maybe it would be useful to having another command that only shows the bindings in the active keymaps, to complement `describe-mode`?

justbur:
There are different ways you could display the bindings. The point of this pr was to record them. I just wrote one quick describe function to show what could be done later.

jojojames:
Thanks, I'll give this a go!

Issue #145, opened 2018-06-11
Helm integration : Wrong type argument: symbolp (...)


dbeley:
Hello,

I have an issue with helm integration.

A minimal config file containing : 
    
```   
    (use-package evil
    :ensure t
    :init
    (setq evil-want-integration nil)
    :config
    (evil-mode 1)
    )

    (use-package evil-collection
    :after evil
    :ensure t
    :config
    (evil-collection-init)
    (setq evil-collection-setup-minibuffer t)
    )

    (use-package helm
    :after evil-collection
    :ensure t
    :config (helm-mode 1)
    (setq helm-mini-default-sources '(helm-source-buffers-list
                                    helm-source-recentf
                                    helm-source-bookmarks
                                    helm-source-buffer-not-found))
    (global-set-key (kbd "M-x") #'helm-M-x)
    (global-set-key (kbd "C-x r b") #'helm-filtered-bookmarks)
    (global-set-key (kbd "C-x C-f") #'helm-find-files)
    (global-set-key (kbd "C-x f") #'helm-recentf)
    (global-set-key (kbd "C-x C-b") #'helm-mini)
    (global-set-key (kbd "C-x b") #'helm-mini)
    )
```

Give me that kind of error on startup :

```
[...]

PR #146, opened 2018-06-11
Fix bugs from transition to new define-key

justbur:
Mostly this fixes unquoted maps in the helm bindings.


justbur:
Might address #145

jojojames:
Looks good to me. Can you double check @Ambrevar ?

PR #148, opened 2018-06-13
Enhance describe bindings functionality.

justbur:
Rename evil-collection-describe-all-bindings to
evil-collection-describe-bindings. By default the function prints all known
evil-collection bindings. With a non-nil interactive argument, print only the
bindings corresponding to active modes in the current buffer.


Ambrevar:
Looks good to me!

PR #150, opened 2018-06-13
Accept no bindings in evil-collection-define-key

justbur:
They all might be filtered by the white and black lists.


jojojames:
Thanks!

PR #151, opened 2018-06-14
Fix up helm minibuffer defaults

ahungry:
This makes the minibuffer usable with a few frequently used helm bindings - the unbound evil equivalents either error out or have no effect, so it is best to restore to the default Emacs style binds.

### DIFF
--- a/evil-collection-helm.el
+++ b/evil-collection-helm.el
@@ -172,6 +172,8 @@
   ;; C-o switches to evil state (again, not useful).
   (evil-collection-define-key '(insert normal) 'helm-map
     (kbd "RET") 'helm-maybe-exit-minibuffer
+    (kbd "M-v") 'helm-previous-page
+    (kbd "C-v") 'helm-next-page
     (kbd "C-p") 'helm-previous-line
     (kbd "C-n") 'helm-next-line
     (kbd "C-o") 'helm-next-source)
@@ -180,6 +182,8 @@
     (kbd "<tab>") 'helm-select-action   ; TODO: Ivy has "ga".
     (kbd "[") 'helm-previous-source
     (kbd "]") 'helm-next-source
+    (kbd "C-u") 'helm-previous-page
+    (kbd "C-d") 'helm-next-page
     "gk" 'helm-previous-source
     "gj" 'helm-next-source
     (kbd "(") 'helm-prev-visible-mark
--- a/evil-collection-helm.el
+++ b/evil-collection-helm.el
@@ -178,12 +178,18 @@
     (kbd "C-n") 'helm-next-line
     (kbd "C-o") 'helm-next-source)
 
+  (when evil-want-C-u-scroll
+    (evil-collection-define-key 'normal 'helm-map
+      (kbd "C-u") 'helm-previous-page))
+
+  (when evil-want-C-d-scroll
+    (evil-collection-define-key 'normal 'helm-map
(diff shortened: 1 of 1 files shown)


evil-collection-helm.el
```diff
@@ -165,10 +165,25 @@
...
+    (kbd "C-v") 'helm-next-page
+    (kbd "C-p") 'helm-previous-line
+    (kbd "C-n") 'helm-next-line
+    (kbd "C-o") 'helm-next-source)
+
   (evil-collection-define-key 'normal 'helm-map
     (kbd "<tab>") 'helm-select-action   ; TODO: Ivy has "ga".
     (kbd "[") 'helm-previous-source
     (kbd "]") 'helm-next-source
+    (kbd "C-u") 'helm-previous-page
```
Ambrevar: This should only be bound if `evil-want-c-u-scrolls` is true.

evil-collection-helm.el
```diff
@@ -165,10 +165,25 @@
     "%" 'helm-ff-run-query-replace-regexp
     "D" 'helm-ff-run-delete-file)       ; Ivy has "D".
 
+  ;; These helm bindings should always exist, the evil equivalents do
+  ;; nothing useful in the minibuffer (error or pure failure).
+  ;; RET can't do a second line in the minibuffer.
+  ;; The C-n/C-p completions error with 'helm in helm' session.
+  ;; C-o switches to evil state (again, not useful).
+  (evil-collection-define-key '(insert normal) 'helm-map
+    (kbd "RET") 'helm-maybe-exit-minibuffer
```
Ambrevar: I don't understand why we would need to bind this.  The bindings are already good for me and some Helms like Helm Find Files use a different binding (e.g. `helm-ff-RET`).
Can give an example?

evil-collection-helm.el
```diff
@@ -165,10 +165,25 @@
...
+    (kbd "M-v") 'helm-previous-page
```
Ambrevar: This binding was already fine by me.  I wonder why our setups are different.

evil-collection-helm.el
ahungry: With this package as is, if I press RET in the minibuffer (with minibuffer setup enabled) during a helm session, it only works if I'm in INSERT mode, not in NORMAL mode, which is counter intuitive, as j / k scroll through results, it is a pain to then press i RET instead of just RET.

evil-collection-helm.el
ahungry: I may have forgot to test M-v, was C-v fine as well? (that's the one that I ran into - it did nothing), so I added M-v for explicit completeness

evil-collection-helm.el
ahungry: If you go in minibuffer, type some text, press ESC to leave insert mode and go to NORMAL mode, pressing RET will select your entry?  I'll try testing in a vanilla/minimal but it may be a bit until I get to.

evil-collection-helm.el
ahungry: Sure

ahungry: updated

evil-collection-helm.el
Ambrevar: Sorry for the delay.  You were right, I can reproduce now.

Ambrevar:
Thanks for the patch!

ahungry:
You're welcome!

PR #153, opened 2018-06-16
lsp-ui-imenu: make 'h' goto previous section and 'l' to next

Dickby:
Feels more logical.


Ambrevar:
I suppose it was a typo.  I don't use lsp however.

Note that in general `C-j`/`C-k` and `[`/`]` are used to browse sections.  See Magit, Emms, Helm, Man...

Dickby:
I make a new pull-request to make it match those modes.

PR #154, opened 2018-06-17
fix bindings for lsp-ui next-kind and prev-kind

Dickby:
change the bindings to C-j and C-k to match other mode's bindings


Ambrevar:
@jojojames @noctuid Anyone using lsp-ui?

jojojames:
Looks ok to me. @Ambrevar I use it a little.

jojojames:
This looks good @Dickby . I need to find some time to test this (won't be a while..). Just wondering if h and l were used for anything else for that mode? Does navigation by hjkl make sense? If it doesn't make sense, maybe we can keep h and l along with c-j and c-k.

Anyways, I'll merge this soon. Thanks!

PR #155, opened 2018-06-18
Improve evil-collection-describe-bindings

justbur:
Use better version of the map active check and remove the old one. Don't print
out menu-items and lambdas.


Ambrevar:
Looks good to me.

PR #160, opened 2018-06-28
neotree: Replace neotree-make-executor with custom macro

### LINKED ISSUE #158: Incompatibility with neotree: `evil-collection-neotree-setup: Invalid function: neotree-make-executor`
* [init.el](https://gitlab.com/mrbig/dotfiles/blob/master/emacs/emacsv2/init.el)
* [.emacs.d](https://gitlab.com/mrbig/dotfiles/tree/master/emacs/emacsv2)
* [my complete environment (refer to the Manjaro install)](https://gist.github.com/mrbig033/63fd7f4f51b6a8ae379abae74c52aff2)

After I customized neotree and rebooted Emacs, it started showing the following message:
`evil-collection-neotree-setup: Invalid function: neotree-make-executor` after `M-x neotree` and `M-x neotree-toggle. It doesn't work anymore. Here's the [use-package](https://gitlab.com/mrbig/dotfiles/blob/master/emacs/emacsv2/customizations/custom-settings/use-packages.el#L265). I also tried without the extra settings (hook, theme etc). Same results. But when I disabled Evil-collection, neotree worked. For this reason I'm posting the issue here.

justbur:
neotree-make-executor is a macro, so it will get expanded correctly at compile
time if neotree is available. If neotree is not available, somehow
neotree-make-executor gets tagged as an invalid function at compile time, even
if neotree is available at run time.

It's not a good idea to call a macro from a foreign library when you don't know
if that library will be available at compile time. There are other alternatives
to this fix, but since neotree-make-executor is so simple this seems like the
best option.

Fixes #158

### DIFF
--- a/evil-collection-neotree.el
+++ b/evil-collection-neotree.el
@@ -31,8 +31,17 @@
 (require 'evil-collection)
 (require 'neotree nil t)
 
-(declare-function neotree-make-executor "neotree.el")
 (defconst evil-collection-neotree-maps '(neotree-mode-map))
+
+(defmacro evil-collection-neotree-exec (&optional file-fn dir-fn)
+  "Replacement for `neotree-make-executor'."
+  ;; `neotree-make-executor' is a macro, so it will get expanded correctly at
+  ;; compile time if neotree is available. If neotree is not available, somehow
+  ;; `neotree-make-executor' gets tagged as an invalid function at compile time.
+  `(lambda (&optional arg)
+     (interactive "P")
+     (neo-global--select-window)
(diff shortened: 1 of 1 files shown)


Ambrevar:
OK, this is the same problem as we had with the Helm macros.

justbur:
Another option would be to prevent the files from being byte compiled when they use macros.

Ambrevar:
True.  I suppose byte-compilation does not bring much benefit here.
Plus this could save us from some code duplication.  It would be a better option.

justbur:
OK, see #161

PR #163, opened 2018-06-28
Convert evil-collection-inhibit-insert-state to function

justbur:
Move to evil-collection.el. Make requires consistent across all package
files. Delete evil-collection-util.el.

This might be a little controversial, but I think these changes make things more consistent with respect to uses of require and the binding functions. Deleting evil-collection-util.el was a judgement call. I prefer it next to the main binding function, and the extra file with one function made it easy to forget to require that file.


Ambrevar:
Agreed.  I'm Ok to merge this.
@jojojames?

jojojames:
Fine by me, might be worth looking at the search util file too.

Ambrevar:
@justbur Want to move `evil-collection-search-util.el` as well?

justbur:
@Ambrevar I'm not sure what you mean

Ambrevar:
What @jojojames suggest, add another commit to move the function from `evil-collection-search-util.el` to `evil-collection.el`. 

But no worries, I can also do it later.

justbur:
I see. Sure, I just did it

PR #164, opened 2018-06-29
add dd emulation to eshell-mode

akacase:
I wanted to emulate "dd" behavior in other vi shell modes, this behavior is analogous to its bash/zsh counterparts.

### DIFF
--- a/evil-collection-eshell.el
+++ b/evil-collection-eshell.el
@@ -51,6 +51,14 @@
   (eshell-interrupt-process)
   (evil-insert 1))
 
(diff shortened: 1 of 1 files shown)


Ambrevar:
I think this would need more work to function properly.
See https://github.com/emacs-evil/evil/issues/603 and #123.
@jojojames, @noctuid ?

akacase:
bummer, you're right. I'll have to figure it out more, it interferes with the other `d` prefix bindings, i.e. `dw`.

PR #173, opened 2018-07-11
mu4e: Convert evil-define-key to evil-collection-define-key

justbur:


Ambrevar:
Thanks for the catch!

PR #178, opened 2018-07-13
move package files into subdirectories

justbur:
This is the first step towards having READMEs available for each package


jojojames:

I think I might have to play around with the melpa recipe first https://github.com/melpa/melpa/blob/a9b93a8e3750e4e7767498e418f46d553d814604/recipes/evil-collection to pull the files correctly. I might be wrong but I think it ignores subdirectories by default.

justbur:
@jojojames Yes, you do need to fix the melpa recipe. I think adding `*/*.el` to the defaults should be fine.

jojojames:
Ah looks like melpa automatically pulls files from nested directories to the root of the project.

```
;; For testing.
(evil-collection :fetcher github :repo "justbur/evil-collection"
                 :branch "dirs"
                 :files (:defaults "*/*.el"))
```

Hmnn.. Not sure if we want a split in how evil-collection is consumed when developing vs melpa.

I didn't see a way to keep the folder structure (didn't look too long..).

If anyone else knows more about it, let me know.

I don't mind too much if there's a split as long as things still work reasonably but I'd be interested in others thoughts too @fbergroth @noctuid @Ambrevar @justbur .

woman/evil-collection-woman.el
```diff
@@ -28,7 +28,8 @@
 
 ;;; Code:
 (require 'evil-collection)
-(require 'evil-collection-man) ; WoMan's keymap inherits from Man.
+(unless (featurep 'evil-collection-man)
+  (evil-collection-init '(man)))
```
Ambrevar: Good catch, but can you leave the comment? :)

Ambrevar:
Looks good to me too!

I'm also against a split in the way Evil Collection is deployed: I think it would heavily defeat the Rule of Least Surprise.

I don't much about MELPA build files but I'm sure there must be a rule to preserve the folder structure.

fbergroth:
Do you really need create subdirectories? You could just also just put the readmes in, say, `doc/evil-collection-magit.org`, and bundle that folder?

Ambrevar:
I'm not opposed to that idea, although I think the main point was that the READMEs would display nicely and GitHub.  All READMEs in a `doc` folder would prevent that.

jojojames:
I think being able to see the readmes in github is a nice goal and the split in deployment if it comes to it is worth the cost. Though there's probably a way to do it...

justbur:
Ok, I think I have a reasonable solution to the concerns about melpa. I put all of the new directory under the `modes` directory. If you make the melpa recipe `:files (:defaults "modes")` then melpa copies the directory structure. I tested building the package and loading it locally, but someone else should test it for themselves.

jojojames:
That works for me. @Ambrevar You good with that approach?

I can make the melpa recipe PR before we merge this.

Ambrevar:
Looks good to me.
I cannot test right now, but if someone else has tested it then feel free to merge.

Thanks @justbur !

jojojames:
We'll need to get the change in melpa first before merging I think.

I'll test it again soon and then make a PR to melpa.

jojojames:
Opened a PR on melpa here: https://github.com/melpa/melpa/pull/5617

jojojames:
Melpa change is in @justbur

justbur:
Great

justbur:
opened a new PR at #252 with the changes by @agsdot

PR #179, opened 2018-07-14
add missing bindings and map vdiff-3way as well

bling:


evil-collection-vdiff.el
```diff
@@ -32,26 +32,34 @@
...
-                                evil-collection-delete-operators)
-                      #'vdiff-send-changes)))))
+
+  (dolist (mode '(vdiff-mode vdiff-3way-mode))
+    (evil-define-minor-mode-key 'normal mode
+      "zc" 'vdiff-close-fold
+      "zM" 'vdiff-close-all-folds
+      "zo" 'vdiff-open-fold
+      "zR" 'vdiff-open-all-folds
+      "go" 'vdiff-receive-changes
```
jojojames: I wonder if go and gp are good keys for these two commands. Maybe others can comment.

Otherwise, they are acceptable to me.

jojojames:

evil-collection-vdiff.el
bling: it was copy/pasted from the readme: https://github.com/justbur/emacs-vdiff#saving-updating-and-exiting

the key improvement is making sure that both `vdiff` and `vdiff-3way` are mapped, as the latter is used by default for magit diffs.

bling: speakin, of which, my copy/paste is wrong.  the readme has it under motion state, rebasing...

Ambrevar:
I don't use vdiff so looks good to me as far as I can tell :p

jojojames:
Ok, lets go with those bindings for now and improve upon them later if anyone has any better ideas.

evil-collection-vdiff.el
```diff
@@ -32,26 +32,35 @@
...
+
+  (dolist (mode '(vdiff-mode vdiff-3way-mode))
+    (evil-define-minor-mode-key 'normal mode
+      "zc" 'vdiff-close-fold
+      "zM" 'vdiff-close-all-folds
+      "zo" 'vdiff-open-fold
+      "zR" 'vdiff-open-all-folds
+      "]c" 'vdiff-next-hunk
+      "[c" 'vdiff-previous-hunk)
+    (evil-define-minor-mode-key 'motion mode
```
jojojames: Just noticed this. Is there a reason this is in motion? I don't use vdiff so I'm not familiar with it.

jojojames: Is this operator or normal?

bling: i moved `go` and `gp` back to normal state.

jojojames:
Cool, thanks @bling !

edkolev:
This change shouldn't be necessary - evil already has vdiff folding support https://github.com/emacs-evil/evil/blob/230b87212c81aaa68ef5547a6b998d9c365fe139/evil-vars.el#L1665

Refs https://github.com/justbur/emacs-vdiff/issues/16

However, I know one (corner-case?) scenario where the evil vdiff folding doesn't work smoothly - when the buffer being vdiff-ed has outline-mode enabled -- then, evil tries to fold/unfold using outline-mode because it's preceding vdiff in `evil-fold-list`. I'm planning to move vdiff higher in the `evil-fold-list`.

bling:
hmmm, interesting.  on my config, without these changes using zR/zM will only fold the active buffer, and leave it unsynchronized.  i'll have to try again from a clean init.el.

however, `vdiff-3way-mode` being absent in `evil-fold-list` would still be an issue, right?

bling:
mystery solved.  i had `origami` enabled, which is higher in `evil-fold-list`, so it clobbered the values of `vdiff`.

also confirmed that `vdiff-3way-mode` needs something set (either via evil-collection or evil-fold-list).

edkolev:
I just submitted a PR in evil to address this - https://github.com/emacs-evil/evil/pull/1072

I can go ahead and merge it, but I would rather wait until another set of eyes take a look at the code. I've tested `vdiff-3way-mode`, works fine.

PR #181, opened 2018-07-19
reverse space and shift-space keys

japhir:


Ambrevar:
I think the current behaviour is correct: SPC scrolls up (i.e. moves down) while S-SPC scrolls down (i.e. moves up).
Can you explain why you would reverse them?

japhir:
aah I misread it and it didn't work according to what I expected, because I, stupidly, had space set as my evil-leader. Please ignore! :)

PR #182, opened 2018-07-19
add zi, zo, zz zoom keyboard shortcuts

japhir:
similar to vim vixen

### DIFF
--- a/evil-collection-pdf.el
+++ b/evil-collection-pdf.el
@@ -120,9 +120,12 @@
 
     ;; zoom
     "+" 'pdf-view-enlarge
-    "-" 'pdf-view-shrink
-    "0" 'pdf-view-scale-reset
+    "zi" 'pdf-view-enlarge
     "=" 'pdf-view-enlarge
+    "-" 'pdf-view-shrink
+    "zo" 'pdf-view-shrink		      
+    "0" 'pdf-view-scale-reset
+    "zz" 'pdf-view-scale-reset
 
     ;; TODO: Why are those image-* bindings in pdf-tools?
     "a+" 'image-increase-speed


Ambrevar:
OK for the `=` fix.

Are you talking about this Vim Vixen? https://github.com/ueokande/vim-vixen

I don't know if `zi`/`zo` are "standard" enough.  I'd tend to think that `+`/`-` are more convenient.

japhir:
the first thing was indeed just moving the things around so they're grouped by aliases.
Yes, sorry, should have provided a link.

[Zathura](https://pwmt.org/projects/zathura/) (the pdf reader that I just set up perfectly and then found out about pdf-tools) uses them too according to their [man-page](https://man.cx/zathura(1))

Ambrevar:
Why not.  Should we add the `z`-bindings to the Zoom section in the rationale then?
@jojojames ?

japhir:
Do you think z0 (zero) is better or zz for zooming back to default settings? I think I'd rarely use it, so in hindsight I'd say z0 is probably better.

japhir:
regarding convenience: zi and zo are both very close to the home row, + and - are right-pinky or numpad keys that are harder to reach ;-).

Ambrevar:
`zz` could be dangerously confused with `ZZ`! :p

jojojames:
What are the z keys for folding? I'm not sure if those keys are standard enough either but don't have a strong opinion against it. Up to you @Ambrevar

Ambrevar:
Oopsy!  @japhir: I forgot to mention that commits should be prefixed with the mode they target, in this case `pdf: `.  Too late...

japhir:
Ah didn't know, but it makes sense. I'll make sure to do so if I do another pull request in the future!

PR #183, opened 2018-07-21
Debbugs: Tweak

jojojames:
@Ambrevar Can you double check these changes look good before merging?

### DIFF
--- a/evil-collection-debbugs.el
+++ b/evil-collection-debbugs.el
@@ -47,9 +47,10 @@
     "d" 'debbugs-gnu-display-status
 
     ;; filter
-    (kbd "s") 'debbugs-gnu-narrow-to-status
+    "s" 'debbugs-gnu-narrow-to-status
+    "S" 'debbugs-gnu-search
     ;; "S" 'debbugs-gnu-widen ; Useless if we can just press "s RET" (empty filter).
-    "S" 'debbugs-gnu-toggle-suppress
+    "x" 'debbugs-gnu-toggle-suppress
     "r" 'debbugs-gnu-show-all-blocking-reports
 
     ;; sort
@@ -65,6 +66,10 @@
 
     ;; refresh
     "gr" 'debbugs-gnu-rescan
+
+    "g?" 'debbugs-gnu-manual
+    "gt" 'debbugs-gnu-view-bug-triage
+    "g#" 'debbugs-gnu-bugs
 
     ;; quit
     "q" 'quit-window


evil-collection-debbugs.el
```diff
@@ -49,7 +49,7 @@
     ;; filter
     (kbd "s") 'debbugs-gnu-narrow-to-status
     ;; "S" 'debbugs-gnu-widen ; Useless if we can just press "s RET" (empty filter).
-    "S" 'debbugs-gnu-toggle-suppress
+    "x" 'debbugs-gnu-toggle-suppress
```
Ambrevar: Why not.
In general many toggle keys are z-prefixed.  What do you think about `zs` or `zh`?

evil-collection-debbugs.el
```diff
@@ -49,7 +49,7 @@
     ;; filter
     (kbd "s") 'debbugs-gnu-narrow-to-status
```
Ambrevar: We can leave out the `kbd` here! :)

evil-collection-debbugs.el
```diff
@@ -66,6 +66,11 @@
     ;; refresh
     "gr" 'debbugs-gnu-rescan
 
+    "g?" 'debbugs-gnu-manual
+    "gt" 'debbugs-gnu-view-bug-triage
+    "g#" 'debbugs-gnu-bugs
+    "S" 'debbugs-gnu-search
```
Ambrevar: Much better indeed.  I think I'd move this line next to "s" though, to highlight the link between the two.

evil-collection-debbugs.el
jojojames: I think x is open and it was the original keybinding. I don't have an opinion otherwise though so we can put it at zs if you think that's better.

evil-collection-debbugs.el
jojojames: I think it was there before, but I can remove it for sure.

evil-collection-debbugs.el
jojojames: Will do.

evil-collection-debbugs.el
Ambrevar: Yes, that was my mistake :p

evil-collection-debbugs.el
Ambrevar: Fine with `x`, bindings are quite simple so far.  We could always move it to `zs` if bindings get more crowded.

PR #185, opened 2018-07-23
vdiff: don't bind fold keys, which are handled by evil-fold-list

edkolev:
Refs https://github.com/emacs-evil/evil-collection/pull/179, https://github.com/emacs-evil/evil/pull/1072

@bling I've removed the `go` and `gp` binding because `do` and `dp` are already bound, and are in-line with vim


edkolev:
You're more than welcome!

PR #186, opened 2018-07-28
Use annalist for recording/describing keybindings

noctuid:
This is not at all ready to be merged (and the package is not on MELPA), but I thought I'd post what I have in case anyone wants to try it out or look at it (I've only tested it with evil-collection's proced keybindings). Issues/problems with the package are listed in the todo.org. I'll post here again once I've finished the readme. I'll also post an example of how tagging/filtering would work (it's possible to filter by the recorded data or by recorded metadata that isn't printed).

[Here](https://github.com/noctuid/annalist.el) is a link to the repo.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -8,7 +8,7 @@
 ;; Pierre Neidhardt <mail@ambrevar.xyz>
 ;; URL: https://github.com/emacs-evil/evil-collection
 ;; Version: 0.0.2
-;; Package-Requires: ((emacs "25.1") (cl-lib "0.5") (evil "1.2.13"))
+;; Package-Requires: ((emacs "25.1") (cl-lib "0.5") (evil "1.2.13") (annalist "1.0"))
 ;; Keywords: evil, tools
 
 ;; This program is free software; you can redistribute it and/or modify
@@ -33,9 +33,9 @@
 ;; Some minibuffer-related packages such as Helm rely on this option.
 
 ;;; Code:
-(eval-when-compile (require 'subr-x))
 (require 'cl-lib)
 (require 'evil)
+(require 'annalist)
 
 (defvar evil-want-integration)
 (defvar evil-want-keybinding)
@@ -257,12 +257,6 @@ This is a list of strings that are suitable for input to `kbd'."
   :type '(repeat string)
   :group 'evil-collection)
 
-(defvar evil-collection--bindings-record (make-hash-table :test 'eq)
-  "Record of bindings currently made by Evil Collection. This is
-a hash-table with the package symbol as a key.  The associated
-values are the package's bindings which are stored as a list of
-the form ((STATE KEY BINDING)).")
-
 (defvar evil-collection-setup-hook nil
   "Hook run by `evil-collection-init' for each mode that is evilified.
 This hook runs after all setup (including keybindings) for a mode has already
@@ -272,36 +266,27 @@ Evil Collection for that mode. More arguments may be added in the future, so
 functions added to this hook should include a \"&rest _rest\" for forward
 compatibility.")
 
-(defvar evil-collection-describe-buffer "*evil-collection*"
-  "Name for Evil Collection buffer used to describe bindings.")
-
 (defun evil-collection-define-key (state map-sym &rest bindings)
   "Wrapper for `evil-define-key*' with additional features.
-Unlike `evil-define-key*' MAP-SYM should be a quoted keymap other
-than the unquoted keymap required for `evil-define-key*'.  This
-function adds the ability to filter keys on the basis of
-`evil-collection-key-whitelist' and
-`evil-collection-key-blacklist'. It also stores bindings in
-`evil-collection--bindings-record'."
+Unlike `evil-define-key*' MAP-SYM should be a quoted keymap other than the
+unquoted keymap required for `evil-define-key*'. This function adds the ability
+to filter keys on the basis of `evil-collection-key-whitelist' and
+`evil-collection-key-blacklist'. It also records bindings with annalist.el."
   (declare (indent defun))
   (let* ((whitelist (mapcar 'kbd evil-collection-key-whitelist))
          (blacklist (mapcar 'kbd evil-collection-key-blacklist))
-         (record (gethash map-sym evil-collection--bindings-record))
          filtered-bindings)
     (while bindings
       (let ((key (pop bindings))
             (def (pop bindings)))
         (when (or (and whitelist (member key whitelist))
                   (not (member key blacklist)))
-          (if (consp state)
-              (dolist (st state)
-                (push (list (if st st 'all) (key-description key) def)
-                      record))
-            (push (list (if state state 'all) (key-description key) def)
-                  record))
+          (annalist-record 'evil-collection 'keybindings
+                           (list map-sym state key def)
+                           :local (or (eq map-sym 'local)
+                                      (local-variable-p map-sym)))
           (push key filtered-bindings)
           (push def filtered-bindings))))
-    (puthash map-sym record evil-collection--bindings-record)
     (setq filtered-bindings (nreverse filtered-bindings))
     (cond ((null filtered-bindings))
           ((and (boundp map-sym) (keymapp (symbol-value map-sym)))
@@ -357,49 +342,26 @@ This is particularly useful for read-only modes."
         (string-lessp a-state b-state)
       (string-lessp a-key b-key))))
 
+(annalist-define-view 'keybindings 'evil-collection-valid
+  (list (list 'keymap :sort #'annalist-string-<)
+        (list 'state :sort #'annalist-string-<))
+  :inherit 'valid)
+
+(annalist-define-view 'keybindings 'evil-collection-active
+  (list (list 'keymap :sort #'annalist-string-<)
+        (list 'state :sort #'annalist-string-<))
+  :inherit 'active)
+
 (defun evil-collection-describe-bindings (&optional arg)
   "Print bindings made by Evil Collection to separate buffer.
 
 With non-nil ARG, restrict to bindings corresponding to active
 modes in the current buffer."
   (interactive "P")
-  (let ((orig-buf (current-buffer))
-        (desc-buf (get-buffer-create evil-collection-describe-buffer)))
-    (switch-to-buffer-other-window desc-buf)
-    (with-current-buffer desc-buf
-      (erase-buffer)
-      (org-mode)
-      (dolist (keymap
-               (sort (hash-table-keys evil-collection--bindings-record)
-                     (lambda (a b)
-                       (string-lessp (symbol-name a)
-                                     (symbol-name b)))))
-        (when (or (null arg)
-                  (with-current-buffer orig-buf
-                    (and (boundp keymap)
-                         (memq (symbol-value keymap) (current-active-maps)))))
-          (insert "\n\n* " (symbol-name keymap) "\n")
-          (insert "
-| State | Key | Definition |
-|-------|-----|------------|
-")
-          (cl-loop
-           for (state key def) in
-           (sort (copy-sequence
-                  (gethash keymap evil-collection--bindings-record))
-                 #'evil-collection--binding-lessp)
-           do
-           (when (and def (not (eq def 'ignore)))
-             (insert (format "| %s | %s | %s |\n"
-                             state
-                             (replace-regexp-in-string "|" "¦" key)
-                             (cond ((symbolp def) def)
-                                   ((functionp def) "(lambda ...)")
-                                   ((consp def)
-                                    (format "(%s ...)" (car def)))
-                                   (t "??"))))))
-          (org-table-align)))
-      (goto-char (point-min)))))
+  (annalist-describe 'evil-collection 'keybindings
+                     (if arg
+                         'evil-collection-active
+                       'evil-collection-valid)))
 
 (defun evil-collection--translate-key (state keymap-symbol
                                              translations
--- a/Cask
+++ b/Cask
@@ -6,4 +6,5 @@
 (development
  (depends-on "f")
  (depends-on "ert-runner")
- (depends-on "package-lint"))
+ (depends-on "package-lint")
+ (depends-on "annalist"))


Ambrevar:
Looks nice!
The `todo.org` is missing though :p 

Also could you provide other examples where this package would be useful?

noctuid:
> The todo.org is missing though

Yeah, my bad. I forgot to push.

> Also could you provide other examples where this package would be useful?

With some modification, it could be used to automatically generate tables listing the keybindings made in every package for this repo. 

It would be more work and may not be the best idea, but tagging/filtering could also be used for classifying different keybindings. For example, it would be possible to tag a definition as creating a sorting keybinding. The user could then only display sorting keybindings made in evil-collection. The user could also potentially control the behavior of specific types of keybindings (e.g. they could specify that sort keybindings should use their own keybinding instead of `o`); this would be a more fine-grained alternative to using key translation. If evil-collection ever did start using general (or some spinoff tailored towards use in other packages), general could potentially take care of recording, blacklisting, whitelisting, key translation, etc. The main changes would be tagging all keybindings (which would take a while). Again, I'm not suggesting that this would be ideal, but it is a possibility and may be worth considering.

It will be used in general.el in the future. General provides similar wrapper functions for defining settings, hooks, and advice. These all have additional functionality or more convenient syntax than the defaults (e.g. the idea of the settings wrapper is to automatically trigger defcustom `:set` expressions like customize does but with the same syntax as `setq`; it needs some work to match `setq` perfectly). With annalist, they will also record all user settings, hooks, and advice to be viewable later.

As for potential examples specifically for non-keybindings and evil-collection, it could potentially be used to record/display any settings, hooks, or advice that evil-collection made for other packages. It's potentially useful for anything that would make sense to display in a table.

I tried to generalize the idea of recording/displaying information. It was more work than I expected, and I changed how types (e.g. `keybindings` is a type; `settings` will be another) were defined several times because I wasn't happy with the syntax. Any feedback on the current syntax would be appreciated. I think once I add the readme, the capabilities will be made more clear. The basics are support for custom formatting, sorting, and filtering.

evil-collection.el
```diff
@@ -36,6 +36,7 @@
 (eval-when-compile (require 'subr-x))
 (require 'cl-lib)
 (require 'evil)
+(require 'annalist)
```
jojojames: Is there a performance/startup impact on this require? I know as packages grow, these requires become linearly expensive with that growth.

(An example that springs to mind is lispy which requires a lot of packages!)

jojojames:
Did a quick skim but I think I'll need to read the code (annalist) more closely.

Looks good on the initial skim.

evil-collection.el
noctuid: Annalist only requires cl-lib and subr-x (and I will keep dependencies to a minimum). Requiring org is necessary for displaying keybindings but not when loading annalist or when recording keybindings, so I don't think there should be an issue. Requiring annalist could also be deferred until evil-collection actually makes keybindings, but it's probably unnecessary.

jojojames:
Haven't had a chance yet to clone this and play with it but planning on soon!

noctuid:
I finally had some free time today and finished most of the readme. I mostly know what needs to be fixed with the internals, but feedback on the available functionality and type definition syntax would be appreciated (if you think anything should be changed). I believe that I have accounted for all of @justbur's functionality here and general's old functionality (though some things need to be fixed/addressed). I will probably still be pretty busy, so it may take me a few weeks or longer to address these issues.

I left off/didn't finish a section about the internal implementation. Right now how the data is stored is tied to how it is defined (true for both evil-collection and general as well). Heading items are keys in hash tables (doesn't really offer a performance benefit over alists currently, but it's arguably more readable and could offer a performance benefit if you only wanted to print keybindings in a specific keymap, for example). Changing which items are printed as headings would require recording data multiple times currently. For example, if I just stored a giant list of records instead, it would be possible to print just one table (no headings) in correct recorded order, but there would be performance implications when recording (since annalist checks for duplicate records) and when sorting by multiple things. More importantly, sorting would also be a lot harder to implement. If it was hash tables/alists/plists all the way down, it wouldn't be possible to print records in their exact recorded order. I'd kind of like to use a database, but that would almost certainly be overkill and may not be easily possible either.

Since for most types of recordable things one way of printing makes the most sense, this probably won't be much of an issue, but I thought I'd mention it anyway.

jojojames:
@noctuid Cool, thanks. Looking to check it out some time this week or weekend.

jojojames:
Had a little bit of time to clone the branch and check it out:

Some quick comments/questions.

In the readme, it might be cool to provide some sample 'results' so the user can see what to expect given some input.

For example:

```
(defun my-describe-active-keybindings ()
  (interactive)
  (let ((annalist-override-settings
         (list :items
           ;; :title is only specified for readability (not actually changing)
           (list 0 (list :title "Keymap"
                     ;; only show keys bound in active keymaps
                     :predicate #'annalist--active-keymap-p
                     ;; sort keymaps alphabetically
                     #'annalist--string-<)))))
    (annalist-describe 'my 'keybindings)))
```

It might be nice to show the result of running that function, given some sample keymap, of course.

When I ran: evil-collection-describe-bindings

1. I was surprised so many bindings were already created. I'm guessing all the 'with-eval-after-loads ran already. Not a big deal and it might be because I already turned all those modes on.

2. It might worth thinking about whether the default behavior is to show keybinds for the current mode/current context vs all keybinds that evil-collection makes. e.g. What does a prefix argument cause evil-collection-describe-bindings to do. I don't have a strong opinion here.

3. When I ran M-u evil-collection-describe-bindings,

I got this:

```


* evil-motion-state-map

| State | Key                                  | Definition                      |
|-------+--------------------------------------+---------------------------------|
| all   | <remap> <avy-goto-char-2-above>      | evil-avy-goto-char-2-above      |
| all   | <remap> <avy-goto-char-2-below>      | evil-avy-goto-char-2-below      |
| all   | <remap> <avy-goto-char-2>            | evil-avy-goto-char-2            |
| all   | <remap> <avy-goto-char-in-line>      | evil-avy-goto-char-in-line      |
| all   | <remap> <avy-goto-char-timer>        | evil-avy-goto-char-timer        |
| all   | <remap> <avy-goto-char>              | evil-avy-goto-char              |
| all   | <remap> <avy-goto-line-above>        | evil-avy-goto-line-above        |
| all   | <remap> <avy-goto-line-below>        | evil-avy-goto-line-below        |
| all   | <remap> <avy-goto-line>              | evil-avy-goto-line              |
| all   | <remap> <avy-goto-subword-0>         | evil-avy-goto-subword-0         |
| all   | <remap> <avy-goto-subword-1>         | evil-avy-goto-subword-1         |
| all   | <remap> <avy-goto-symbol-1-above>    | evil-avy-goto-symbol-1-above    |
| all   | <remap> <avy-goto-symbol-1-below>    | evil-avy-goto-symbol-1-below    |
| all   | <remap> <avy-goto-symbol-1>          | evil-avy-goto-symbol-1          |
| all   | <remap> <avy-goto-word-0>            | evil-avy-goto-word-0            |
| all   | <remap> <avy-goto-word-1-above>      | evil-avy-goto-word-1-above      |
| all   | <remap> <avy-goto-word-1-below>      | evil-avy-goto-word-1-below      |
| all   | <remap> <avy-goto-word-1>            | evil-avy-goto-word-1            |
| all   | <remap> <avy-goto-word-or-subword-1> | evil-avy-goto-word-or-subword-1 |


* ivy-mode-map

| State | Key      | Definition               |
|-------+----------+--------------------------|
| all   | <escape> | minibuffer-keyboard-quit |

```

It migth be because I'm using ivy/counsel-M-x but it would've been nicer for the describe-bindings to be for before I ran the command.

I still need to dig more into the code but this looks really good so far.

noctuid:
> In the readme, it might be cool to provide some sample 'results' so the user can see what to expect given some input.

Agreed. I'll do this when I get a chance.

> I don't have a strong opinion here.

Me neither. I think the current behavior is fine though.

Your listed output is from the old command; annalist does not use a state column by default and additionally has a previous column. Annalist uses the same active keymap check, so the behavior is probably still the same. Maybe there should be a blacklist for active keymaps (e.g. to not show minibuffer keymaps like ivy's)?

evil-collection.el
```diff
@@ -296,43 +282,20 @@ This is particularly useful for read-only modes."
...
-                                   (t "??"))))))
-          (org-table-align)))
-      (goto-char (point-min)))))
+  (let ((annalist-override-settings
+         (list :items
+           (list 0 (list :title "Keymap"
+                         :predicate (if arg
+                                        #'annalist--active-keymap-p
+                                      #'annalist--valid-keymap-p)
+                         :sort #'annalist--string-<)
```
justbur: It seems like this should be `annalist-string-<`

evil-collection.el
```diff
@@ -296,43 +282,20 @@ This is particularly useful for read-only modes."
...
+                 1 (list :title "State"
+                         :predicate
+                         (if arg
+                             #'annalist--valid-state-p
+                           #'annalist--valid-state-and-evil-on-p)
+                         :sort #'annalist--string-<)))))
```
justbur: Same comment here

noctuid:
@justbur One thing that is possible with your current code for evil-collection is to rename the describe buffer. Should I add this/is there a case where this would be useful?

Also didn't realize `hash-table-keys` was using `defsubst` (or that nothing in `subr-x` is necessary at load time; wonder why that is) until I saw the `eval-when-compile` here. Updated that in annalist.

justbur:
> @justbur One thing that is possible with your current code for evil-collection is to rename the describe buffer. Should I add this/is there a case where this would be useful?

Not terribly useful I don't think, no.

jojojames:
@noctuid Bump! There was some awesome work here to it would be sad to see it languish!

I'd be happy with 'good' and not 'perfect' if it's in good shape.

noctuid:
Sorry for never finishing this. Unfortunately the amount of of free time I have to work on Emacs stuff dropped to close to zero soon after I made this PR. I haven't prioritized this since both general and evil-collection at least have a working implementation of this idea. The package isn't ready, but it was close to being usable, so I will try to take a look at it this month. Feel free to ping me in a month if there aren't any updates (I may forget).

I ended up being unhappy with how far I could easily go using hash tables vs. a real database and sql. Most of the functionality that would be hard with the current data formatting isn't essential, but I want the package to be able to grow to support it. I won't worry about that for now besides ensuring that the definition syntax will make sense and work even if how data is stored changes. When I get some time, I'll try to fix the remaining issues and update this PR.

jojojames:
> dropped to close to zero soon after I made this PR. 
Ha! Can totally empathize!

> When I get some time, I'll try to fix the remaining issues and update this PR.

Sounds good with me, and please take your time!

noctuid:
I had some time this weekend to look it at. I've updated the definition syntax to be simpler and made various improvements. There is now a distinction between settings used for storing vs. settings used for displaying records. Alternate display settings are now defined as "views" instead of using `annalist-override-settings`. The same two variables are used for storing information instead of creating new variables for every new type. I'm now preprocessing keybindings, so that `'normal 'global` shows up the same as `nil 'evil-normal-state-map` (this may not matter for evil-collection; I'm not sure it makes any keybindings like this).

The syntax updates will make it pretty easy to switch to a different storage method in the future for more flexibility if desired (more information about current limitations in todo.org). I still need to make a few more changes before it's ready for MELPA (see todo.org; mainly I need to add tests and make sure I didn't break some keywords with the syntax rewrite). It is useable with evil-collection now though if anyone wants to try (with the `definition-syntax-update` branch).

I've updated the PR. It now additionally checks whether the keymap variables is buffer local to determine if the keybinding is buffer local.

jojojames:
Sounds good, I'll give it a shot over the weekend or next.

noctuid:
Auto formatting extracted lisp with `lispy-multiline`:
2019 08 01_21:36:42_-_emacsbahamut

And with `lispy-alt-multiline`:
2019 08 01_21:46:25_-_emacsbahamut

Turned out pretty nice I think. I think `lispy-alt-multiline` normally looks better, so I will probably go with that (`lispy-multiline` can result in extremely long lines it seems, e.g. 160 characters).

jojojames:
Got a chance to test this out. Seems pretty good to me.

Some notes:

I'm guessing annalist will have to be added to Cask when it's ready so evil-collection can run make on it.

Without annalist/this PR, I was getting an extra 30ms in startup time. From 70ms to 100ms (everything including annalist bytecompiled). I think 30ms is pretty low in a vacuum but I'd prefer to get startup time as minimal as possible. 

Even now it'd be nice if evil-collection didn't take 70ms to start up. (10 packages each taking 100 ms to start up is already 1 second :\).

Otherwise, seems pretty good to me.

noctuid:
Improving speed is on the todo list. I'll make some improvements this week (though I don't know if they will be major). Are you actually requiring all the packages evil-collection makes `with-eval-after-load`s for? With most keybindings deferred, annalist's speed doesn't matter as much. According to `benchmark-run`, annalist still adds 50ms to my init time for ~500 keybindings though, which is less than ideal. Worst case scenario, I can add a variable to disable annalist for users who don't want it and/or a variable to disable checking for duplicate records (which I'm guessing takes the most time), but I can hopefully make it faster without doing that.

jojojames:
Yeah, I think the use case of running annalist (e.g. via evil-collection-describe-keybindings) doesn't matter to me since it's going to be relatively slow (loading org, etc) either way and most users, I don't think, will care.

So I'm mainly commenting on emacs startup time.

e.g. with

```
(use-package evil-collection
  :ensure t
  :after evil
  :init
  :config
  (evil-collection-init))
```

I'm not sure where that time is added though (didn't really look too hard). Not sure if it's during the portion of time where evil-collection is executing on with-eval-after-loads (that then generates the annalist records?) that's loading at the same time as emacs startup (packages that I don't defer).

noctuid:
Yeah I was specifically talking about the recording part/startup time. If a user didn't ever want to use annalist, they might want to disable recording entirely (even if it was much faster than it is currently).

I've written most of the tests now. I'm going to finish that first before changing how the data is stored, so I'll know right away if I break something. Right now, annalist iterates through the list of recorded keybindings for a particular keymap/state for every new record to check for duplicates. I'm guessing this and maybe the processing functions are the cause of the added time. I'll profile first to see what's slow, but I'll probably try to rewrite it to use another hash table at the end instead of a list, so it's not necessary to iterate through previous records anymore.

Everything else is pretty much done for now. My goal is to be done and make a MELPA PR before next week.

General.el is not fast either. Speed is something I'll also have to consider if I decide to make a version intended for packages like evil-collection.

jojojames:
Right, sounds good.

noctuid:
Should be much faster now (went from 0.05 for all keybindings in my init to about 0.01). Please try and let me know if it's acceptable.

jojojames:
Nice! Sounds good. I'll try and give it a shot this weekend or next.

jojojames:
Seems fine to me, I think it was only an extra 5-10 ms with annalist, so I'm happy with that.

noctuid:
Alright, I've merged the branch to master. I'll make a PR to melpa and update the cask file once it's merged. Thanks for testing.

noctuid:
Can you rerun the travis build? Hopefully should pass now.

jojojames:
Looks good now. Lemme know when it's ready.

noctuid:
Yeah, I don't have anything else to add. I'll try to check sometime this week to make sure nothing is broken, and then it should be ready.

noctuid:
I've tried it out after running `(evil-collection-init)` and don't see anything horribly broken (and it is pretty well tested at this point). I did notice that I need to change how I'm handling the "previous" definition, so that it will stay as nil if the key was previously unbound and you re-evaluate an `evil-collection-init` call, but I think it should be fine to merge. Sorry this has taken so long!

noctuid:
And the issue I mentioned is fixed now. I don't see any other issues.

jojojames:
Awesome!! Thanks! Looks good.

PR #192, opened 2018-08-21
Add beginend

balajisivaraman:
[Beginend](https://github.com/DamienCassou/beginend) is a great package which replaces the normal `M-<` and `M->` bindings with better ones that make sense in the context of the current mode. For example, `M-<` in a Magit status buffer goes to the  first section instead of the first line. (There are numerous other modes supported.)

In Evil, `gg` and `G` has been bound to `evil-goto-first-line` and `evil-goto-line` respectively. It made sense to me that if I have `beginend-mode` enable for my Emacs, I would want `gg` and `G` to behave like `M-<` and `M->` does in the normal Emacs state. With this commit, I have added support for that behaviour.

I'm not really sure whether the current implementation is the best way to do it. But there are a few advantages we get: namely pressing `gg` once takes you to the overridden beginning of buffer. Pressing it again takes you the actual beginning of buffer. And repeated presses behave as expected. (This is how `beginend` behaves with `M-<`.)

I'll be open to suggestions for modifications, of course.

Thank you in advance.

### DIFF
--- a/evil-collection-beginend.el
+++ b/evil-collection-beginend.el
@@ -30,6 +30,8 @@
 (require 'evil-collection)
 (require 'beginend nil t)
 
+(defvar beginend-modes)
+
 (defun evil-collection-beginend-setup ()
   "Set up `evil' bindings for `beginend'."
   (mapc (lambda (pair)
@@ -37,8 +39,8 @@
                  (mode-name (symbol-name mode))
                  (map-name (intern (format "%s-map" mode-name))))
             (evil-collection-define-key 'normal map-name
-              "gg" 'beginning-of-buffer
-              "G" 'end-of-buffer)))
+              "gg" (command-remapping 'beginning-of-buffer nil (eval map-name))
+              "G" (command-remapping 'end-of-buffer nil (eval map-name)))))
         beginend-modes))
 
 (provide 'evil-collection-beginend)
--- a/evil-collection-beginend.el
+++ b/evil-collection-beginend.el
@@ -1,0 +1,46 @@
+;;; evil-collection-beginend.el --- Bindings for `beginend-mode' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018 James Nguyen
+
+;; Author: James Nguyen <james@jojojames.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <ambrevar@gmail.com>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, emacs, tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;; Bindings for `beginend-mode'
+
+;;; Code:
+(require 'evil-collection)
+(require 'beginend)
+
+(defun evil-collection-beginend-setup ()
+  "Set up `evil' bindings for `beginend'."
+  (mapc (lambda (pair)
+          (let* ((mode (cdr pair))
+                 (mode-name (symbol-name mode))
+                 (map-name (intern (format "%s-map" mode-name))))
+            (evil-collection-define-key 'normal map-name
+              "gg" 'beginning-of-buffer
+              "G" 'end-of-buffer)
+            ))
+        beginend-modes))
+
+(provide 'evil-collection-beginend)
+;;; evil-collection-beginend.el ends here
--- a/evil-collection-beginend.el
+++ b/evil-collection-beginend.el
@@ -28,7 +28,7 @@
 
 ;;; Code:
 (require 'evil-collection)
-(require 'beginend)
+(require 'beginend nil t)
 
 (defun evil-collection-beginend-setup ()
   "Set up `evil' bindings for `beginend'."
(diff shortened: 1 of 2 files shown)


evil-collection-beginend.el
```diff
@@ -0,0 +1,46 @@
...
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;;; Bindings for `beginend-mode'
+
+;;; Code:
+(require 'evil-collection)
+(require 'beginend)
```
jojojames: This should be (require 'beginend nil t).

evil-collection-beginend.el
```diff
@@ -0,0 +1,46 @@
...
+(defun evil-collection-beginend-setup ()
+  "Set up `evil' bindings for `beginend'."
+  (mapc (lambda (pair)
+          (let* ((mode (cdr pair))
+                 (mode-name (symbol-name mode))
+                 (map-name (intern (format "%s-map" mode-name))))
+            (evil-collection-define-key 'normal map-name
+              "gg" 'beginning-of-buffer
+              "G" 'end-of-buffer)
+            ))
```
jojojames: Can probably avoid the trailing )).

jojojames:
Does anyone else use beginend? I don't have a problem merging if it makes sense for people that use beginend. @Ambrevar ?

Ambrevar:
I haven't tested but I think the main issue with the current implementation is that it's seems to be irreversible: turning off beginend will not have the desired effect.  Correct me if I'm wrong.

That said, I'm sure I completely get the intent of the package: it seems to me that many of the covered modes already implement this feature (e.g. Magit and mu4e).  Am I missing something?

balajisivaraman:
I'm not sure what you mean by "turning off beginend". Is it turning off in `evil-collection` or disabling the mode altogether?

If I turn off Beginend in `evil-collection-mode-list`, but leave it enabled in my Emacs config, then the evil customizations aren't added to the respective beginend maps. However, `M-<` and `M->` work as expected.

If I enable Beginend in `evil-collection`, but disable it in my Emacs config, then the entire package is disabled altogether.

This overriding behaviour only works if I have it enabled in both places. I have tested this locally.

I'm not sure whether there's some Evil Collection specific thing I'm missing in my current implementation.

As for the latter part. I have `magit` enabled in `evil-collection-mode-list` and `evil-magit` installed. However, pressing `gg` or `G` only takes me to the first and last line of the Magit status buffer and not to the first and last sections like Beginend does. Is there some other customization I need to do to change this?

Beginend also covers other commonly used modes such as Dired. For example, pressing `M-<` in any Emacs Lisp file takes me to the first line after the initial commentary section and not to the first line of the file. I've found these behaviours to be really useful.

Ambrevar:
My bad for magit, it seems that I was wrong.

What I meant with "turning off" is the following:
- Enable the beginend minor mode.
- Load Evil Collection.
- Disable the beginend minor mode.

Now `G` and `gg` will still be bound to the beginend commands while they should not.

balajisivaraman:
Hmmm, I see what you mean. I just tried the following in `Dired` mode.

- Enable `beginend-dired-mode`
- Load Evil Collection by calling `evil-collection-init`
- Disable `beginend-dired-mode`

With the above sequence of events, the `G` and `gg` bindings are not even taking effect. This is true even if I close the buffer and open it and enable the minor mode. I can see the bindings in `beginend-dired-mode-map` but they don't take effect. I'm having a look at why this is the case.

However, if I enable `beginend-dired-mode` from Emacs init and then disable the minor mode in a dired buffer, it works as expected. `G` and `gg` regain their original functionalities.

jojojames:
I think @Ambrevar has a good point.

https://github.com/DamienCassou/beginend/blob/master/beginend.el

Taking a look at:

```
(defvar beginend-modes
  '(
    (mu4e-view-mode-hook . beginend-message-mode)
    (mu4e-compose-mode-hook . beginend-message-mode)
    )
  "List all beginend modes.
Each element has the form (STD-MODE-HOOK . BEGINEND-MODE).  STD-MODE-HOOK
is the standard mode hook (e.g., `dired-mode-hook') to which
BEGINEND-MODE (e.g., function `beginend-dired-mode') should be added.")
```

Looks like you want to add the keybind to beginend-message-mode (or whatever would be in the cdr of the cons pair.

So for example, given  (mu4e-view-mode-hook . beginend-message-mode), you'd probably want (evil-collection-define-key) to be defined on beginend-message-mode-map.

jojojames:
Deleted my previous message, upon looking at  https://github.com/DamienCassou/beginend/blob/master/beginend.el, I think the implementation is fine.

The keymap being bound should be the custom beginend-*specific*-mode-map.

jojojames:
OK to merge from me, @Ambrevar .

I'm still not too familiar with the package (only thing iffy for me is remapping to:

```
              "gg" 'beginning-of-buffer
              "G" 'end-of-buffer
```

but I think beginend probably does a [remap] on those somewhere.

fbergroth:
I think you can do it like this to directly bind to the proper functions:
```
"gg" (command-remapping 'beginning-of-buffer nil map-name)
"G" (command-remapping 'end-of-buffer nil map-name)
```

balajisivaraman:
@fbergroth, Thanks for the tip. I've changed it to use `command-remapping` instead. I've also fixed the build.

jojojames:
@balajisivaraman I wonder if you have to call 'evil-normalize-keymaps on beginend hook? Some of the modes require that since it seems evil doesn't automatically pull in the new keybindings.

balajisivaraman:
@jojojames, That also does not help. :disappointed: 

Enabling `beginend-global-mode` or adding the Beginend Minor Mode to the hook works, for eg: `(add-hook 'dired-mode-hook #'beginend-dired-mode)`. But if I open a buffer in the major mode (`dired-mode`) and then enable the Beginend minor mode, it does not work.

jojojames:
Lemme take a closer look when I have time.

jojojames:
Sorry it's taking so long. :)

>  But if I open a buffer in the major mode (dired-mode) and then enable the Beginend minor mode, it does not work.

Doesn't the same happen with normal emacs-Q?

dired-mode -> beginend-global-mode -> current dired mode won't have beginend keybindings

kill dired buffer and relaunch same dired -> dired buffer will have beginend keybindings

If that's true^, can you check if the same happens with your patch?

balajisivaraman:
@jojojames, Thank you for the guidance. I won't be able to check this for another week as I'm currently held up on something else. But once I get back, I will definitely check this out and let you know. Thanks again!

jojojames:
Closing this due to inactivity. Feel free to do another PR though. Thanks.

Issue #193, opened 2018-08-28
diff-mode: unable to quit with 'q'


wberrier:
I'm noticing that in the latest evil-collection-diff-mode I can't press `q` to quit diff mode.

It seems like this would be a good addition?

NOTE: I'm accessing diff-mode from vc.

jojojames:
Thanks!

PR #194, opened 2018-08-28
diff-mode: allow quitting with 'q'

wberrier:
Seems to be standard for bindings for other modes

This is a fix for #193


jojojames:
Thanks!

PR #195, opened 2018-08-28
log-view: start in normal mode by default

wberrier:
Must specify for the specific log-view modes


jojojames:
I might move these to specific evil-collection-* files but for now this is good! Thanks!

PR #198, opened 2018-09-09
add 'gr' as a bind for `package-menu-refresh' in package-menu-mode

jtecca:
Adds 'gr' bind in normal mode in order to be able to call `package-menu-refresh' when in package-menu-mode.  In default emacs bindings, 'r' is used to refresh, but I thought 'gr' would be more appropriate based on the conventions in the evil-collection README.


jojojames:
Looks good to me. Thanks.

PR #199, opened 2018-09-12
Handle evil want keybinding

jojojames:
After is merged, we need to handle the latest changes.

https://github.com/emacs-evil/evil/pull/1087

We should probably tag the last commit that still requires evil-want-integration to nil.

@Ambrevar


evil-collection.el
```diff
@@ -38,16 +38,30 @@
...
+(if (featurep 'evil-want-keybinding)
+    (if evil-want-keybinding
         (display-warning
          '(evil-collection)
-         "Make sure to set `evil-want-integration' to nil before loading evil \
-or evil-collection.")
+         "Make sure to set `evil-want-keybinding' to nil before loading evil \
+or evil-collection.\
+\n
+See https://github.com/emacs-evil/evil-collection/issues/60 for more details. ")
```
Ambrevar: Spurious trailing space.
Also why the `\`?  I think
```
or evil-collection.

See...
```
would be more readable.

Ambrevar:
Looks good to me!

evil-collection.el
jojojames: The formatting was better for me with that^.

PR #202, opened 2018-09-23
Add a keybinding for eglot-rename

endoffile78:


jojojames:
Wonder if this is a good key to bind for *users*. "gr" looks good to me and is intuitive in my opinion. gR also works for me.

@Ambrevar @noctuid ?

Ambrevar:
`gr` is often used for other things, and the `g` prefix is often used for movements.
Maybe `zR`?  Does anyone know what Vim does for those things?

noctuid:
I feel like this might be a place where it might be good to have something like localleader, but I don't have any great suggestions. `gr` Does seem confusing, and `zR` is pretty hard to type. Might be better to leave this binding to the user. I'd have to think about it more to give a better suggestion.

jojojames:
Here are some usages of g[] that aren't movements so it's probably not the worst thing in the world for it to be on gr.

```
It is bound to g a, g 8, C-x =.

(what-cursor-position &optional DETAIL)


It is bound to g i.

(evil-insert-resume COUNT)


It is bound to g u.

(evil-downcase BEG END &optional TYPE)


It is bound to g v.

(evil-visual-restore)


It is bound to g w.

(evil-fill BEG END)

```

Others, feel free to comment too. @fbergroth @A6GibKm @jgkamat @edkolev @Dickby  etc

jgkamat:
I personally wouldn't like `gr`, if only because it conflicts with [evil replacewithregister](https://github.com/Dewdrops/evil-ReplaceWithRegister) (I personally use `gr` rather than `gR`), and that's useful almost everywhere (and not just when eglot is enabled).

I think this kind of thing would make the most sense under a leader key (it's not needed so often it should be under 'g', but often enough you don't want to M-x it), but unless we have one, `gr` seems good to me.

edkolev:
I don't think vim has a key binding for "rename" (or other refactorings).

I believe it's fine to not bind `eglot-rename` to anything, it's not possible to bind every command that ever existed to a key even if we wanted to.

endoffile78:
I think binding it under a leader key could be a good solution if you guys don't like binding it to `gr`.

jojojames:
@endoffile78 Sounds good, thanks for PR'ing.

PR #204, opened 2018-10-15
Autoload "evil-collection-*-setup" functions

ezchi:
With autoload, it doesn't need "require" the package before call
evil-collection-*-setup function.

### DIFF
--- a/evil-collection-ag.el
+++ b/evil-collection-ag.el
@@ -32,6 +32,7 @@
 
 (defconst evil-collection-ag-maps '(ag-mode-map))
 
+;;;###autoload
 (defun evil-collection-ag-setup ()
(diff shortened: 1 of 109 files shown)


Ambrevar:
Which `require`?  What's the practical advantage of adding an autoload?

ezchi:
Without autoload, you need to require the package first. 

For example: 

    (require 'evil-collection-dired)
    (evil-collection-dired-setup)

With autoload, the "require" line is not needed.

Ambrevar:
You mean in case you don't want to use `evil-collection-init`?
In this case, I agree your changes could be useful.  @jojojames ?

jojojames:
LGTM

PR #207, opened 2018-11-23
Let "dd" mark line for deletion in wgrep

angerborn:
I think this is a decent binding since dd deletes a line normally and this will mark a line to be deleted when wgrep finishes.


terlar:
Se this: https://github.com/emacs-evil/evil-collection/pull/123

I have not had time to look into the feedback, but if you do, feel free.

Binding dd like that will make the other d operations to stop working

angerborn:
Thanks, you're right, it does break.

PR #208, opened 2018-11-27
Add setup hook to ediff-keymap-setup-hook

ezchi:
As the 'evil-collection-ediff-startup-hook' remaps the keybind, add it to
'ediff-keymap-setup-hook' instead of 'ediff-startup-hook'

### DIFF
--- a/evil-collection-ediff.el
+++ b/evil-collection-ediff.el
@@ -181,7 +181,7 @@
   "Initialize evil-ediff."
   (interactive)
   (evil-set-initial-state 'ediff-mode 'normal)
-  (add-hook 'ediff-startup-hook 'evil-collection-ediff-startup-hook)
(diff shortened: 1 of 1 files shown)


jojojames:
This looks ok to me but I'm not an expert on ediff. Is there a functional difference @ezchi?

Deferring to @justbur and @Ambrevar

Ambrevar:
Just had a quick look at Ediff's source code, and this seems correct.
Could be wrong though, as Ediff's startup code is rather complex.
@justbur ?

justbur:
The change looks ok to me, but the problem doesn't make a lot of sense to me. If anything, `ediff-startup-hook` should account for more cases rather than fewer, since it's run later. 

In any case, I think this change makes sense.

jojojames:
Thanks @ezchi .

> And tried with the same spacemacs setup, it works.

Would be curious how evil-collection is working with spacemacs too.

PR #210, opened 2018-12-06
Add tetris

MaxSt:
Evil keybindings for tetris mode
(https://www.emacswiki.org/emacs/TetrisMode)

### DIFF
--- a/evil-collection-tetris.el
+++ b/evil-collection-tetris.el
@@ -1,0 +1,56 @@
+;;; evil-collection-tetris.el --- Bindings for `tetris' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018 James Nguyen
+
+;; Author: MaxSt
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <mail@ambrevar.xyz>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, emacs, tools
+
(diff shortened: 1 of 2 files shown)


evil-collection-tetris.el
```diff
@@ -0,0 +1,55 @@
...
+(defun evil-collection-tetris-setup ()
+  "Set up `evil' bindings for `tetris'."
+  (evil-set-initial-state 'tetris-mode 'normal)
+  (evil-collection-define-key 'normal 'tetris-mode-map
+    "h" 'tetris-move-left
+    "l" 'tetris-move-right
+    "k" 'tetris-rotate-next
+    "K" 'tetris-rotate-prev
+    "j" 'tetris-move-down
+    "J" 'tetris-move-bottom
```
jojojames: Does this move the piece all the way to the bottom? 'G' might be better or both 'G' and 'J'. Don't have a strong opinion though.

evil-collection-tetris.el
```diff
@@ -0,0 +1,55 @@
+;;; evil-collection-tetris.el --- Bindings for `tetris' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018 James Nguyen
+
+;; Author: James Nguyen <james@jojojames.com>
```
jojojames: You can set yourself as the author.

evil-collection-tetris.el
MaxSt: :+1: Adding 'G' is a good idea, but i think keeping 'J' is also fine, because its easier to reach when playing.

Issue #215, opened 2019-01-01
evil-want-keybinding warning due to evil-leader


willbush:
I'm getting the following warning despite using the recommended `use-package`
examples in the readme.

```
Warning (evil-collection): `evil-want-keybinding' was set to nil but not before loading evil.

Make sure to set `evil-want-keybinding' to nil before loading evil or evil-collection.

See https://github.com/emacs-evil/evil-collection/issues/60 for more details.
```

I found to fix it I had to `(setq evil-want-keybinding nil)` before loading
evil-leader, but I have no idea why.

Here is a minimal emacs configuration to demonstrate the issue:

```elisp
(require 'package)

(setq package-enable-at-startup nil)

(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/"))

(package-initialize)

;; bootstrap use-package
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))

(require 'use-package)

(use-package evil-leader
  :ensure t
  :init
  ;; (setq evil-want-keybinding nil)
  (global-evil-leader-mode))

(use-package evil
  :ensure t
  :init
  (setq evil-want-integration t) ;; This is optional since it's already set to t by default.
  (setq evil-want-keybinding nil)
  :config
  (evil-mode 1))

(use-package evil-collection
  :after evil
  :ensure t
  :config
  (evil-collection-init))

```
Uncomment the line following line in the use-package init block for `evil-leader` to fix the issue.

```
  ;; (setq evil-want-keybinding nil)
```

Loading `evil-leader` after `evil` will also fix the issue, but consider the
following from the [evil-leader readme](https://github.com/cofi/evil-leader):

>Note: You should enable global-evil-leader-mode before you enable evil-mode,
 otherwise evil-leader won’t be enabled in initial buffers (*scratch*,
 *Messages*, …).

## Emacs version:

GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.1) of 2018-12-25

Also note I'm using all latest packages from Melpa since my testing involved `rm
-rf ~/.emacs.d` and using the above minimal config to reload them.

jojojames:
Thanks for the detailed report. I think evil-leader probably loads evil on its own which is why your problem comes up. 

I think, if you're using evil-leader, you'll have to set that variable before (like you did) evil-leader loads.

willbush:
[...]

PR #218, opened 2019-01-03
Add mention of other packages loading evil to readme

willbush:
See the following: https://github.com/emacs-evil/evil-collection/issues/215

I ended up just mentioning it right after it says when those variables must be set. If you rather it be an addition to the Faq, let me know.


jojojames:
Thanks, looks good.

PR #222, opened 2019-01-16
mu4e queued mail status needs to be dynamic

pjones:
This fixes a fairly major bug where mail could be queued but not shown
to the user.  (This happened to me, I had a month of email sitting in
the queue!)

Since the number of messages in the queue can change, and the status
of queuing mail or direct mail can also change, the display should not
be in a defvar which only gets evaluated once when the library is
first loaded.

This change makes it so the queue status is updated each time the main
mu4e view is drawn.


Ambrevar:
Looks good to me, thanks!

PR #234, opened 2019-03-31
Bind proced-why

yilinwei:
Bind `proced-why` to `?`.


jojojames:
Maybe g?

Is ? not used in proced?

yilinwei:
@jojojames 

Hmm, I think you may be right - it should probably be bound to `proced-help`.

Main reason for the PR is that I got an error message with something along the lines of `error sending signal, press ?` so when doing describe key I had `?` bound to something from `evil`. I'll close this for now until I can repro the error and take a closer look.

PR #236, opened 2019-04-18
Fix non-prefix key errors

amosbird:
byte compilation results in 
```
evil-collection error: (error Key sequence g o starts with non-prefix key g)
```

### DIFF
--- a/evil-collection-dired.el
+++ b/evil-collection-dired.el
@@ -35,6 +35,11 @@
 ;;;###autoload
 (defun evil-collection-dired-setup ()
   "Set up `evil' bindings for `dired'."
+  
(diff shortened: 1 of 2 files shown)


jojojames:
When I byte compile it with make, it works fine for me.

Maybe an emacs -Q repro would be good first.

Thanks!

amosbird:
Then I guess it's not reproducible.

dvcrn:
Having the same issue here. How did you resolve it?

jojojames:
It's probably something with your config. Emacs -Q, disable dired in evil-collection's mode list, etc.

dvcrn:
Figured it out! I had a dired keybinding that was conflicting with evil-collection. Removed it and now everything is working fine

PR #243, opened 2019-07-05
Load lsp-ui-imenu instead of lsp-ui

hlissner:
This causes lsp-ui to be loaded twice.

To reproduce (in vanilla emacs):

```elisp
(with-eval-after-load 'lsp-ui-imenu ; simulates evil-collection-lsp-ui-imenu
  (require 'lsp-ui))

(with-eval-after-load 'lsp-ui
  (message "hello"))

(require 'lsp-ui)  ;; -> "hello [2 times]"
```

This complicates configuring lsp-ui with `with-eval-after-load`.


jojojames:
Nice, thanks.

PR #245, opened 2019-07-25
notmuch: Bind "|" to notmuch-show-pipe-message in notmuch-tree.

jellelicht:
this is also what would be the case in the emacs state, so it is just a matter of making things consistent


jojojames:
Thanks, looks good.

PR #247, opened 2019-08-16
Catch errors in evil-collection-define-key

justbur:
Like errors related to prefix keys already being defined


Ambrevar:
Looks good to me.
I suppose those errors are benign enough that we don't want to use `warn` instead of `message`.

@jojojames ?

justbur:
The annoying thing that can happen is if you get a bad keybinding in one of the `after-load-functions`. That can prevent loading a file

jojojames:
Looks OK to me.

PR #254, opened 2019-10-05
Fix "h" in grep mode

rgrinberg:
By default this is bound to describe-mode, which is easily accessible
through other means.


Ambrevar:
Thanks!

PR #255, opened 2019-10-05
Fix recompile binding for grep mode

rgrinberg:
It should be "gr" rather than "g". While this fix works, other bindings like `gg` remain broken for example. What's the correct way to make `g <x>` be forwarded to the appropriate normal mode command?

### DIFF
--- a/evil-collection-compile.el
+++ b/evil-collection-compile.el
@@ -30,32 +30,35 @@
 (require 'evil-collection)
 (require 'compile)
 
-(defconst evil-collection-compile-maps '(compilation-mode-map))
+(defconst evil-collection-compile-maps
+  '(compilation-mode-map compilation-minor-mode-map))
 
 ;;;###autoload
 (defun evil-collection-compile-setup ()
   "Set up `evil' bindings for `compile'."
   (evil-set-initial-state 'compilation-mode 'normal)
 
-  (evil-collection-define-key 'normal 'compilation-mode-map
-    "g?" 'describe-mode
-    "?" evil-collection-evil-search-backward
-    "gg" 'evil-goto-first-line
-    "0" 'evil-digit-argument-or-evil-beginning-of-line
-    [mouse-2] 'compile-goto-error
-    [follow-link] 'mouse-face
-    (kbd "RET") 'compile-goto-error
+  (dolist (map evil-collection-compile-maps)
+    (evil-collection-define-key 'normal map
+      "g?" 'describe-mode
+      "?" evil-collection-evil-search-backward
+      "gg" 'evil-goto-first-line
+      "0" 'evil-digit-argument-or-evil-beginning-of-line
+      [mouse-2] 'compile-goto-error
+      [follow-link] 'mouse-face
+      (kbd "RET") 'compile-goto-error
 
-    "go" 'compilation-display-error
-    (kbd "S-<return>") 'compilation-display-error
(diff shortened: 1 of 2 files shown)


Ambrevar:
Hmm, I cannot reproduce.  For me `g` is unbound.

By the way I merged your other issue with `h` but now I'm realizaing I cannot reproduce that issue either.
Wouldn't you have another minor mode by anything?  Can you `C-h b` to see which mode is binding what?

rgrinberg:
Thanks for the hint and apologies for the hasty patch.

 I reverted this commit and the one you already merged and I think I got to the bottom of this. It seems like evil-collections simply omitted patching the compilations-minor-mode-map. This map should be patched in addition to compilation-minor-mode.

The way my last commit does it makes sure the same keybindings are active in both, but it’s probably overkill as not all keys are overridden in both maps. Nevertheless, it’s probably the simplest way to ensure consistency so unless it has any concrete problems, that seems like the best solution to me.

With my latest patch gr, gg, etc all work. However, keybindings like ga are still absent. This is a bit disappointing as I would expect all normal mode key maps to work.

Ambrevar:
Nice!

Any clue why I cannot reproduce the issue?  `compilation-minor-mode` is not loaded for me in `grep` buffers.

Ambrevar:
`ga` is bound to `what-cursor-position` in `evil-normal-state-map`.  It should be bound in the grep buffer, I'm not sure why it's not.

rgrinberg:
> Any clue why I cannot reproduce the issue? compilation-minor-mode is not loaded for me in grep buffers.

No clue unfortunately. How are you launching grep? I'm able to reproduce it consistently with `grep-find` for example. I'm fairly certain I got the issue nailed this time around by the way. Here's where emacs override `g` in this compilaion-minor-mode-map btw: https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/compile.el#L1935

rgrinberg:
@Ambrevar Here's the bit where `grep` is [inheriting](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/grep.el#L272) from the `compilation-minor-mode-map`. So it seems like the mode isn't really required, but the map is still used.

Ambrevar:
@jojojames Any clue why "g r" isn't bound for @rgrinberg?  For me it's bound in the `t` minor map, not sure what that means.

rgrinberg:
> Any clue why "g r" isn't bound for @rgrinberg? For me it's bound in the t minor map, not sure what that means.

Which command is `g r` bound to for you?

Also, what is the value of `(keymap-parent grep-mode-map)` for you?

jojojames:
I skimmed through this so I might've missed something but maybe it would be helpful to start with an Emacs -Q recipe and go from there.

rgrinberg:
Sorry for the noise everyone.

jojojames:
@rgrinberg All good, thanks for looking into this stuff. :D

Ambrevar:
Wait, @rgrinberg why did you close this issue?  Is this resolved for you?  Should I merge still?

rgrinberg:
My time budget for this PR ran out. Also, I don’t mind keeping the fix just in my private fork. It works for me at least.

Ambrevar:
OK, then I'll revert the "h" fix like you did in the first commit.
Cheers!

PR #256, opened 2019-10-07
Fix notmuch-search-toggle-flagged

mnick:
Simple one line fix that prevented notmuch-search-toggle-flagged to work


jojojames:
Thanks!

PR #258, opened 2019-10-10
Load evil-collection-custom after cus-edit

justbur:
The cus-edit.el file contains the relevant map/functions, not custom.el.


jojojames:
Thanks!

PR #261, opened 2019-10-13
Add evil-ledger to list of third-party packages

michaelbogdan:


Ambrevar:
Thanks!
What about merging evil-ledger into evil-collection then?

michaelbogdan:
Merging is what I initially thought of, but adding it to the README seemed like a good idea to do first.

`evil-ledger` introduces some new functions to work with transactions similar to text objects and a separate minor mode `evil-ledger-mode`. I wasn't sure whether this was ok within `evil-collection`, but then I saw that for example `evil-collection-restclient.el` does the exact thing.

Also, `evil-ledger` is not my work, so I am not sure what the proper etiquette is to contribute it. Should I fork `evil-collection`, set up `evil-ledger` as remote, pull the commits onto a separate branch and edit the code there to fit with the rest of `evil-collection`? This would preserve the original commits and show that @atheriel did the work of developing the script.

Ambrevar:
I think we would only merge evil-ledger into evil-collection if @atheriel agrees to it, otherwise we wouldn't want to duplicate the code lest we run out of sync and confuse the community.

@atheriel?

atheriel:
Hi all, it's nice to see some interest in my package. I'm happy to see it get bundled with this project, but a couple things come to mind:

1. The existing `evil-ledger` package on MELPA actually gets a substantial number of downloads, so I don't want to take it down and break existing configurations; and
2. I'd like to ensure that issues and feature requests filed here relevant to `evil-ledger` get reported upstream.

If both of these are acceptable to you folks, then I don't see any issue with pasting the upstream file into this project (along with its metadata). Both projects appear to be GPLv3, too.

I would not worry too much about keeping it in sync with upstream -- there hasn't been an update in more than a year and I doubt this will change all that much. Plus, if in the future I am unable to maintain the package, this project is a natural home for it.

michaelbogdan:
@Ambrevar what is the correct procedure for other `evil` packages I find? Should I open a new issue or make pull request like I did here and mention the original author? Because I just found [`evil-markdown`](https://github.com/Somelauw/evil-markdown) which would be a decent fit I think.

Ambrevar:
@michaelbogdan: Sure, what you did in this pull request is perfect.

Ambrevar:
@atheriel: If you want to keep upstream active, then I'd rather not include the project here.  It's confusing for people who use both.
But what we can do is _move_ evil-ledger here and give you access rights.
We would "archive" the evil-ledger repo (in the GitHub sense), update the description to point to here.  Does that make sense?

The decision is all yours ;)

jojojames:
>    The existing evil-ledger package on MELPA actually gets a substantial number of downloads, so I don't want to take it down and break existing configurations; and

I think we can do it similar to evil-ediff https://github.com/emacs-evil/evil-ediff which has been merged and deprecated in the original repo. @justbur is of course, welcome to make any changes he wants to it here.

Similarly, with https://github.com/JorisE/evil-mu4e , @JorisE is welcome to make changes here.

>    I'd like to ensure that issues and feature requests filed here relevant to evil-ledger get reported upstream.

I agree with @Ambrevar . I'd rather give you commit access as a collaborator to help maintain the project here and archive the old project (continue keeping it on melpa of course).

> Because I just found evil-markdown which would be a decent fit I think.

That one and https://github.com/Somelauw/evil-org-mode seem like good candidates if there was any interest on @Somelauw 's side.

PR #262, opened 2019-10-15
Add bindings for dired-subtree

michaelbogdan:
`dired-subtree` allows the insertion of a subtree in place instead of at the bottom of the buffer, as `dired-maybe-insert-subdir` would. Unfortunately `dired-subtree` does not come with a minor mode to cleanly separate keybindings, so I inserted a `featurep` check if `dired-subtree` is available, as it makes sense to overwrite one or more bindings from `evil-dired`.

`dired-subtree` is interesting by itself as it makes navigating directories a bit easier, but it is also a requirement for `dired-sidebar`, a candidate for addition to `evil-collection`.

P.S.: This is my first proper pull request, please advise if I'm doing something wrong.

### DIFF
--- a/modes/dired/evil-collection-dired.el
+++ b/modes/dired/evil-collection-dired.el
@@ -189,7 +189,7 @@
     ";e" 'epa-dired-do-encrypt)
 
     ;; dired-subtree commands
-    (when (featurep 'dired-subtree)
+    (with-eval-after-load 'dired-subtree
 	(evil-collection-define-key 'normal 'dired-mode-map
 	    "I" 'dired-subtree-toggle
 	    (kbd "TAB") 'dired-subtree-cycle
--- a/modes/dired/evil-collection-dired.el
+++ b/modes/dired/evil-collection-dired.el
@@ -191,7 +191,6 @@
     ;; dired-subtree commands
     (with-eval-after-load 'dired-subtree
 	(evil-collection-define-key 'normal 'dired-mode-map
-	    "I" 'dired-subtree-toggle
 	    (kbd "TAB") 'dired-subtree-cycle
 	    "gh" 'dired-subtree-up
 	    "gl" 'dired-subtree-down
--- a/modes/dired/evil-collection-dired.el
+++ b/modes/dired/evil-collection-dired.el
@@ -186,7 +186,17 @@
     ";d" 'epa-dired-do-decrypt
     ";v" 'epa-dired-do-verify
     ";s" 'epa-dired-do-sign
-    ";e" 'epa-dired-do-encrypt))
+    ";e" 'epa-dired-do-encrypt)
+
+    ;; dired-subtree commands
+    (when (featurep 'dired-subtree)
+	(evil-collection-define-key 'normal 'dired-mode-map
+	    "I" 'dired-subtree-toggle
+	    (kbd "TAB") 'dired-subtree-cycle
+	    "gh" 'dired-subtree-up
+	    "gl" 'dired-subtree-down
+	    (kbd "M-j") 'dired-subtree-next-sibling
+	    (kbd "M-k") 'dired-subtree-previous-sibling)))
 
 (provide 'evil-collection-dired)
 ;;; evil-collection-dired.el ends here


Ambrevar:
I think we should use `with-eval-after-load` instead of `featurep`, this way the bindings would be set even if dired-subtree is loaded after evil-collection.

michaelbogdan:
That is definitely an improvement and something I/we can do with other, similar packages.

modes/dired/evil-collection-dired.el
```diff
@@ -186,7 +186,17 @@
     ";d" 'epa-dired-do-decrypt
     ";v" 'epa-dired-do-verify
     ";s" 'epa-dired-do-sign
-    ";e" 'epa-dired-do-encrypt))
+    ";e" 'epa-dired-do-encrypt)
+
+    ;; dired-subtree commands
+    (with-eval-after-load 'dired-subtree
+	(evil-collection-define-key 'normal 'dired-mode-map
+	    "I" 'dired-subtree-toggle
```
jojojames: "I" may be too intrusive since it's replacing:

```
(dired-maybe-insert-subdir DIRNAME &optional SWITCHES NO-ERROR-IF-NOT-DIR-P)
```

which I think can be useful even with dired-subtree.

modes/dired/evil-collection-dired.el
```diff
@@ -186,7 +186,17 @@
...
+	    (kbd "TAB") 'dired-subtree-cycle
```
jojojames: Does dired-subtree bind TAB on its own?

I just looked into my own bindings and see that I use dired-subtree-toggle instead of cycle for both my dired and dired-sidebar setup.

modes/dired/evil-collection-dired.el
```diff
@@ -186,7 +186,17 @@
...
+	    "gj" 'dired-subtree-up
```
jojojames: Doesn't this go up so should be gk.

modes/dired/evil-collection-dired.el
```diff
@@ -186,7 +186,17 @@
...
+	    "gl" 'dired-subtree-down
```
jojojames: gl?

modes/dired/evil-collection-dired.el
michaelbogdan: If the replacement is the issue, we could bind `M-i` to `dired-subtree-toggle`. The key `i` for `insert` is just too tempting not to use.

modes/dired/evil-collection-dired.el
michaelbogdan: `dired-subtree` doesn't bind any keys on its own, it just provides a set of 16 functions.

modes/dired/evil-collection-dired.el
michaelbogdan: This is a plain typo, sorry. It should be `gh` as I wanted it to behave similar to `org-mode` where `gh` brings me one indentation level up and conversely `gl` brings me one indentation level down.

modes/dired/evil-collection-dired.el
jojojames: Hmnn wonder if I should switch to cycle instead of toggle for dired-sidebar. FWIW, they're about the same in functionality. I'm guessing something about wrapping cycle makes dired-sidebar unable to unnest all directories which makes cycle the same as toggle.

This is OK to me.

modes/dired/evil-collection-dired.el
jojojames: Yeah, I think I is too much, don't really have an opinion on toggle. I actually think most people probably just use only one of these functions and wouldn't bother with the other.

Maybe a defcustom on the one to bind works too.

modes/dired/evil-collection-dired.el
jojojames: Sounds good, gh/gl OK with me.

BTW, thanks for reminding me of dired-sidebar. I finally summoned the effort to bring it to evil-collection.

modes/dired/evil-collection-dired.el
michaelbogdan: A defcustom for a single key barely makes sense, I think. Maybe if it bundles some other functions, something like "prefer subtree over subdirectory" or such.

modes/dired/evil-collection-dired.el
michaelbogdan: I just saw that you are the author of `dired-sidebar`, made me wonder why it wasn't part of `evil-collection` before. :) Nice little package, I really like that it carries over any settings from plain `dired`.

modes/dired/evil-collection-dired.el
jojojames: Sure, I'm not sure about M-i though, do you use dired-subtree-toggle?

I think we can just leave it out, probably.

modes/dired/evil-collection-dired.el
jojojames: This should be gh right?

modes/dired/evil-collection-dired.el
michaelbogdan: Exclusively. I think I'll just leave it out for now, we can revisit the keybindings in the future.

My main goal was to get one more package into the collection for review and consistency.

I'll fix my typo later and amend the commits to reflect the feedback a bit later.

modes/dired/evil-collection-dired.el
michaelbogdan: Yes, fixed with --amend and force push now.

michaelbogdan:
Ready to merge now, I think.

jojojames:
Thanks!

jojojames:
I got a chance to try cycle instead of toggle and I think for dired, it makes more sense to use toggle. cycle is actually fairly slow in most use cases and, for me, freezed emacs on most directories.

What do you think @michaelbogdan ?

PR #268, opened 2019-10-17
Add evil-markdown to third-party packages in readme

michaelbogdan:
Making the list more complete.

Also pinging @Somelauw on the chance that they want to integrate their package into `evil-collection`.


jojojames:
Thanks.

PR #275, opened 2019-10-28
Disable <tab>-based bindings by default

### LINKED ISSUE #274: Disable <tab>-based bindings by default
@Ambrevar [made these bindings optional](https://github.com/emacs-evil/evil-collection/commit/28eb46bd4439630bcb3ef4d65c2bff7d6fc7ffe4), which is great, but I'd go one step further and disable them by default.

The reasoning being:

- I'd estimate many more people use org-mode then pure outline-mode so it's more important to have good bindings in org.

- I'd say ``org-shifttab`` is more useful then  ``outline-show-all``, cause the former is a 3-way toggle and the latter is one-way-only.

I might be mistaken in my assertion of outline-mode cause maybe there's other much-used modes derived from it that are commonly used.

Ideally we'd have both bindings available in their respective modes, but [this thread](https://github.com/emacs-evil/evil-collection/issues/53) leads me to believe that's not so easy.

Note: I'm relatively new to emacs, so I might very well be missing a lot.

magnetophon:
fixes #274

### DIFF
--- a/modes/outline/evil-collection-outline.el
+++ b/modes/outline/evil-collection-outline.el
@@ -30,12 +30,12 @@
 (require 'evil-collection)
 (require 'outline)
 
-(defcustom evil-collection-outline-bind-tab-p t
+(defcustom evil-collection-outline-bind-tab-p nil
   "Enable <tab>-based bindings in Outline mode.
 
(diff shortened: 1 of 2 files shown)


Ambrevar:
Looking good, thanks a lot for this PR!

PR #277, opened 2019-11-03
Add dired-filter

Silex:
I originally created `evil-collection-dired-filter` directory, then realised it belongs to dired.

It relies on `dired-filter-mark-map` being assigned last so it overwrites the default bindings like `**` etc.


jojojames:
Looks good to me.

PR #279, opened 2019-11-05
Improve tablist bindings

Silex:
This move `tablist-do-kill-lines` to `K` like in `dired` and `ibuffer`.


jojojames:
Sounds good to me!

PR #283, opened 2019-11-23
Fix conflicting keybindings in gnus configuration

0x28:
There were two different `C-k` keybindings, so I removed the one for `gnus-summary-kill-same-subject`. Now `C-j` and `C-k` behave in a similar way.

The `k` keybinding was conflicting with `evil-previous-line`, so I removed that one too.


Ambrevar:
Thanks!

PR #286, opened 2019-12-10
vterm-exit-functions now sends event as 2nd arg.

cnsunyour:
emacs-libvterm@376db7cf416 had a breaking change for
`vterm-exit-functions` which sends the event as the second argument.


jojojames:
Sounds good, thanks!

PR #291, opened 2019-12-31
Ensure compatibility with terminal for various modes.

crocket:
I rebound several keybindings to make evil-collection work in terminal.

### DIFF
--- a/modes/info/evil-collection-info.el
+++ b/modes/info/evil-collection-info.el
@@ -49,11 +49,11 @@
     "0" 'evil-digit-argument-or-evil-beginning-of-line
     (kbd "M-h") 'Info-help              ; "h"
     (kbd "C-t") 'Info-history-back      ; "l"
-    (kbd "C-o") 'Info-history-back
+    (kbd "gh") 'Info-history-back
     " " 'Info-scroll-up
     (kbd "C-]") 'Info-follow-nearest-node
     (kbd "DEL") 'Info-scroll-down
-    (kbd "g TAB") 'Info-history-forward
+    (kbd "gl") 'Info-history-forward
 
     "d" 'Info-directory
     "u" 'Info-up
--- a/modes/info/evil-collection-info.el
+++ b/modes/info/evil-collection-info.el
@@ -42,7 +42,7 @@
   (evil-collection-define-key 'normal 'Info-mode-map
     "l" 'evil-forward-char
     "h" 'evil-backward-char
-    (kbd "<tab>") 'Info-next-reference
+    (kbd "TAB") 'Info-next-reference
     (kbd "S-<tab>") 'Info-prev-reference
 
     ;; From evil-integration.el.
@@ -53,7 +53,7 @@
     " " 'Info-scroll-up
     (kbd "C-]") 'Info-follow-nearest-node
     (kbd "DEL") 'Info-scroll-down
-    (kbd "C-i") 'Info-history-forward
+    (kbd "g TAB") 'Info-history-forward
 
     "d" 'Info-directory
     "u" 'Info-up
--- a/modes/custom/evil-collection-custom.el
+++ b/modes/custom/evil-collection-custom.el
@@ -39,7 +39,7 @@
 
   (evil-collection-define-key 'normal 'custom-mode-map
     ;; motion
-    (kbd "<tab>") 'widget-forward
+    (kbd "TAB") 'widget-forward
     (kbd "S-<tab>") 'widget-backward
     (kbd "<backtab>") 'widget-backward
     (kbd "SPC") 'scroll-up-command
--- a/modes/dired/evil-collection-dired.el
+++ b/modes/dired/evil-collection-dired.el
@@ -134,7 +134,7 @@
     ;; open
     (kbd "RET") 'dired-find-file
     (kbd "S-<return>") 'dired-find-file-other-window
-    (kbd "M-<return>") 'dired-display-file
+    (kbd "M-RET") 'dired-display-file
     "gO" 'dired-find-file-other-window
     "go" 'dired-view-file
     ;; sort
--- a/modes/help/evil-collection-help.el
+++ b/modes/help/evil-collection-help.el
@@ -43,11 +43,11 @@
     (kbd "S-SPC") 'scroll-down-command
     (kbd "C-f") 'scroll-up-command
     (kbd "C-b") 'scroll-down-command
-    (kbd "<tab>") 'forward-button
+    (kbd "TAB") 'forward-button
     (kbd "<backtab>") 'backward-button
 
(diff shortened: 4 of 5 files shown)


jojojames:
Looks reasonable, what do you think @Ambrevar @noctuid ?

modes/info/evil-collection-info.el
```diff
@@ -53,7 +53,7 @@
     " " 'Info-scroll-up
     (kbd "C-]") 'Info-follow-nearest-node
     (kbd "DEL") 'Info-scroll-down
-    (kbd "C-i") 'Info-history-forward
+    (kbd "g TAB") 'Info-history-forward
```
yunhao94: you bind "g TAB" here, but leave "C-o" for Info-history-back, I think it's also a kind of inconsistency, because I think people tend to think "C-i" is the reverse version of "C-o". what do you think about bind "C-i/o" to Info-history-forward/back, and use "g C-i/o" for the evil-jump-foreward/backward?

jojojames: Does it make sense to keep C-i for GUIs?

yunhao94: IMHO it's more sensible to either keep C-i and C-o together, or remap them together and set a fallback like "g C-i/o" to the original evil-jump-foreward/backward, no matter for  GUI or TTY

crocket: `C-i` is TAB. I want `TAB` to behave similarly to `<tab>` for terminal compatibility.

yunhao94: there is a variable evil-want-C-i-jump, and its default is t, which  means TAB shouldn't behave similarly to <tab> by default, add a condition seems better

crocket:
Copying every vim keybinding is not going to work well with emacs. We want to make a better vim that cooperates with emacs keybindings.

`gh` and `gl` are ergonomically superior to `C-o` and `C-i`. Look at the commit.

modes/info/evil-collection-info.el
crocket: evil-collection ignores evil-want-C-i-jump because it has its own bindings.

Ambrevar:
I'm overall OK with these changes but first I'd like to bring up a few comments:

> gh and gl are ergonomically superior to C-o and C-i. Look at the commit.

Why?  `gh` / `gl` are harder to chain than `C-o` / `C-i`, and chaining matters for those browsing commands.

About the question of terminal compatibility vs. ergonomy, I think we need to figure out which one is more important.  I don't think that "terminals" are the way forward, and thus I'm not sure that sacrificing bindings / ergonomy / Vim-compatibility for the sake of terminal compatibility is worth it.

As a rule of thumb, I'd say "keep terminal compatibility whenever you can, if it's free; if it comes at a cost, keep GUI functionalities."

Thoughts?

crocket:
* Vim compatibility with tab key is already broken on GUI. Breaking it also on terminal is more consistent.
  * Either break vim compatibility on both GUI and terminal, or don't break it on both.
* I want to use emacs remotely. A remote ssh session is good for this.

modes/info/evil-collection-info.el
yunhao94: Thanks for your illustration! The consistency in my mind is about <kbd>C-i</kbd> with <kbd>C-o</kbd>, but not with <kbd>\<tab\></kbd>, and unfortunately terminal can't make use of <kbd>\<tab\></kbd>. It looks good to me, thanks for your work!

Ambrevar:
- Where is Vim compatibility broken precisely?

- For remote Emacs sessions, I suggest using TRAMP.  I know some users report it does not work well for them.  I think then fixing TRAMP might be a better option.  Ultimately, terminals are holding us back, it's high time we implemented proper solutions going forward :)

crocket:
I don't have access to GUI emacs on iPad.

jojojames:
> About the question of terminal compatibility vs. ergonomy, I think we need to figure out which one is more important. I don't think that "terminals" are the way forward, and thus I'm not sure that sacrificing bindings / ergonomy / Vim-compatibility for the sake of terminal compatibility is worth it.

I use terminal emacs quite often, through SSH + Tmux and I also agree with this. I think the terminal compatibility level should be around "here's a fallback in the terminal, it's not the best but you got it" and leave the most ideal version of the bindings for the GUI.

> As a rule of thumb, I'd say "keep terminal compatibility whenever you can, if it's free; if it comes at a cost, keep GUI functionalities."

I agree.

>    I don't have access to GUI emacs on iPad or android.

These keys just seem like a "what do we want to break?" type change right? If we merge this PR, TAB behaves consistently with regards to reference browsing in the Terminal but "jumping"/C-o/C-i consistency" breaks in the GUI. Lemme know if I have this wrong..

For most users, I guess they use the GUI more so I'm guessing which case are we optimizing for here?

crocket:
Adding more keybindings for reference browsing functions doesn't break anything.

I recommend binding `<` and `>` or `C-j` and `C-k` to emacs reference browsing commands.
I don't really need to use tab key for reference browsing in emacs.

yunhao94:
> These keys just seem like a "what do we want to break?" type change right? If we merge this PR, TAB behaves consistently with regards to reference browsing in the Terminal but "jumping"/C-o/C-i consistency" breaks in the GUI. Lemme know if I have this wrong.

Let's see what vanilla Emacs does:
  * <kbd>TAB</kbd>/<kbd>C-i</kbd> is `indent-for-tab-command` in the scratch buffer, `forward-button` in the help buffer, and `Info-next-reference` in the Info buffer
  *  <kbd>\<tab\></kbd> is often `TAB (translated from <tab>)` in GUI

There's no inconstancy, because <kbd>C-i</kbd> doesn't have the function like `evil-jump-forward`.

There's already a `evil-want-C-i-jump` option, so what do you think about add a `evil-collection-want-C-i-jump` option which behaves like:
  * when its value is `t`, make <kbd>C-i</kbd> the reverse of <kbd>C-o</kbd>, if <kbd>C-o</kbd> is remapped to functions like `help-go-back` and `Info-history-back`, which keeps the "jumping"/C-o/C-i consistency"
  * otherwise keep <kbd>TAB</kbd> as the reverse of <kbd>\<backtab\></kbd> (`forward-button`, `Info-next-reference`, etc), which is what vanilla Emacs does, and TAB behaves consistently with regards to reference browsing

crocket:
@yunhao94 With `evil-collection-want-C-i-jump`, you lose either `C-i` and `C-o` or `TAB` and Shift-`TAB` on terminal. By binding `<` and `>` or `C-j` and `C-k` to emacs reference browsing keys, you gain redundant reference browsing keys that will work even on terminal.

crocket:
I think we should close this pull request for now and discuss further in the issue.

Issue #292, opened 2020-01-01
Avoid binding `zb` in outline mode.


crocket:
In other modes, `zb` is bound to `evil-scroll-line-to-bottom` which I'm accustomed to.

I suggest rebinding `zB` and `zb` somewhere else in outline mode.

Here's my current setting.

```
(use-package evil-collection
  :after evil
  :config
  (evil-collection-init)
  (add-hook 'org-mode-hook
            (lambda ()
              (define-key evil-normal-state-local-map
                "zb" #'evil-scroll-line-to-bottom))))

```

Ambrevar:
Any suggestion for good bindings?  Outline mode has so many of them, it's really tricky to get them right :p

crocket:
`zd` and `zD` are vacant in outline mode and org mode.

Ambrevar:
Sounds good!

PR #295, opened 2020-01-04
Do not override `zb` in outline mode.

crocket:
`zb` is usually bound to `evil-scroll-line-to-bottom`. By moving `outline-hide-entry` to `zE`, you also get more consistency because `ze` is bound to `outline-show-entry`.

It fixes https://github.com/emacs-evil/evil-collection/issues/292


jojojames:
This looks good to me but I don't use these keys often so will defer to @Ambrevar

Ambrevar:
Looks good, thanks again!

PR #296, opened 2020-01-05
Add keybindings for toggling children in pdf-outline-mode

MaxSt:
I noticed that there where no keybindings defined to toggle the visibility of children in `pdf-outline-mode`. This PR adds this functionality.


Ambrevar:
Good catch, thanks!

PR #299, opened 2020-01-26
Magit-todos: unbind "j" in magit-todos-item-section-map

yunhao94:
"jT" in `magit-todos-item-section-map` prevents evil users from stepping into the section too, this PR fix it


jojojames:
Thanks!

PR #300, opened 2020-01-31
Replace `g]` with `g TAB`. Eliminate `g[`.

crocket:
I used `g[` because I thought `<backtab>` was impossible on linux virtual terminal. However, I just had to make a new keymap that inherits /usr/share/keymaps/i386/qwerty/us.map.gz to make emacs
recognize shift+tab as `<backtab>` on linux virtual terminal.

linux kernel community distributes kbd package which contains /usr/share/keymaps/i386/qwerty/us.map.gz which maps shift+tab to alt+tab. I don't see a reason to map shift+tab to alt+tab anywehre. I suspect somebody was trying to sabotage linux virtual
terminal.

Binding `g TAB` to the function that `<tab>` is bound to makes sense now because pressing `g TAB` and Shift+Tab is easy.

By default, Right Alt is mapped to AltGr, and Shift+Tab is mapped to Alt+Tab on linux virtual terminal. If you want to map Right Alt to Alt and Shift+Tab to Shift+Tab on linux virtual terminal, follow the steps below.

* Create /usr/share/keymaps/i386/qwerty/better-us.map.gz
```
include "us.map"
include "linux-with-two-alt-keys"

alt keycode 15 = Meta_Tab
shift keycode 15 = F26
string F26 = "\033[Z"
```
* Set keymap to `better-us` in the init script that sets linux console keymap during boot.
* Reboot
* After reboot, you will be able to press Right Alt and Shift+Tab in emacs on linux console.

### DIFF
--- a/modes/help/evil-collection-help.el
+++ b/modes/help/evil-collection-help.el
@@ -44,10 +44,8 @@
     (kbd "C-f") 'scroll-up-command
     (kbd "C-b") 'scroll-down-command
     (kbd "<tab>") 'forward-button
+    (kbd "g TAB") 'forward-button
     (kbd "<backtab>") 'backward-button
-    ;; This exists because <tab> is recognized as C-i on terminals.
-    "g]" 'forward-button
-    "g[" 'backward-button
 
     (kbd "C-o") 'help-go-back
     (kbd "C-i") 'help-go-forward
--- a/modes/info/evil-collection-info.el
+++ b/modes/info/evil-collection-info.el
@@ -43,10 +43,8 @@
     "l" 'evil-forward-char
     "h" 'evil-backward-char
     (kbd "<tab>") 'Info-next-reference
-    (kbd "S-<tab>") 'Info-prev-reference
-    ;; This exists because <tab> is recognized as C-i on terminals.
-    "g]" 'Info-next-reference
-    "g[" 'Info-prev-reference
+    (kbd "g TAB") 'Info-next-reference
+    (kbd "<backtab>") 'Info-prev-reference
 
     ;; From evil-integration.el.
     "0" 'evil-digit-argument-or-evil-beginning-of-line
--- a/readme.org
+++ b/readme.org
@@ -190,8 +190,8 @@ more.
          - Tab key is recognized as ~<tab>~ in GUI and ~TAB~ in terminal.
            ~TAB~ equals ~C-i~.
          - Since ~C-i~ is bound to jumping forward for vim compatibility,
-           bind ~g[~ and ~g]~ to the functions that Shift+Tab and Tab
-           are bound to on GUI for terminal compatibility.
+           bind ~g TAB~ to the function that ~<tab>~ is bound to for
+           terminal compatibility.
        - Enter key
          - Enter key is recognized as ~<return>~ in GUI and ~RET~ in terminal.
            ~RET~ equals ~Ctrl+m~.


jojojames:
(Keep in mind I haven't read the PR super closely.)

What about terminals from other systems like OSX / Windows?

crocket:
GUI emacs is available on Mac OSX and Windows.
On iOS and android, you are supposed to use ssh terminal which is not linux virtual console.
I haven't really tried a hardware keyboard on iOS or android.

jojojames:
> GUI emacs is available on Mac OSX and Windows.

Sure but we're talking about the case where someone is using terminal emacs.

crocket:
As far as I know, MacOS has great terminal emulators that should handle `g TAB` and Shift+Tab.
I don't know much about terminal emulators on windows because I don't use Mac or Windows.

jojojames:
Where I'm going with this, is this change seems really specific to your workflow (e.g. on linux virtual terminal) and at the same time seems like it would "ruin" other workflows. (e.g. osx -> unknown, windows -> unknown, bsd -> unknown)

crocket:
No. It is not specific to linux virtual terminal. As you know `g TAB` would work wherever emacs works.
Shift+Tab is backtab which is `Esc[Z` or `^[[Z`. This is ANSI reverse-tab. Various terminal environments adopted ANSI reverse-tab. [ncurses](https://invisible-island.net/ncurses/terminfo.src.html) enforces `\E[Z` across various terminal environments.

jojojames:
> No. It is not specific to linux virtual terminal. As you know g TAB would work wherever emacs works.

I think you're misunderstanding me, "linux virtual terminal" was just an example since that's what you mentioned.

Refer to the second half.

> and at the same time seems like it would "ruin" other workflows. (e.g. osx -> unknown, windows -> unknown, bsd -> unknown)

> Various terminal environments adopted ANSI reverse-tab. ncurses enforces \E[Z across various terminal environments.

Again, I'm asking about platforms other than Linux.

> I think Esc[Z is something that unix-like terminal environments all need to support.

**Need to** or already does? That's an important distinction, no?

crocket:
* I ran mintty on windows 10 in a virtual machine. I confirmed that mintty, which hosts msys, msys2, cygwin, and windows subsystem for linux, is compatible with xterm and maps Shift+Tab to backtab.
* iTerm2 which is the most popular terminal emulator on MacOS and recommends setting $TERM to xterm-256color. Because mintty is also compatible with xterm, I think all xterm-compatible terminal emulators should support backtab.
* Alacritty which is my primary terminal emulator on Gentoo Linux maps Shift+Tab to backtab. Alacritty is also available on MacOS.
* suckless terminal maps Shift+Tab to backtab
* xfce4-terminal maps Shift+Tab to backtab and sets $TERM to xterm-256color.
* Termius which is an android terminal emulator also sets $TERM to xterm-256color which should support backtab on hardware keyboards.

backtab should be possible on terminal emulators in graphical environments. Even if iTerm2 is an exception, there are alacritty and kitty on MacOS.

* Linux console doesn't support Right Alt and backtab by default.
* FreeBSD console doesn't support Right Alt and backtab by default.

I don't think it is a good idea to cater to linux console and BSD consoles which are stuck in the past. Those are places where you bootstrap graphical environments, not where you work and play. Even terminal junkies use terminal emulators in graphical environments.

jojojames:
> Even if iTerm2 is an exception, there are alacritty and kitty on MacOS.

The default terminal is Terminal.app on OSX. All three options here require the user to do "extra" things.

How about supporting both scenarios?

crocket:
https://en.wikipedia.org/wiki/Terminal_(macOS) says

> The preferences dialog for Terminal.app in OS X 10.8 (Mountain Lion) and later offers choices for values of the TERM environment variable. Available options are ansi, dtterm, nsterm, rxvt, vt52, vt100, vt102, xterm, xterm-16color and xterm-256color

It supports setting TERM to xterm-256color. Although I haven't tested, it should work.

jojojames:
If you C-h k in emacs in a terminal, what do you expect to see?

crocket:
`C-h k` and which key presses?

jojojames:
Shift-tab

crocket:
I expect to see `<backtab>` when I press Shift+Tab in terminal.

jojojames:
Terminal.app
```
M-[ runs the command (lambda (&optional arg) (interactive "P")
(sp-wrap-with-pair "[")) (found in global-map), which is an interactive Lisp
function.

It is bound to M-[, M-].

(anonymous &optional ARG)

Not documented.
```

Iterm2

```
M-[ runs the command (lambda (&optional arg) (interactive "P")
(sp-wrap-with-pair "[")) (found in global-map), which is an interactive Lisp
function.

It is bound to M-[, M-].

(anonymous &optional ARG)

Not documented.
```

TERM is xterm-256color.

GUI

```
<S-tab> runs the command org-shifttab, which is an interactive compiled Lisp
function in ‘org.el’.

It is bound to <S-tab>, <visual-state> <S-tab>, <normal-state> <S-tab>,
<backtab>, S-TAB, <S-iso-leftab>, <menu-bar> <Tbl> <Previous Field>, <menu-bar>
<Org> <Show/Hide> <Cycle Global Visibility>.
```

Maybe there's some additional configuration here but I'm not sure it's a good idea to bind something when the "out of the box" experience isn't set up for it.

crocket:
> it's a good idea to bind something

You mean "unbind something".

jojojames:
> You mean "unbind something".

Yup, I think it's reasonable to keep g]/g[ and add the extra <backtab> you're wanting.

crocket:
I will come up with a new pull request.

PR #304, opened 2020-02-14
Add bindings for Dashboard

kisaragi-hiu:
Added bindings to [Dashboard](https://github.com/emacs-dashboard/emacs-dashboard)'s own commands, like next/previous section, refreshing the buffer, and jumping to sections.

I had to use an advice as dashboard.el defines some commands (the ones that jump to sections) as unnamed functions bound to a local variable, *after* the other parts of the map has been defined.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -128,6 +128,7 @@ through removing their entry from `evil-collection-mode-list'."
     compile
     (custom cus-edit)
     cus-theme
+    dashboard
     daemons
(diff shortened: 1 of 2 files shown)


Ambrevar:
Looking good!  I don't use dashboard so maybe @jojojames has a better opinion than me for these bindings.
Otherwise feel free to merge!

jojojames:
Thanks!

PR #305, opened 2020-02-15
Add `g TAB` as a substitute on terminal for Tab.

crocket:
`g TAB` feels intuitive.

### DIFF
--- a/readme.org
+++ b/readme.org
@@ -189,9 +189,11 @@ more.
        - Tab key
          - Tab key is recognized as ~<tab>~ in GUI and ~TAB~ in terminal.
            ~TAB~ equals ~C-i~.
-         - Since ~C-i~ is bound to jumping forward for vim compatibility,
-           bind ~g[~ and ~g]~ to the functions that Shift+Tab and Tab
-           are bound to on GUI for terminal compatibility.
+         - ~C-i~ is bound to jumping forward for vim compatibility.
+           If Shift+Tab is not relevant, just bind ~g TAB~ to the function
+           that Tab is bound to. If Shift+Tab is relevant, bind ~g]~ and
+           ~g TAB~ to the function that Tab is bound to, and bind ~g[~ to
+           the function that Shift+Tab is bound to for terminal compatibility.
        - Enter key
          - Enter key is recognized as ~<return>~ in GUI and ~RET~ in terminal.
            ~RET~ equals ~Ctrl+m~.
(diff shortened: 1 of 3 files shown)


jojojames:
Looks reasonable. @Ambrevar ?

readme.org
```diff
@@ -189,9 +189,12 @@ more.
...
          - Tab key is recognized as ~<tab>~ in GUI and ~TAB~ in terminal.
            ~TAB~ equals ~C-i~.
-         - Since ~C-i~ is bound to jumping forward for vim compatibility,
-           bind ~g[~ and ~g]~ to the functions that Shift+Tab and Tab
-           are bound to on GUI for terminal compatibility.
+         - ~C-i~ is bound to jumping forward for vim compatibility.
+           If Shift+Tab is not relevant, just bind ~g TAB~ to the function
+           that Tab is bound to. If Shift+Tab is relevant,
+           bind ~g]~ and ~g TAB~ to the function that Tab is bound to,
+           and bind ~g[~ or ~g}~ to the function that Shift+Tab is bound to
```
Ambrevar: Why `g}`?

readme.org
```diff
@@ -189,9 +189,12 @@ more.
        - Tab key
          - Tab key is recognized as ~<tab>~ in GUI and ~TAB~ in terminal.
            ~TAB~ equals ~C-i~.
-         - Since ~C-i~ is bound to jumping forward for vim compatibility,
-           bind ~g[~ and ~g]~ to the functions that Shift+Tab and Tab
-           are bound to on GUI for terminal compatibility.
+         - ~C-i~ is bound to jumping forward for vim compatibility.
```
Ambrevar: vim -> Vim.

Ambrevar:
Looks good to me beside the comments in the README.  Thanks!

readme.org
crocket: Because Shift+Tab may not be functionally related to Tab in some modes such as org mode.

Ambrevar:
But why not just `g[` then?

crocket:
Because `g[` feels like something that does something opposite to what `g]` does.

Ambrevar:
Isn't Shift-Tab the opposite of Tab?

crocket:
Not always. In org-mode, Shift-Tab is not functionally opposite to Tab.

Ambrevar:
So what are you suggesting?  That we use `g}` in Org mode for `shift-Tab`?

`g[` would be OK to me here too.  I find it simpler to have just one key instead of two.

crocket:
Ok. I'm going to omit `g}` since `g}` is hard to press anyway.

jojojames:
@crocket Sorry if I missed this, is this ready?

crocket:
I think it is ready.

jojojames:
Sorry @crocket , merging.

PR #306, opened 2020-02-16
fix byte-compile errors

yunhao94:
Recently introduced evil-collection-ebuku.el and evil-collection-dashboard.el require third-party packages directly, which causes byte-compile error when install from MELPA.

----

#


jojojames:
Nice, thanks!

Issue #184, opened 2018-07-22
evil-mc


jojojames:
evil-mc has this keymap:

```
(defvar evil-mc-key-map
  (let ((map (make-sparse-keymap))
        (keys '(("grm" . evil-mc-make-all-cursors)
                ("gru" . evil-mc-undo-all-cursors)
                ("grs" . evil-mc-pause-cursors)
                ("grr" . evil-mc-resume-cursors)
                ("grf" . evil-mc-make-and-goto-first-cursor)
                ("grl" . evil-mc-make-and-goto-last-cursor)
                ("grh" . evil-mc-make-cursor-here)
                ("grj" . evil-mc-make-cursor-move-next-line)
                ("grk" . evil-mc-make-cursor-move-prev-line)
                ("M-n" . evil-mc-make-and-goto-next-cursor)
                ("grN" . evil-mc-skip-and-goto-next-cursor)
                ("M-p" . evil-mc-make-and-goto-prev-cursor)
                ("grP" . evil-mc-skip-and-goto-prev-cursor)
                ("C-n" . evil-mc-make-and-goto-next-match)
                ("grn" . evil-mc-skip-and-goto-next-match)
                ("C-t" . evil-mc-skip-and-goto-next-match)
                ("C-p" . evil-mc-make-and-goto-prev-match)
                ("grp" . evil-mc-skip-and-goto-prev-match))))
    (dolist (key-data keys)
      (evil-define-key 'normal map (kbd (car key-data)) (cdr key-data))
      (evil-define-key 'visual map (kbd (car key-data)) (cdr key-data)))
    map))
```

Which takes over the g/gr prefix. I think anyone using evil-collection and evil-mc is going to find some weird binding conflicts between the two.

I've had this HACK for a long time for evil-mc to remove the keymap completely and install my own.

```
  (defun +remove-evil-mc-keymap (&rest _args)
    "Brute force remove `evil-mc-key-map' from `evil-mode-map-alist'."
    (when-let ((evil-mc-map (assq 'evil-mc-mode evil-mode-map-alist)))
      (setq evil-mode-map-alist
            (delq evil-mc-map evil-mode-map-alist))))

  (advice-add 'evil-normalize-keymaps :after '+remove-evil-mc-keymap)
```

Anyone have a better idea on how to do this so we can make sure installing both packages is still a good experience?

Ambrevar:
[...]

PR #310, opened 2020-02-28
Bind evil-mc-cursors-map to `gC`

Silex:
Hello,

This follows https://github.com/gabesoft/evil-mc/pull/109, don't merge it yet there are things to sort out before.

For starter, I think https://github.com/emacs-evil/evil-collection/blob/master/modes/evil-mc/evil-collection-evil-mc.el#L34 is weird because there's no `evil-mc-map`. As you see below the keymap is named `evil-mc-key-map`. Also thus there's no `evil-mc-mode-map`?

``` emacs-lisp
(define-minor-mode evil-mc-mode
  "Toggle evil multiple cursors in a single buffer."
  :group 'evil-mc
  :init-value nil
  :keymap evil-mc-key-map
```

1. Should I rename both `evil-mc-map` and `evil-mc-mode-map` to `evil-mc-key-map`? What's the purpose of `evil-collection-evil-mc-maps`?
2. Should I add a defvar for `evil-mc-cursors-map`?
3. What do you think of `gC`? Something else comes to mind?

### DIFF
--- a/modes/evil-mc/evil-collection-evil-mc.el
+++ b/modes/evil-mc/evil-collection-evil-mc.el
@@ -31,15 +31,13 @@
 (require 'evil-mc nil t)
 (eval-when-compile (require 'subr-x)) ; `if-let*' and `when-let*'
 
-(defvar evil-mc-map)
-(defconst evil-collection-evil-mc-maps '(evil-mc-mode-map))
+(defvar evil-mc-key-map)
+(defconst evil-collection-evil-mc-maps '(evil-mc-key-map))
 
 (defun evil-collection-evil-mc-clear-keymap (&rest _args)
   "Brute force remove `evil-mc-key-map' from `evil-mode-map-alist'."
-  (evil-collection-when-let*
-      ((evil-mc-map (assq 'evil-mc-mode evil-mode-map-alist)))
-    (setq evil-mode-map-alist
-          (delq evil-mc-map evil-mode-map-alist))))
+  (evil-collection-when-let* ((map (assq 'evil-mc-mode evil-mode-map-alist)))
+    (setq evil-mode-map-alist (delq map evil-mode-map-alist))))
 
 ;;;###autoload
 (defun evil-collection-evil-mc-setup ()
@@ -51,6 +49,9 @@
   ;; details.
   (advice-add 'evil-normalize-keymaps
               :after 'evil-collection-evil-mc-clear-keymap)
+
+  (setq evil-mc-key-map (make-sparse-keymap))
+  (evil-define-key* '(normal visual) evil-mc-key-map (kbd "gz") evil-mc-cursors-map)
 
   ;; https://github.com/gabesoft/evil-mc/issues/70
   (add-hook 'evil-mc-after-cursors-deleted
--- a/modes/evil-mc/evil-collection-evil-mc.el
+++ b/modes/evil-mc/evil-collection-evil-mc.el
@@ -34,10 +34,13 @@
 (defvar evil-mc-key-map)
 (defconst evil-collection-evil-mc-maps '(evil-mc-key-map))
 
-(defun evil-collection-evil-mc-clear-keymap (&rest _args)
+(defun evil-collection-evil-mc-change-keymap (&rest _args)
   "Brute force remove `evil-mc-key-map' from `evil-mode-map-alist'."
   (evil-collection-when-let* ((map (assq 'evil-mc-mode evil-mode-map-alist)))
-    (setq evil-mode-map-alist (delq map evil-mode-map-alist))))
+    (setq evil-mode-map-alist (delq map evil-mode-map-alist))
+    (let ((map (make-sparse-keymap)))
+      (evil-define-key* '(normal visual) map (kbd "gz") evil-mc-cursors-map)
+      (add-to-list 'evil-mode-map-alist '(evil-mc-mode . map)))))
 
 ;;;###autoload
 (defun evil-collection-evil-mc-setup ()
@@ -48,10 +51,7 @@
   ;; See https://github.com/emacs-evil/evil-collection/issues/184 for more
   ;; details.
   (advice-add 'evil-normalize-keymaps
-              :after 'evil-collection-evil-mc-clear-keymap)
-
-  (setq evil-mc-key-map (make-sparse-keymap))
-  (evil-define-key* '(normal visual) evil-mc-key-map (kbd "gz") evil-mc-cursors-map)
+              :after 'evil-collection-evil-mc-change-keymap)
 
   ;; https://github.com/gabesoft/evil-mc/issues/70
   (add-hook 'evil-mc-after-cursors-deleted


Silex:
Previous history at #184

Ambrevar:
I don't use multi cursors, so I'm not the best judge here.

The evil-collection-evil-mc-maps is passed to the mode hook on init, see `evil-collection-init`.

So yes, I think the keymap names were a mistake and renaming to evil-mc-key-map makes sense.

"gC" seems good, althgough a bit hard to type maybe.

You only need to add a defvar if you are going to use the value of the symbol within the file.  In doubt, add the defvar.

Silex:
@Ambrevar: Thanks for the clarifications. About `gC` I agree it's not the best, I'll come up with something better.

I'll ping you when it's ready.

DanielRichtmann:
@Silex: Concerning the keybinding: How about `gz` like in doom?
https://github.com/hlissner/doom-emacs/blob/develop/modules/config/default/+evil-bindings.el

Silex:
@DanielRichtmann: interesting idea. In evil-magit `gz` is bound to `magit-jump-to-stashes`, but I guess we can live with that.

I was wondering about a binding more quick to use, but only `C-d` seems to be free & reasonable (it's currently bound to `delete-char` which almost no evil user use.

DanielRichtmann:
I agree on the former.
Personally the latter isn't too much of a concern for me since I mostly just use the C-n/C-p/C-t bindings.

Silex:
> Personally the latter isn't too much of a concern for me since I mostly just use the `C-n`/`C-p`/`C-t` bindings.

FYI those bindings won't be bound with this PR, only the `gr*` ones.

jojojames:
@Silex Was there an AI on our part or are you still noodling on the specific bindings?

> is weird because there's no evil-mc-map

You are probably right.

Silex:
@jojojames: sorry, I forgot about this PR. I have trouble finishing the implementation because I'm lost in keymaps land.

I read https://github.com/syl20bnr/spacemacs/wiki/Keymaps-guide and https://github.com/noctuid/evil-guide. I think I understand that there is `evil-mode-map-alist` that is used by evil to enable/disable keymaps according to the modes.

Then each mode defines normal/insert/visual keymaps and evil use these along with the state to activate/disable keymaps.

Then evil-collection comes along and uses `with-eval-after-load` for each mode, which calls `evil-collection-*-setup`, which usually calls `evil-collection-define-key` overwriting the mode's keymap.

1. Correct me if I'm wrong in this introduction.
1. Why can't I edebug `evil-collection-*-setup`? Is it actually only called once?
1. In evil-mc's case, I see that you simply remove evil-mc's keymap from evil (https://github.com/emacs-evil/evil-collection/blob/master/modes/evil-mc/evil-collection-evil-mc.el#L42). I tried then simply adding the new one (with `gz`), but that fails (see the TMP commit). Maybe this is due to my inability to properly debug/test my changes tho.
1. Any advices welcome. I want thinking, what about if we simply loop over `evil-mc-key-map` and unbind everything there? Then simply add the `gz` binding. Isn't that less tricky?

p.s: what does `AI` means in "Was there an AI on our part" ?

AndreaOrru:
What's the state of this? Currently I'm not able to use `evil-mc` and `evil-collection` together.

Silex:
@AndreaOrru: it's still planned, but I have more urgent things to do at the moment.

The questions from my last message are still open, I'm lost between all these keymaps and precedance and what not. Basically everything is _almost ready_ we just need to make it work, that is bind `evil-mc-cursors-map` to `gz`.

jojojames:
Sorry, didn't notice you left your last comment with questions. I thought you were perpetually busy so didn't re-read this thread until now.

> Correct me if I'm wrong in this introduction.

Corrrect

> Why can't I edebug evil-collection-*-setup? Is it actually only called once? (I C-u C-M-x on it, then M-x some-mode twice but it does not break. If I M-x eval-expression (evil-collection-*-setup) then it breaks.

Err, in this case, I recommend looking at the parts that make that file and interactively debugging/developing it. It only runs once with with-eval-after-load.

> In evil-mc's case, I see that you simply remove evil-mc's keymap from evil (https://github.com/emacs-evil/evil-collection/blob/master/modes/evil-mc/evil-collection-evil-mc.el#L42).

Yup.


> I tried then simply adding the new one (with gz), but that fails (see the TMP commit). Maybe this is due to my inability to properly debug/test my changes tho.

This mode is just tricky I think. The keybinding in it is not the easiest to work with, and evil bindings are a little harder to reverse. 


> Any advices welcome. I was thinking, what about if we simply loop over evil-mc-key-map and unbind everything there? Then simply add the gz binding. Isn't that less tricky?

I tried the below out real quickly and it seems to work for me. Using combinations of evil-define-key/collection-define-key, etc doesn't seem to work as the binding seems to be *sticky* on the original binding.

If we use define-key and manually switch out the keymaps, it should work.. but we run into the problem that evil-collection-define-key does more than just bind keys, it also checks white/blacklist and records the keybinding. So I added that below. I'm not sure I'm happy about moving these bits of logics into the modes though. I'd prefer it stay in evil-collection.el.

You might be able to play with this further to get you where you want. Ideally a solution that uses just evil-collection-define-key would be ideal but evil-mc might also be a special case...

```
(defun evil-collection-can-bind-key-p (key)
  "Check if KEY can be bound in `evil-collection'."
  (let ((whitelist (mapcar 'kbd evil-collection-key-whitelist))
        (blacklist (mapcar 'kbd evil-collection-key-blacklist)))
    (or (and whitelist (member key whitelist))
        (not (member key blacklist)))))

(defun evil-collection-evil-mc-change-keymap (&rest _args)
  ""
  (evil-collection-when-let*
      ((evil-mc-map (assq 'evil-mc-mode evil-mode-map-alist)))

      (when (evil-collection-can-bind-key-p "gz")
        (let ((map-sym 'evil-mc-key-map)
              (key "gz")
              (def evil-mc-cursors-map))
          (annalist-record 'evil-collection 'keybindings
                           (list map-sym 'normal key def)
                           :local (or (eq map-sym 'local)
                                      (local-variable-p map-sym)))
          (define-key map key def)))
      (setq evil-mc-key-map map)
      (setf (cdr (assq 'evil-mc-mode minor-mode-map-alist)) map)
      (setf (cdr (assq 'evil-mc-mode evil-mode-map-alist)) map))))

  (advice-add 'evil-normalize-keymaps
              :after 'evil-collection-evil-mc-change-keymap)


```


> p.s: what does AI means in "Was there an AI on our part" ?

Action Item

Silex:
> Sorry, didn't notice you left your last comment with questions. I thought you were perpetually busy so didn't re-read this thread until now.

Unfortunately I think you are right. I just don't have enough time to maintain my Emacs activities 😢 

I'll see if I can find some time to implement your ideas, but I'm thinking of finding maintainers for my packages.

jojojames:
Sounds good, will close this for now. :D

Silex:
I'm somewhat back to Emacs :partying_face: 

I'll try to reopen this soonish.

Silex:
@jojojames can you reopen this? I found something very simple and it works!

jojojames:
It's not allowing me to due to some 'master branch was force pushed' message. Better to create a new PR I guess.

Silex:
Continuing in #565

PR #315, opened 2020-03-26
Updated gnus bindings

MorganJamesSmith:
Added bindings for the remaining 3 gnus modes (group, server, and
browse) and also reworked some bindings that seemed unintuitive.

These aren't perfect, but it's a start. When I started using gnus it was jarring for only some of the modes to be supported. This fixes that problem.

It compiles using cask, and the linter didn't throw any new warnings

### DIFF
--- a/modes/gnus/evil-collection-gnus.el
+++ b/modes/gnus/evil-collection-gnus.el
@@ -233,8 +233,8 @@
     "a"         'gnus-server-add-server
     "y"         'gnus-server-copy-server
     "e"         'gnus-server-edit-server
-    "g"         'gnus-server-regenerate-server
-    "h"         'describe-mode
+    "gr"        'gnus-server-regenerate-server
+    "g?"        'describe-mode
     "i"         'gnus-server-toggle-cloud-server
     "d"         'gnus-server-kill-server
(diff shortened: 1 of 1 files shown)


modes/gnus/evil-collection-gnus.el
```diff
@@ -192,7 +185,75 @@
...
+    "I"         'gnus-server-set-cloud-method-server
+    "L"         'gnus-server-offline-server
+    "O"         'gnus-server-open-server
+    "R"         'gnus-server-remove-denials
+    "S"         'gnus-server-show-server
+    "a"         'gnus-server-add-server
+    "y"         'gnus-server-copy-server
+    "e"         'gnus-server-edit-server
+    "g"         'gnus-server-regenerate-server
+    "h"         'describe-mode
```
Ambrevar: Avoid binding to `hjkl` it can be frustrating.  Prefer `g?` for this one.

MorganJamesSmith: My bad! That one slipped through copy pasted from the original bindings. It's one of the lesser used modes so I wasn't bugged by it yet. Thanks for catching that!

Ambrevar:
Thanks!

Issue #320, opened 2020-04-19
evil-collection-unimpaired-insert-space-above  works not as expected


anuvyklack:
Description :octocat:
The `evil-collection-unimpaired-insert-space-above` command seems to work not like expect when the line is empty, or when the cursor is at the first column.

Reproduction guide 🐞
Open a file with enabled Evil mode and navigate to a line with the text.
Execute the `evil-collection-unimpaired-insert-space-above` command (`[ SPC`).

Observed behavior: 👀 💔
When the cursor is in the middle of a line that has text, the cursor stays in its position in the text and the current line shifts downward, relatively, as blank lines appear above the current line. This is what we want.

But when the cursor is in a blank line, it seems to stay in place, but blank lines appear beneath the current line.

And when the cursor is in a line containing text and positioned at the first column, the new line appears above but the cursor moves into it.

Expected behavior: ❤️ 😄
The `evil-collection-unimpaired-insert-space-above` should always insert lines above the cursor, regardless of whether the current line has any text content.

I spent some time trying to figuring out why, and it seems that this behavior is due to the [save-excursion](https://github.com/emacs-evil/evil-collection/blob/493d523c9bcdeb2213f18cbf73c9e33de854dd34/modes/unimpaired/evil-collection-unimpaired.el#L81) function: it does not restore the point position (or do it wrong) if the cursor is at the beginning of the line.

anuvyklack:
I am learning lisp third day, and sure there is a better solution but this function seems to solve the problem:

```lisp
(defun evil-collection-unimpaired-insert-space-above (count)
  "Insert space above current line"
  (interactive "p")
  (dotimes (_ count)
    ;; (defvar-local position (point))
    (setq position (point))
    (save-excursion (evil-insert-newline-above))
    (goto-char (+ position 1))
    (makunbound 'position)
  )
)
```

If use `defvar-local` instead of `setq` it doesn't work. I don't know why.

jojojames:
If it's save-excursion that may be the problem, maybe save-mark-and-excursion?

anuvyklack:
No `save-mark-and-excursion` also doesn't work.

anuvyklack:
This is better soluion:

```lisp
(defun evil-collection-unimpaired-insert-space-above (count)
  "Insert space above current line"
  (interactive "p")
  (dotimes (_ count)
    (save-excursion (evil-insert-newline-above))
    (if (bolp) (forward-char 1))
  )
)
```

PR #321, opened 2020-04-20
Fix bug in unimpaired module.

anuvyklack:
Related to #320.
Also, took `save-excursion` out of the loop, to make it executed only once instead of calling it on each iteration.

### DIFF
--- a/modes/unimpaired/evil-collection-unimpaired.el
+++ b/modes/unimpaired/evil-collection-unimpaired.el
@@ -78,12 +78,13 @@
 (defun evil-collection-unimpaired-insert-space-above (count)
   "Insert space above current line"
   (interactive "p")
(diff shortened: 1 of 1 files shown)


modes/unimpaired/evil-collection-unimpaired.el
```diff
@@ -78,12 +78,13 @@
 (defun evil-collection-unimpaired-insert-space-above (count)
   "Insert space above current line"
   (interactive "p")
-  (dotimes (_ count) (save-excursion (evil-insert-newline-above))))
+  (save-excursion (dotimes (_ count) (evil-insert-newline-above)))
+  (if (bolp) (forward-char count)))
```
Ambrevar, line 82: Prefer `when` to `if` when there is no "else" form.  Otherwise looks good to me!

Ambrevar:
Thanks!

PR #326, opened 2020-05-05
Add imenu jumps to the Evil jump list

fishyfriend:
Adding imenu navigation to the jump list is natural, since many or most other navigation actions already add to the jump list by default. (Search for `:jump` in [evil-commands.el](https://github.com/emacs-evil/evil/blob/master/evil-commands.el.) to see.)

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -175,6 +175,7 @@ through removing their entry from `evil-collection-mode-list'."
     image
     image-dired
     image+
+    imenu
     imenu-list
     indium
     info
--- a/modes/imenu/evil-collection-imenu.el
+++ b/modes/imenu/evil-collection-imenu.el
@@ -1,0 +1,38 @@
+;;; evil-collection-imenu.el --- Evil integration for `imenu' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Jacob First
+
+;; Author: Jacob First <jacob.first@member.fsf.org>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; URL: https://github.com/emacs-evil/evil-collection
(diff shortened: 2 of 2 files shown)


Ambrevar:
I'm not familiar with "imenu jumps" or Evil jump list.  What do they do?

fishyfriend:
By "imenu jump" I simply mean the movement that occurs when a user navigates to a definition using `imenu`. So, for example, let's say I'm editing an Elisp file that imenu has indexed, and I have positioned point somewhere inside `(defun foo () ...)`. Now I run `M-x imenu` and select `bar` from the list of index items. This causes point to move to `(defun bar ...)`. This movement of point would constitute one "jump."

(Imenu is an Emacs built-in package that indexes definitions in a file and allows navigating to them via the `imenu` command. Here's the [manual page](https://www.gnu.org/software/emacs/manual/html_node/emacs/Imenu.html) for slightly more.)

The jump list is a Vi feature implemented by Evil. It's a list of locations where point has been previously. The default bindings `C-o` and `C-i` allow to navigate backwards and forwards through the list.  The list is updated only after certain commands like searching (`/`), visiting next match (`n`), or navigating to a marker (<code>`</code>), so it is useful for navigating quickly through the history of locations you have visited in a file.  It is somewhat like Emacs' built-in [mark ring](https://www.gnu.org/software/emacs/manual/html_node/emacs/Mark-Ring.html), except that the locations are added to the list automatically.

I have added a rationale to the PR description so it's clearer.

Ambrevar:
Thanks for the details and nice patch!

fishyfriend:
Cheers - thanks for the review!

PR #329, opened 2020-05-08
Changed mu4e d key-bind

gspe:
Changed mu4e d key-bind in mu4e-headers-mode-map form
mu4e-headers-mark-for-trash to mu4e-headers-mark-or-move-to-trash.
Changed mu4e d key-bind in mu4e-view-mode-map form mu4e-view-mark-for-trash to mu4e-view-mark-or-move-to-trash.


Ambrevar:
Thanks!

PR #338, opened 2020-05-23
Add delete and prev/next prompt bindings for vterm

ajgrf:
This PR fixes the delete key for vterm, and adds bindings to jump to the previous or next prompt.


Ambrevar:
Thanks!

PR #339, opened 2020-05-26
Add bindings for auto-package-update

condy0919:
@Ambrevar

### DIFF
--- a/modes/auto-package-update/evil-collection-auto-package-update.el
+++ b/modes/auto-package-update/evil-collection-auto-package-update.el
@@ -1,0 +1,44 @@
+;;; evil-collection-auto-package-update.el --- Evil bindings for auto-package-update -*- lexical-binding: t -*-
+
+;; Copyright (C) 2020 Zhiwei Chen
+
+;; Author: Zhiwei Chen <condy0919@gmail.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <mail@ambrevar.xyz>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, auto-package-update, tools
+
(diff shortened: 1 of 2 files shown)


Ambrevar:
Is there no binding other than quit?

condy0919:
Yes, only a `quit-window`. `auto-package-update.el` provides automatic package update without other user operations.

Ambrevar:
Is it still worth a new file then?
@jojojames what do you think?

condy0919:
Binding `q` to `quit-window` makes the behavior consistent.
If there is no such binding, user must use `evil-delete-window` (<kbd>C-w c</kbd>) or `delete-window` (<kbd>C-x 0</kbd>) which are inconvenient than pressing `q`.

modes/auto-package-update/evil-collection-auto-package-update.el
```diff
@@ -0,0 +1,41 @@
...
+;; Evil bindings for auto-package-update.
+
+;;; Code:
+(require 'evil-collection)
+(require 'auto-package-update nil t)
+
+;;;###autoload
+(defun evil-collection-auto-package-update-setup ()
+  "Set up `evil' bindings for `auto-package-update'."
+  (evil-define-minor-mode-key 'normal 'auto-package-update-minor-mode
```
jojojames, line 36: Can you route through to evil-collection-define-key instead?

I think you'll also want a defvar of the keymap.

See https://github.com/emacs-evil/evil-collection/blob/master/modes/hg-histedit/evil-collection-hg-histedit.el

jojojames:
@Ambrevar Yeah, I think it's reasonable to add.

modes/auto-package-update/evil-collection-auto-package-update.el
condy0919: Finally done after found https://github.com/emacs-evil/evil/issues/301

condy0919:
@Ambrevar

Ambrevar:
Thanks!

Issue #345, opened 2020-06-04
Slightly annoying insert-mode activation when using org-mu4e-compose-mode


dakling:
`evil-collection` currently sets the initial evil state for `mu4e-compose-mode` to `'insert`, which generally of course makes a lot of sense. However, there is an edge case where this leads to an annoying behaviour, namely when using `org-mu4e`. `org-mu4e` allows writing html mails from emacs, and it works by enabling `org-mode` when in the message body, and `mu4e-compose-mode` when in the message header. With the current configuration, this obviously puts one into insert mode when moving from the message body to the message header, which is quite annoying.

I have a couple of ideas how to fix this:
1. set the initial evil state for `mu4e-compose-mode` to `'normal`, but add a `mu4e-compose-pre-hook` that enters insert state (feels like a hack, but would be the simplest solution)
2. perhaps a bit cleaner, but requiring more code:
```
(defun evil-collection-mu4e-org-set-header-to-normal-mode ()
  (evil-set-initial-state 'mu4e-compose-mode 'normal))

(defun evil-collection-mu4e-org-set-header-to-insert-mode ()
  (evil-set-initial-state 'mu4e-compose-mode 'insert))

    (add-hook 'org-mode-hook
              #'evil-collection-mu4e-org-set-header-to-normal-mode)
    (add-hook 'mu4e-compose-pre-hook
              #'evil-collection-mu4e-org-set-header-to-insert-mode)
```

Let me know what you think - if you like the general idea, I'll create a PR. Thanks!

Ambrevar:
I don't use mu4e but if the above works for you, it makes sense to me.
Please send a PR :)

PR #346, opened 2020-06-05
fix insert state activation issue in org-mu4e

dakling:
see  #345


Ambrevar:
The hooks should be run in the setup function.  See other uses of `add-hook` in other Evil Collection bindings.

Ambrevar:
Thanks!

PR #350, opened 2020-06-13
TAB can move the point to the next link

condy0919:
The bindings are stolen from `evil-collection-man.el`

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -8,7 +8,7 @@
 ;; Pierre Neidhardt <mail@ambrevar.xyz>
 ;; URL: https://github.com/emacs-evil/evil-collection
 ;; Version: 0.0.2
-;; Package-Requires: ((emacs "25.1") (cl-lib "0.5") (evil "1.2.13") (annalist "1.0"))
+;; Package-Requires: ((emacs "25.1") (evil "1.2.13") (annalist "1.0"))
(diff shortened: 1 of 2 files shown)


Ambrevar:
This already works for me when the point is on links.
Maybe your local configuration is overriding these bindings?

condy0919:
> This already works for me when the point is on links.
> Maybe your local configuration is overriding these bindings?

TAB on links runs `shr-next-link`. `forward-button` is more general, it can move the point to the next link even if the point is at plain text.

Ambrevar:
Are you sure?  When I apply your patch and press "TAB", it echoes "No
buttons!"

But I noticed that in Emacs state, TAB is bound to
elfeed-show-next-link, which does what you want.

condy0919:
Bind TAB to `elfeed-show-next-link` now.

Ambrevar:
Can you squash the commits?  Thanks!

condy0919:
Squashed!

Ambrevar:
Sorry for being picky here :p  Could you prefix your commit message with the package you are targeting?
Here, it would be

```
elfeed: Bind TAB to move the point to the next link
```

Thank you for your patience :)

condy0919:
Fixed

Ambrevar:
Thanks a lot!

PR #357, opened 2020-07-09
Bind eshell-kill-input in Evil insert state

leungbk:
----

#

### DIFF
--- a/modes/eshell/evil-collection-eshell.el
+++ b/modes/eshell/evil-collection-eshell.el
@@ -116,6 +116,10 @@
     ;; motion
     (kbd "M-h") 'eshell-backward-argument
     (kbd "M-l") 'eshell-forward-argument)
+  ;; TODO: What if the user changes `evil-want-C-u-delete' after this is run?
+  (when evil-want-C-u-delete
+    (evil-collection-define-key 'insert 'eshell-mode-map
+      (kbd "C-u") 'eshell-kill-input))
   (evil-collection-define-key 'visual 'eshell-mode-map
     ;; motion
     ;; TODO: This does not work with `evil-visual-line'.
--- a/modes/eshell/evil-collection-eshell.el
+++ b/modes/eshell/evil-collection-eshell.el
@@ -79,7 +79,8 @@
                  type register yank-handler)))
 
 ;; Taken from Doom Emacs
-(evil-define-operator evil-collection-eshell-evil-delete-line (_beg end type register yank-handler)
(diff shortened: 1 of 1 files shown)


Ambrevar:
Looks good to me, but I'm not familiar with these bindings.  @jojojames ?

leungbk:
In terminals, C-u is commonly understood to delete text backwards until
reaching the beginning of the shell prompt. I assume that this influenced
Vi's choice of bindings here (or vice-versa). Vanilla Emacs also has C-c
C-u in Eshell as a default binding to backward-kill the prompt.

leungbk:
Should we go ahead and change https://github.com/emacs-evil/evil-collection/blob/c68aed6a9694836602ada224160bc5b4c0a37919/modes/eshell/evil-collection-eshell.el#L82? Even if we don't really use `_beg`, the `evil-define-operator` macro expands to something that has to access `_beg` for bookkeeping purposes.

Ambrevar:
Brian Leung <notifications@github.com> writes:

> In terminals, C-u is commonly understood to delete text backwards until
> reaching the beginning of the shell prompt. I assume that this influenced
> Vi's choice of bindings here (or vice-versa). Vanilla Emacs also has C-c
> C-u in Eshell as a default binding to backward-kill the prompt.

Makes sense, thanks for the explanation.

Ambrevar:
> Should we go ahead and change https://github.com/emacs-evil/evil-collection/blob/c68aed6a9694836602ada224160bc5b4c0a37919/modes/eshell/evil-collection-eshell.el#L82? Even if we don't really use `_beg`, the `evil-define-operator` macro expands to something that has to access `_beg` for bookkeeping purposes.

To clarify, this is not related to your patch, or is it?

leungbk:
> Should we go ahead and change https://github.com/emacs-evil/evil-collection/blob/c68aed6a9694836602ada224160bc5b4c0a37919/modes/eshell/evil-collection-eshell.el#L82? Even if we don't really use `_beg`, the `evil-define-operator` macro expands to something that has to access `_beg` for bookkeeping purposes.
> To clarify, this is not related to your patch, or is it?

It's not, it seems to have been introduced when the `doom`-inspired `eshell` functions were. The CI complaint is that `evil-define-operator` does use `_beg`, even though it winds up playing no role in any meaningful computations.

jojojames:
> It's not, it seems to have been introduced when the doom-inspired eshell functions were. The CI complaint is that evil-define-operator does use _beg, even though it winds up playing no role in any meaningful computations.

Suggestion?

Otherwise, seems fine to merge.

leungbk:
@jojojames I fixed the CI failure with a follow-up commit, so everything should be OK.

Issue #23, opened 2017-11-09
image-mode and pdf-tools conflict


Ambrevar:
Depending on the load order, one mode takes precedence over the other which makes the latter unusable.

This might have to do with an inheritance problem.
Bug was reported upstream: https://github.com/politza/pdf-tools/issues/324

jojojames:
I don't use pdf-tools so won't be able to look at it directly.

I'm curious what the minor map alist looks like before/after you normalize keymaps?

(evil-normalize-keymaps)

evil-mode-map-alist <----------

Ambrevar:
The before/after keymaps are the same.
Let me know if you want a dump.

A first glance at the problem reveals that
- doc-view and image-mode suffer from the same conflict.
- image-mode-map contains quite a few "remap".  This could be the very root of the issue.

Ambrevar:
By the way since doc-view is also concerned, you should be able to reproduce :)

Ambrevar:
Some first experiments:

- The conflict happens in any state, Emacs state included.

- Emacs without evil is fine.

- I don't think it has anything to do with `remap` as `j` is not a remapped
binding.  I am not sure though.

- If doc-view / pdf-view maps are set first, then everything is OK.

- If image-mode map is set first, then the last binding will override the rest.


I tried using that result over the loading process of evil-collection to get
doc-view and image-mode to work but it did not.  There must be something else.


Initial setup:

```
(require 'evil)
(evil-mode 1)
```

Some tests:

```
;;; Issue: doc-view overrides image-mode.

(evil-set-initial-state 'image-mode 'motion)
(evil-define-key 'motion image-mode-map "j" 'message-news)
(find-file "bar.png")

(evil-set-initial-state 'doc-view-mode 'motion)
(evil-define-key 'motion doc-view-mode-map "j" 'emacs-version)
(find-file "foo.odt")
```

```
;;; Change order.
;;; Issue: None.

(evil-set-initial-state 'doc-view-mode 'motion)
(evil-define-key 'motion doc-view-mode-map "j" 'emacs-version)
(find-file "foo.odt")

(evil-set-initial-state 'image-mode 'motion)
(evil-define-key 'motion image-mode-map "j" 'message-news)
(find-file "bar.png")
```

```
;;; Emacs state suffers from the same conflict.
;;; Issue: doc-view overrides image-mode.

(evil-set-initial-state 'image-mode 'emacs)
(evil-define-key 'emacs image-mode-map "j" 'message-news)
[...]

PR #360, opened 2020-08-05
SPC, DEL, S-SPC scroll binds for DocView

ngenisis:
Currently getting bound to image-scroll-up and image-scroll-down from evil-collection-image.el. Not sure if it is relevant, but evil-collection-image.el has a TODO referencing this closed issue: https://github.com/emacs-evil/evil-collection/issues/23


Ambrevar:
Good catch, thanks!

PR #363, opened 2020-08-12
tab-bar: add evil bindings

condy0919:


Ambrevar:
Is this tab-bar the one from Emacs 27?  If so, we should update the header to mention 27.1 instead of 25.1, and only load tab-bar if the Emacs version meets the requirements.

condy0919:
Updated.

condy0919:
`package-lint` failed due to `Package-Requires: ((emacs "27.1"))`, I opened an issue for this https://github.com/purcell/package-lint/issues/183

condy0919:
It must be `(require 'tab-bar nil t)` due to the Makefile https://github.com/emacs-evil/evil-collection/blob/master/Makefile#L3-L8

condy0919:
@Ambrevar

PR #364, opened 2020-08-13
elfeed: Bind R and r to 'elfeed-search-fetch and 'elfeed-search-updat…

mssdvd:
…e--force

Binding R and r to 'elfeed-search-fetch and 'elfeed-search-update--force should be more ergonomic than gR and gr.


jojojames:
Looks ok to me. @Ambrevar ?

mssdvd:
Yeah, probably is better to stick to the specs. For anyone who is interested in this functionality just add those line to you config:

```elisp
  (evil-define-key 'normal 'elfeed-search-mode-map "R" 'elfeed-search-fetch)
  (evil-define-key 'normal 'elfeed-search-mode-map "r" 'elfeed-search-update--force)
```

Ambrevar:
Thanks.

PR #369, opened 2020-08-31
fix bindings for `buffer-menu'

MorganJamesSmith:
I noticed when using the buffer-menu (M-x buffer-menu) I couldn't use n for evil-search-next or gg for evil-goto-first-line. This patch fixes that.


Ambrevar:
Looks good to me.  @jojojames ?

jojojames:
Thanks!

PR #375, opened 2020-09-08
ec-company: showing popup only when in insertable state

kiennq:
This allows `company-mode` popup only showing up while we're in insertable state.
Related issue https://github.com/emacs-lsp/lsp-mode/issues/1447

### DIFF
--- a/modes/company/evil-collection-company.el
+++ b/modes/company/evil-collection-company.el
@@ -30,7 +30,7 @@
 (require 'company nil t)
 (require 'evil-collection)
 
-(declare-function company-tng-configure-default "company-tng")
+(declare-function company-tng-mode "company-tng")
 
 (defgroup evil-collection-company nil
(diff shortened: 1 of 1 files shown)


modes/company/evil-collection-company.el
```diff
@@ -47,11 +47,22 @@ be set through custom or before evil-collection loads."
   :group 'evil-collection-company
   :type 'boolean)
 
+(defcustom evil-collection-company-supported-states `(insert replace emacs)
```
Ambrevar, line 50: Use a simple quote instead of a backquote since you are not unquoting.

Ambrevar:
Looking good, please fix the nit I pointed out and I'll merge.

modes/company/evil-collection-company.el
kiennq: Done

Ambrevar:
Thanks!

PR #379, opened 2020-09-27
tabulated-list: add more bindings

condy0919:
The original bindings make no sense.

``` elisp
(evil-collection-define-key nil 'tabulated-list-mode-map
  "n" nil
  "p" nil)
```

### DIFF
--- a/modes/tabulated-list/evil-collection-tabulated-list.el
+++ b/modes/tabulated-list/evil-collection-tabulated-list.el
@@ -39,7 +39,14 @@
 
   (evil-collection-define-key nil 'tabulated-list-mode-map
     "n" nil
-    "p" nil))
+    "p" nil)
+
+  (evil-set-initial-state 'tabulated-list-mode 'normal)
+  (evil-collection-define-key 'normal 'tabulated-list-mode-map
+    "S" 'tabulated-list-sort
+    "{" 'tabulated-list-narrow-current-column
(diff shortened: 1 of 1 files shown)


Ambrevar:
There is a reason, see https://github.com/emacs-evil/evil-collection/pull/369.  Do `n` and `p` work properly with you patch?

condy0919:
Added it back.

Ambrevar:
So why did you add it back then? O.o :)

condy0919:
> So why did you add it back then? O.o :)

The map is defined with:

``` elisp
(defvar tabulated-list-mode-map
    (set-keymap-parent map (make-composed-keymap
                            button-buffer-map
                            special-mode-map))
    (define-key map "n" 'next-line)
    (define-key map "p" 'previous-line)
    (define-key map "S" 'tabulated-list-sort)
    (define-key map "}" 'tabulated-list-widen-current-column)
    (define-key map "{" 'tabulated-list-narrow-current-column)
    (define-key map [follow-link] 'mouse-face)
    (define-key map [mouse-2] 'mouse-select-window)
    map)
  "Local keymap for `tabulated-list-mode' buffers.")
```

Binding n/p to `nil` overrides `next-line`/`previous-line`. The other way is to bind n to `evil-search-next` and p to `evil-paste-after`.

Ambrevar:
Thanks!

PR #384, opened 2020-10-05
gnus: R can reply in normal state.

condy0919:
In addition,

- Rebind gnus-summary-mail-forward from C-c C-f to cf
- Remove E gnus-article-read-summary-keys which is not much useful


Ambrevar:
Thanks!

PR #389, opened 2020-10-08
Fix #386: Remove broken (interactive) specs

### LINKED ISSUE #386: Arity errors on evil-delete-backward-char
I have `evil-collection` and `hungry-delete` installed, but I don’t use `hungry-delete-mode`. Whenever I press `X` (`evil-delete-backward-char`) I get an error: `Wrong number of arguments: (2 . 4), 0`.

I believe this is caused by `evil-collection-hungry-delete` and `evil-collection-hungry-delete-for-join` having `(interactive)` specs that override those of `evil-delete-backward-char` and `evil-delete-backward-char-and-join`. In other words, I believe [this line](https://github.com/emacs-evil/evil-collection/blob/e3a521b7e527d0e57341803838f22877d94b2e73/modes/hungry-delete/evil-collection-hungry-delete.el#L38) and [this line](https://github.com/emacs-evil/evil-collection/blob/e3a521b7e527d0e57341803838f22877d94b2e73/modes/hungry-delete/evil-collection-hungry-delete.el#L47) need to be removed.

This is the relevant section of `C-h f add-function`:
> If one of FUNCTION or OLDFUN is interactive, then the resulting function
> is also interactive.  There are 3 cases:
> - FUNCTION is not interactive: the interactive spec of OLDFUN is used.
> - The interactive spec of FUNCTION is itself a function: it should take one
>   argument (the interactive spec of OLDFUN, which it can pass to
>   `advice-eval-interactive-spec`) and return the list of arguments to use.
> - Else, use the interactive spec of FUNCTION and ignore the one of OLDFUN.

Janfel:
Remove `(interactive)` specs of `evil-collection-hungry-delete` and `evil-collection-hungry-delete-for-join` as they break `evil-collection-hungry-delete` and `evil-collection-hungry-delete-for-join` when `hungry-delete-mode` is bound but not enabled.
This fixes issue #386.

### DIFF
--- a/modes/hungry-delete/evil-collection-hungry-delete.el
+++ b/modes/hungry-delete/evil-collection-hungry-delete.el
@@ -35,7 +35,6 @@
 (defun evil-collection-hungry-delete (f &rest args)
   "Wrapper function to run `hungry-delete-backward' if
 `hungry-delete-mode' is on."
-  (interactive)
   (if (and (bound-and-true-p hungry-delete-mode)
            (fboundp 'hungry-delete-backward))
       (hungry-delete-backward 1)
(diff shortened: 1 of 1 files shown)


Ambrevar:
Thanks!

PR #392, opened 2020-10-17
compile: Apply bindings to compilation-minor-mode-map

MorganJamesSmith:
Before this change compilation bindings didn't work in grep-mode which inherits
compilation-minor-mode-map

### DIFF
--- a/modes/compile/evil-collection-compile.el
+++ b/modes/compile/evil-collection-compile.el
@@ -40,6 +40,9 @@
 
   (dolist (keymap evil-collection-compile-maps)
 
+    (evil-collection-define-key nil keymap
+      "g" nil)
+
     (evil-collection-define-key 'normal keymap
       (kbd "RET") 'compile-goto-error
 
--- a/modes/compile/evil-collection-compile.el
+++ b/modes/compile/evil-collection-compile.el
@@ -30,7 +30,8 @@
 (require 'evil-collection)
 (require 'compile)
 
-(defconst evil-collection-compile-maps '(compilation-mode-map compilation-minor-mode-map))
+(defconst evil-collection-compile-maps '(compilation-mode-map
+                                         compilation-minor-mode-map))
 
 ;;;###autoload
 (defun evil-collection-compile-setup ()
@@ -40,12 +41,6 @@
   (dolist (keymap evil-collection-compile-maps)
 
     (evil-collection-define-key 'normal keymap
-      "g?" 'describe-mode
-      "?" evil-collection-evil-search-backward
-      "gg" 'evil-goto-first-line
-      "0" 'evil-digit-argument-or-evil-beginning-of-line
-      [mouse-2] 'compile-goto-error
-      [follow-link] 'mouse-face
       (kbd "RET") 'compile-goto-error
 
       "go" 'compilation-display-error
--- a/modes/compile/evil-collection-compile.el
+++ b/modes/compile/evil-collection-compile.el
@@ -30,32 +30,34 @@
 (require 'evil-collection)
 (require 'compile)
 
-(defconst evil-collection-compile-maps '(compilation-mode-map))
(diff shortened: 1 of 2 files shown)


Ambrevar:
Looking good!

In vanilla Emacs some bindings are not the same between the two maps, e.g. next-error and previous-error are not in `compilation-minor-mode-map`.  Any idea why?

MorganJamesSmith:
I mean the comment in the source code mentions that they're different because of the menu bar which makes sense. However, I'm not sure why the keybindings are different. Looking only at the functions in both, the only two functions that don't appear in minor-mode-map are recompile and compilation-display-error. I think those functions will still work in 90% of compile buffers so I don't see any issues.

Looking at the keybinding in evil-collection-compile.el, I feel like some of them are unnecessary.

I'm not sure how you feel about this, but I think binding "g?" to describe-mode is unnecessary. If you really like that bind, you can just put it in the global map. I'd personally would only bind "g?" to mode specific help functions like 'gnus-summary-help-map.

The bindings for [mouse-2], and [follow-link] are already in the compilation-mode-map, so I think they're redundant.

I don't see why we need to bind "?". It doesn't look to me like it's being rebound anywhere.

We have to rebind all the "g" commands because "g" is bound. I wish there was a way to just unbind "g" from the previous map though. Previously for evil-collection in tabulated-list-mode-map I just did:
```
(evil-collection-define-key nil 'tabulated-list-mode-map
  "n" nil
  "p" nil))
```
But this isn't ideal because it unbinds though keys in emacs mode. I tried defining the key to nil only in 'normal mode but that didn't seem to work.

Ambrevar:
> I'm not sure how you feel about this, but I think binding "g?" to describe-mode is unnecessary. If you really like that bind, you can just put it in the global map. I'd personally would only bind "g?" to mode specific help functions like 'gnus-summary-help-map.

No strong opinion, but what do you want to bind `g?` to?

> The bindings for [mouse-2], and [follow-link] are already in the compilation-mode-map, so I think they're redundant.

OK.

> I don't see why we need to bind "?". It doesn't look to me like it's being rebound anywhere.

@jojojames ?

MorganJamesSmith:
I don't have anything else I want to bind to "g?". I just think that bind is unnecessary since it just binds to the very generic function `describe-mode'.

MorganJamesSmith:
So "?" is actually bound in `special-mode-map' so that can be "fixed" by using 
```
(evil-collection-define-key nil 'special-mode-map
  "?" nil)
```
But again that means that now "?" is unbound in emacs mode. We could fix that by adding a
```
(evil-collection-define-key 'emacs 'special-mode-map
  "?" 'describe-mode)
```
but it'd be nice if we had a function in place that would move bindings from a local map into an Emacs map within the local map. We already "ruin" a few emacs bindings throughout evil-collection so this function would get a good bit of use.

Ambrevar:
Which Emacs bindings do we "ruin"?

Otherwise your `?` fix looks good to me.  Feel free to update your patch with these changes, preferably in separate commits.  Thanks!

MorganJamesSmith:
We "ruin" 
j in magit-todos-section-map and magit-todos-item-section-map
q,j,k,? in which-key-C-h-map
h,l,H,L in indent-rigidly-map
n,p in tabulated-list-mode-map
g,G in edebug-mode-map
C-j,C-k in popup-menu-keymap

And a few more. Almost all calls of:
(evil-collection-define-key nil ...
overwrite the emacs bindings.

MorganJamesSmith:
So I think I have everything working if you are using an evil mode. Do you want me to go through and add some
```
(evil-collection-define-key 'emacs ...
```

so we don't change any functionality when in Emacs mode. Or do you want to make a separate issue for that and fix everything at once.

Ambrevar:
Thanks for the details!

Can you please fix the Emacs bindings in a separate commit?

PR #396, opened 2020-10-25
Preserve emacs bindings 

MorganJamesSmith:
@Ambrevar This is my response to your request (https://github.com/emacs-evil/evil-collection/pull/392#issuecomment-714981635)

This change affects a lot of modes so it should be reviewed thoroughly and with care. It seems to work for me (emacs 27.1), but I wouldn't consider my testing to be thorough. Also a little bikeshedding might be in order (is translate what we're doing? did I name the function right?)

The gist of it is:

We unbind bindings so that the keypress will fall through that keymap and be caught by another keymap. For example, we often unmap "j" so it will fall through to an evil layer that will call a movement function instead of being caught by the current mode. However our current method of unmapping applies to the emacs state as well. Obviously we want a way to unmap "j" in every state except the emacs state.

To accomplish this I've introduced a function called `evil-collection-translate-key-to-state' which allows us to move any key binding from one state to another. This way instead of unbinding keys, we can simply move them to the emacs state.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -365,6 +365,18 @@ to filter keys on the basis of `evil-collection-key-whitelist' and
                                 err))))))
              (add-hook 'after-load-functions fun t))))))
 
+(cl-defun evil-collection-move-bindings-to-emacs-state (map-sym &rest keys)
+  "Move KEYS in MAP-SYM to the `emacs' state."
+  (let* ((keymap (symbol-value map-sym))
+         (bindings (mapcan (lambda (key)
+                             (list key (lookup-key keymap key)))
+                           keys))
+         (unbindings (mapcan (lambda (key)
+                               (list key nil))
+                             keys)))
+    (apply #'evil-collection-define-key nil map-sym unbindings)
+    (apply #'evil-collection-define-key 'emacs map-sym bindings)))
+
 (defun evil-collection-inhibit-insert-state (map-sym)
   "Unmap insertion keys from normal state.
 This is particularly useful for read-only modes."
--- a/modes/company/evil-collection-company.el
+++ b/modes/company/evil-collection-company.el
@@ -69,26 +69,19 @@ be set through custom or before evil-collection loads."
 ;;;###autoload
 (defun evil-collection-company-setup ()
   "Set up `evil' bindings for `company'."
-  (evil-collection-define-key nil 'company-active-map
-    (kbd "C-n") 'company-select-next-or-abort
-    (kbd "C-p") 'company-select-previous-or-abort
-    (kbd "C-j") 'company-select-next-or-abort
-    (kbd "C-k") 'company-select-previous-or-abort
-    (kbd "M-j") 'company-select-next
-    (kbd "M-k") 'company-select-previous)
+  (evil-collection-move-bindings-to-emacs-state 'company-active-map
+    (kbd "C-n") (kbd "C-p") (kbd "C-j") (kbd "C-k") (kbd "M-j") (kbd "M-k"))
 
   (when evil-want-C-u-scroll
-    (evil-collection-define-key nil 'company-active-map (kbd "C-u") 'company-previous-page))
+    (evil-collection-move-bindings-to-emacs-state 'company-active-map
+      (kbd "C-u")))
 
   (when evil-want-C-d-scroll
-    (evil-collection-define-key nil 'company-active-map (kbd "C-d") 'company-next-page))
+    (evil-collection-move-bindings-to-emacs-state 'company-active-map
+      (kbd "C-d")))
 
-  (evil-collection-define-key nil 'company-search-map
-    (kbd "C-j") 'company-select-next-or-abort
-    (kbd "C-k") 'company-select-previous-or-abort
-    (kbd "M-j") 'company-select-next
-    (kbd "M-k") 'company-select-previous
-    (kbd "<escape>") 'company-search-abort)
+  (evil-collection-move-bindings-to-emacs-state 'company-search-map
+    (kbd "C-j") (kbd "C-k") (kbd "M-j") (kbd "M-k") (kbd "<escape>"))
 
   ;; Sets up YCMD like behavior.
   (when evil-collection-company-use-tng
--- a/modes/ivy/evil-collection-ivy.el
+++ b/modes/ivy/evil-collection-ivy.el
@@ -37,8 +37,10 @@
 ;;;###autoload
 (defun evil-collection-ivy-setup ()
   "Set up `evil' bindings for `ivy-mode'."
-  (evil-collection-define-key nil 'ivy-mode-map
-    (kbd "<escape>") 'minibuffer-keyboard-quit)
+
+  (evil-collection-move-bindings-to-emacs-state 'ivy-mode-map
+    (kbd "<escape>"))
+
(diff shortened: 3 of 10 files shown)


jojojames:
> This change affects a lot of modes so it should be reviewed thoroughly and with care. 

Right, we may want to test this a bit before merging and or..-

> (is translate what we're doing? did I name the function right?)

Seems close in naming to some of the other translation functions we have which might be confusing but I dont have any suggestions for naming.

> This way instead of unbinding keys, we can simply move them to the emacs state.

I think it'd be safer, if in your function there was a killswitch variable to get back the old behavior in case we break someone's esoteric workflow or if there's a bug, etc. If this is not feasible, feel free to voice that opinion.

MorganJamesSmith:
I mean if someone really hates the behavior they can use override advice on the function to return the old behavior.
```
;; Not tested
(advice-add 'evil-collection-translate-key-to-state
 :override (lambda (_ _ map-sym &rest keys) (apply #'evil-collection-define-key nil map-sym keys)))
```
 I'd only put a killswitch in if you think it'd be common to not want this functionality. Honestly, I really can't imagine a reason to disable it. It does technically change default behavior, but only in each modes emacs state. If a user wants to override bindings, they should be in a higher priority keymap anyways and I'm not sure how much people use evil's emacs state anyways (I mean no one has complained about the problem this PR fixes yet and I actually don't care about the problem. I just want to code to be proper).

jojojames:
> I mean if someone really hates the behavior they can use override advice on the function to return the old behavior.

Fair enough.

jojojames:
@Ambrevar ?

evil-collection.el
```diff
@@ -363,6 +363,26 @@ to filter keys on the basis of `evil-collection-key-whitelist' and
                                 err))))))
              (add-hook 'after-load-functions fun t))))))
 
+(cl-defun evil-collection-translate-key-to-state (state
+                                                  new-state
```
Ambrevar, line 367: It seems that `state` and `new-state` are always `nil` and `'emacs` respectively.  Can we remove these parameters then?

evil-collection.el
```diff
@@ -363,6 +363,26 @@ to filter keys on the basis of `evil-collection-key-whitelist' and
                                 err))))))
              (add-hook 'after-load-functions fun t))))))
 
+(cl-defun evil-collection-translate-key-to-state (state
```
Ambrevar, line 366: About the name, what about `evil-collection-forward-to-emacs-state`?

Ambrevar:
Other than the above comment, looks good to me, thanks for this fix!

evil-collection.el
MorganJamesSmith: I went with `evil-collection-move-bindings-to-emacs-state` instead because yours sounds like it could be an interactive function.

MorganJamesSmith:
I addressed @Ambrevar 's concerns. Please let me know what you think

MorganJamesSmith:
I made a silly mistake that the CI caught. That's fixed now. Also I've been dog fooding this code for a while and I haven't had any issues personally.

jojojames:
It looks good to me, let me find some time within the next 1-2 weeks (maybe this upcoming weekend) to test it out and merge if it it passes an initial smoke test.

If no response in 1-2 weeks, feel free to ping and I'll get the merge going. Thanks for your patience! (@Ambrevar Feel free to merge at your convenience also)

Ambrevar:
I'm flooded with other projects these days, so I won't be able to sit down and test this thoroughly.

I'll be using it for the next 1-2 weeks so be sure that I'll report if something goes wrong! ;)

condy0919 (changes requested):

modes/indent/evil-collection-indent.el
```diff
@@ -37,11 +37,8 @@
 ;;;###autoload
 (defun evil-collection-indent-setup ()
   "Set up `evil' bindings for autoloaded files."
-  (evil-collection-define-key nil 'indent-rigidly-map
-    "h" 'indent-rigidly-left
-    "l" 'indent-rigidly-right
-    "H" 'indent-rigidly-left-to-tab-stop
-    "L" 'indent-rigidly-right-to-tab-stop))
+  (evil-collection-translate-key-to-state nil 'emacs 'indent-rigidly-map
+    "h" "l" "H" "L"))
```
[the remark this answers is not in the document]
MorganJamesSmith: Thanks for noticing this! I removed that change.

MorganJamesSmith:
My last change is just rebasing on master. No actual modifications

modes/company/evil-collection-company.el
```diff
@@ -69,26 +69,19 @@ be set through custom or before evil-collection loads."
 ;;;###autoload
 (defun evil-collection-company-setup ()
   "Set up `evil' bindings for `company'."
-  (evil-collection-define-key nil 'company-active-map
-    (kbd "C-n") 'company-select-next-or-abort
-    (kbd "C-p") 'company-select-previous-or-abort
-    (kbd "C-j") 'company-select-next-or-abort
-    (kbd "C-k") 'company-select-previous-or-abort
-    (kbd "M-j") 'company-select-next
```
jojojames, line 77: Ran through this fast and not with an emacs-Q setup but some initial thoughts.
C-n/C-p worked for me but the other keys didn't. I bet I have some layer that binds these keys so doing this unbinding drops evil's key priority into some other mode.

[evil-collection] --> [some random potentially uncontrolled layer that binds random keys] --> [emacs state bindings] 

is my guess at the priority. If we focus on only unbinding keys, we'll run into issues where something else have overwritten the emacs bindings.

modes/ivy/evil-collection-ivy.el
```diff
@@ -37,8 +37,10 @@
 ;;;###autoload
 (defun evil-collection-ivy-setup ()
   "Set up `evil' bindings for `ivy-mode'."
-  (evil-collection-define-key nil 'ivy-mode-map
-    (kbd "<escape>") 'minibuffer-keyboard-quit)
+
+  (evil-collection-move-bindings-to-emacs-state 'ivy-mode-map
```
jojojames, line 41: When this is done, it looks like the unbinding gets us to ivy's default keyboard quit which seems to be 3 escape keys instead of the 1 we're expecting.

modes/magit-todos/evil-collection-magit-todos.el
```diff
@@ -48,11 +48,11 @@
 (defun evil-collection-magit-todos-setup ()
   "Set up `evil' bindings for `magit-todos'."
   ;; magit-todos binds jT which prevents evil users from stepping into the section
-  (evil-collection-define-key nil 'magit-todos-section-map
```
jojojames, line 51: This seems like it would work as intended but I don't currently have this package installed.

jojojames:
> For example, we often unmap "j" so it will fall through to an evil layer that will call a movement function instead of being caught by the current mode. However our current method of unmapping applies to the emacs state as well. Obviously we want a way to unmap "j" in every state except the emacs state.

Looking more carefully at this, I bet there are scenarios where we're not necessarily unbinding stuff but trying to get the highest keybinding precedence.

The two examples I posted above (company & ivy) are ones where unbinding is not necessarily what we want but it does seem like we want this type of functionality (magit-todos is a package that could potentially use it).

PR #397, opened 2020-11-06
occur: bind q to quit

condy0919:
The quit commands are missing in `occur-mode`

### DIFF
--- a/modes/occur/evil-collection-occur.el
+++ b/modes/occur/evil-collection-occur.el
@@ -45,6 +45,9 @@
     ;; Like `wdired-mode'.
     (kbd "C-x C-q") 'occur-edit-mode
 
+    ;; quit
+    "q" 'quit-window
+
     [mouse-2] 'occur-mode-mouse-goto
     (kbd "C-c C-c") 'occur-mode-goto-occurrence
 


Ambrevar:
Works for me:

- `q`: `quit-window`
- `ZQ`: `evil-quit`
- `ZZ`: `evil-save-modified-and-close`

Have you tried with a vanilla config?

condy0919:
The `occur-mode-map` has no `quit-window` binding in normal state.

``` text
key             binding
---             -------

C-c             Prefix Command
C-o             occur-mode-display-occurrence
  (this binding is currently shadowed)
ESC             Prefix Command
SPC             scroll-up-command
  (this binding is currently shadowed)
-               negative-argument
  (this binding is currently shadowed)
<               beginning-of-buffer
  (this binding is currently shadowed)
>               end-of-buffer
  (this binding is currently shadowed)
e               occur-edit-mode
  (this binding is currently shadowed)
g               Prefix Command
n               next-error-no-select
  (this binding is currently shadowed)
o               occur-mode-goto-occurrence-other-window
  (this binding is currently shadowed)
p               previous-error-no-select
  (this binding is currently shadowed)
q               quit-window
  (this binding is currently shadowed)
DEL             scroll-down-command
  (this binding is currently shadowed)
S-SPC           scroll-down-command
<normal-state>  Prefix Command
<remap>         Prefix Command

<normal-state> C-c              Prefix Command
<normal-state> C-j              occur-next
<normal-state> C-k              occur-prev
<normal-state> RET              occur-mode-goto-occurrence
<normal-state> C-x              Prefix Command
<normal-state> c                clone-buffer
<normal-state> g                Prefix Command
<normal-state> r                occur-rename-buffer
<normal-state> <M-return>       occur-mode-display-occurrence
<normal-state> <S-return>       occur-mode-goto-occurrence-other-window
<normal-state> <mouse-2>        occur-mode-mouse-goto

M-n             occur-next
M-p             occur-prev

g r             revert-buffer

<normal-state> g j              occur-next
<normal-state> g k              occur-prev
<normal-state> g o              occur-mode-goto-occurrence-other-window

<normal-state> C-x C-q          occur-edit-mode

<normal-state> C-c C-c          occur-mode-goto-occurrence
<normal-state> C-c C-e          occur-edit-mode
<normal-state> C-c C-f          next-error-follow-minor-mode
```

condy0919:
In `occur-mode-map`

```
q runs the command evil-record-macro (found in evil-normal-state-map), which is
an interactive compiled Lisp function in ‘evil-commands.el’.

It is bound to q.

(evil-record-macro REGISTER)

Record a keyboard macro into REGISTER.
If REGISTER is :, /, or ?, the corresponding command line window
will be opened instead.
```

While

- <kbd>ZZ</kbd> `evil-save-modified-and-close`
- <kbd>ZQ</kbd> `evil-quit`

0x28:
I have the same problem. Might have something to do with the commit 8370d6bd1cfbc9eb5d3f8b1701312f254bec9cb8 where the `quit-window` binding was removed from `special-mode-map`. `occur-mode` is derived from `special-mode`.

Ambrevar:
@0x28 Indeed, #392 seems to be the culprit.

@MorganJamesSmith Any reason you removed the `q` binding from special-mode-map?  Did it get in the way of something?

MorganJamesSmith:
There was no real reason to remove the `q` binding. I saw that it was already defined in `special-mode-map` and felt the need to remove code I thought was redundant. On my emacs (27.1) the value of special mode map has the appropriate quit-window binding `(113 . quit-window)`. However, none of the bindings from `special-mode-map` seem to make it into `occur-mode-map`. Although `occur-mode-map` is a derived-mode, they did not inherit the keymap (I'm not sure why). I've done no testing, but I believe this issue existed before my commit.

Ambrevar:
I like @MorganJamesSmith suggestion.

@condy0919 What do you think of binding `q` in `evil-collection-inhibit-insert-state` and call the latter over special-mode-map?

condy0919:
It feels weird, `q` has nothing to do with insert but bound to `quit-window` by
`evil-collection-inhibit-insert-state`. Further more, `q` still needs to be
rebound to mode specific quit command.

``` shell
# in evil-collection/modes/
rg '"q"' | grep quit-window -c      # 93
rg '"q"' | grep quit-window -v -c   # 68
```

There are still 68 modes left to be tuned by ourselves.

MorganJamesSmith:
Everything is in a hierarchy. All we are doing is putting `quit-window` above `evil-record-macro` when insert is disabled. Any case where we already modify `q` will be above this change so we don't have to touch those 68 modes. If anything, we can remove the `quit-window` binding in the other 25 modes.

I think 99% of the usefulness of macros goes away when I'm not modifying, but if we centralize where we set it (like in the `evil-collection-inhibit-insert-state` function) than it's trivial to add a variable that enables and disables this behavior. I do agree with you that maybe it's odd to put it in `evil-collection-inhibit-insert-state` but I can't think of a better place to put it. Maybe we could just rename it to `evil-collection-set-readonly-bindings`

Ambrevar:
Morgan Smith <notifications@github.com> writes:

> Maybe we could just rename it to `evil-collection-set-readonly-bindings`

I think it's a good idea.

PR #398, opened 2020-11-11
read-only mode bindings

condy0919:
Follow guides in #397 

There are tons of bindings to be refined though.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -391,6 +391,17 @@ This is particularly useful for read-only modes."
     [remap evil-shift-right] #'ignore
     [remap evil-invert-char] #'ignore))
 
+(defun evil-collection-set-readonly-bindings (map-sym)
+  "Unmap insertion keys from normal state. Additionally q can `quit-window'.
+This is particularly useful for read-only modes. Make sure it's
+called before setting up other evil bindings so that it can be
+overriden."
+  (evil-collection-inhibit-insert-state map-sym)
+  (evil-collection-define-key 'normal map-sym
+    "q"  #'quit-window
+    "ZZ" #'quit-window
+    "ZQ" #'evil-quit))
+
 (defun evil-collection--binding-lessp (a b)
   "Comparison function used to sort bindings of the form (state key def)."
   (let ((a-state (symbol-name (nth 0 a)))
--- a/modes/calc/evil-collection-calc.el
+++ b/modes/calc/evil-collection-calc.el
@@ -34,7 +34,7 @@
 ;;;###autoload
 (defun evil-collection-calc-setup ()
   "Set up `evil' bindings for `calc'."
-  (evil-collection-inhibit-insert-state 'calc-mode-map)
+  (evil-collection-set-readonly-bindings 'calc-mode-map)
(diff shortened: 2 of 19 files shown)


condy0919:
@Ambrevar @MorganJamesSmith

MorganJamesSmith:
I don't think you should remap `evil-record-macro`. The idea here is to make `q` quit, not overwrite `evil-record-macro`. However, if we just do an `evil-collection-define-key` for `q` then we have to make sure to call `evil-collection-set-readonly-bindings` before we setup any other bindings or else the readonly bindings will end up on top.

Is there a way to make the readonly bindings a proper keymap that can sit under a keymap? Can you retroactively decide inheritance?

I'm probably not qualified to review pull requests yet. Emacs is still very mysterious to me :P

condy0919:
`evil-record-macro` is bound to `q` in defaults, using `[remap evil-record-macro]` is intended to keep consistent with other remaps. I don't think anyone could bind `evil-record-macro` to other keys, so it's acceptable.

Aftering looking through https://www.masteringemacs.org/article/mastering-key-bindings-emacs, `minor-mode-overriding-map-alist` is a choice.

@Ambrevar Any thoughts?

jojojames:
I'd prefer the list of bindings/modes to be changed to be in the same CL so we can see what the blast radius of this change is. If it's actually really small, it'd beg the question whether we need this convenience function also.

The remap on record-macro seems spotty to me (I could definitely be convinced otherwise), if we make the assumption that most people's record-macro is on q, it seems it'd be easier to just make the assumption it's safe to rebind q and avoid any subtle break when someone does rebind record-macro. Just my quick knee-jerk reaction though.

condy0919:
Can `evil-quit` be replaced with `quit-window`?
@jojojames @Ambrevar

Ambrevar:
I agree with `quit-window`.  In fact it's what we are using in most of the code, but with some inconsistencies.

For instance hackernews and youtube-dl both bind ZZ to evil-quit.

I suggest the following rational update:

- Bind `q` to the mode specific quitting function, `quit-window` if there is none.
- Bind `ZZ` to `quit-window`.
- Bind `ZQ` to `evil-quit`.

Thoughts?

condy0919:
> * Bind `q` to the mode specific quitting function, `quit-window` if there is none.
> * Bind `ZZ` to `quit-window`.
> * Bind `ZQ` to `evil-quit`.

Do agree.

condy0919:
@jojojames @Ambrevar 

Most of modes are refined.

MorganJamesSmith:
I still think it's worth pointing out that `evil-collection-set-readonly-bindings` should be called before setting up other keybinds so that it can be overriden. Adding this information to the docstring would make me happy

Ambrevar:
Can you also update the README rationale?

condy0919:
> I still think it's worth pointing out that `evil-collection-set-readonly-bindings` should be called before setting up other keybinds so that it can be overriden. Adding this information to the docstring would make me happy

Mentioned.

condy0919:
> Can you also update the README rationale?

Updated.

Ambrevar:
Excellent, thanks!

PR #401, opened 2020-11-14
unimpaired: [b and ]b can be called with prefix arg

condy0919:
<kbd>C-u 3 ]b</kbd> can switch to the next 3rd buffer.


Ambrevar:
Thanks!

PR #402, opened 2020-11-14
Add shortdoc evil bindings

condy0919:
`shortdoc` comes from Emacs 28 which has not been released yet so I temporaily
set `Package-Requires` to ``((emacs "27.1"))`` otherwise `package-lint` will
fail.

Additionally the trailing whitespaces in org-present mode are removed.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -237,6 +237,7 @@ through removing their entry from `evil-collection-mode-list'."
     robe
     rtags
     ruby-mode
+    ,@(when (>= emacs-major-version 28) '(shortdoc))
     simple
     slime
     sly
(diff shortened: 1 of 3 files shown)


Ambrevar:
Looking good!

Is there a way to silence package-lint for this particular case?
If not, can you add a TODO: mentioning that we need to update the Package-Requires line when Emacs 28 is out?

condy0919:
https://github.com/purcell/package-lint/blob/master/package-lint.el#L32-L33

> ;; Checks will currently be enabled only if a "Package-Requires:" or
> ;; "Package-Version:" header is present in the file.

The commentary of `package-lint` says it can be disabled by removing `Package-Requires:` but travis complains

> evil-collection-shortdoc.el:1:67: warning: You should depend on (emacs "24.1") if you need lexical-binding.

Ambrevar:
Thanks!

PR #404, opened 2020-11-21
unimpaired: exchange line

condy0919:
The implementation is stolen from [here](https://stackoverflow.com/questions/2423834/move-line-region-up-and-down-in-emacs). Maybe it's not in evil style, revising is welcome.

@jojojames @Ambrevar

### DIFF
--- a/modes/unimpaired/evil-collection-unimpaired.el
+++ b/modes/unimpaired/evil-collection-unimpaired.el
@@ -112,6 +112,41 @@
   (ignore count)
   (evil-collection-unimpaired--encode beg end #'url-unhex-string))
 
+;; https://stackoverflow.com/questions/2423834/move-line-region-up-and-down-in-emacs
+(defun evil-collection-unimpaired--move-text (arg)
+  "Move text down if ARG is positive, otherwise move text up."
+  (cond
+   ((and mark-active transient-mark-mode)
+    (when (> (point) (mark))
+      (exchange-point-and-mark))
+    (let ((column (current-column))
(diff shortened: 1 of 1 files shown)


modes/unimpaired/evil-collection-unimpaired.el
```diff
@@ -112,17 +112,57 @@
   (ignore count)
   (evil-collection-unimpaired--encode beg end #'url-unhex-string))
 
+;; https://stackoverflow.com/questions/2423834/move-line-region-up-and-down-in-emacs
+(defun evil-collection-unimpaired--move-text (arg)
+  "Move text down if ARG is positive, otherwise move text up."
+  (cond
+   ((and mark-active transient-mark-mode)
+    (if (> (point) (mark))
```
Ambrevar, line 120: `if` -> `when`, because there is no "else" clause.

modes/unimpaired/evil-collection-unimpaired.el
```diff
@@ -112,17 +112,57 @@
...
+        (exchange-point-and-mark))
+    (let ((column (current-column))
+          (text (delete-and-extract-region (point) (mark))))
+      (forward-line arg)
+      (move-to-column column t)
```
Ambrevar, line 125: Style: I prefer to replace `t` with `:force` to make it more readable, without looking at the function signature.

Ambrevar:
Looking great!  Fix the nits and I'll merge :)

condy0919:
Fixed!

PR #405, opened 2020-11-25
prodigy: add more keybindings to prodigy-view-mode from prodigy-mode

danielpza:
Hey there, I added some keybindings that were missing on prodigy-view-mode compared to prodigy-mode. This mode is different from prodigy-mode but has similar actions.


jojojames:
Thanks!

Issue #408, opened 2020-12-02
Occur mode C-j and C-k


countvajhula:
Hi there,
Amazing work on this package 😄 🥇 

I noticed that in occur mode, with e.g. `M-x occur <RET> defun` to see a list of functions in the current file, `C-j` and `C-k` call `occur-next` and `occur-previous`, which don't seem to navigate to the corresponding functions in the original window. They don't really seem to do much and are equivalent to `j` and `k` as far as I can tell. On the other hand, [this link](https://www.emacswiki.org/emacs/OccurMode) mentions that `next-error` and `previous-error` will work that way, and I believe this is how `C-j` and `C-k` work for me in e.g. an ivy occur buffer.

So, how about changing this behavior in occur-mode to use `next-error` and `previous-error` instead? I checked and these do work as expected, but they're bound by default to C-x ` and M-g p (!).

Hope this is the right place to discuss this issue.

jojojames:
Sounds good to me, please PR.

countvajhula:
Let me see what I can do @jojojames . Have any PRs / docs / code samples you can point me to as a starting point?

jojojames:
I recommend taking a look at random CLs or smaller/simpler modes in the repo and using those as examples. Past PRs may be helpful.

PR #410, opened 2020-12-04
Occur mode C-j and C-k

countvajhula:
This fixes the C-j and C-k behavior discussed in #408 .

I went with the `-no-select` versions of `next-error` and `previous-error` since these retain focus in the occur buffer while navigating the results with preview (as opposed to visiting the locations and moving focus to the original window).


jojojames:
Thanks!

PR #411, opened 2020-12-04
Add basic support for dictionary-mode

countvajhula:
This is for Emacs's standard [dictionary-mode](https://www.emacswiki.org/emacs/DictMode). The actual major mode is named `dictionary-mode` notwithstanding the title of that wiki page.

Dictionary mode binds `h` to `dictionary-help` which shows the default keybindings for the mode - most of which are still active even under the evil bindings so this is still relevant. Since `h` should be bound to `backward-char` under evil, I bound that function to `?`. Is there a convention or precedent to follow here, or is this fine as is?

PR #413, opened 2020-12-07
Suppress dictionary nonexistence error

condy0919:
#411 adds evil bindings to dictionary-mode without `(require 'dictionary nil t)` which will lead an error:

``` text
Cannot open load file: No such file or directory, dictionary
```


jojojames:
Thanks @condy0919 !

PR #414, opened 2020-12-08
Add basic support for ripgrep

countvajhula:
[rg](https://github.com/dajva/rg.el) and [deadgrep](https://github.com/Wilfred/deadgrep) are already supported. This adds support for [ripgrep](https://github.com/nlamirault/ripgrep.el), yet another ripgrep frontend, and the one that happens to be [used by projectile](https://github.com/bbatsov/projectile/blob/0ff73ecf7f25f47382e2d28c8ab232c18609d515/projectile.el#L3428-L3437).

I followed the `grep` recipe to add minimal support for navigating the search results.

I also removed `dictionary` from the Cask dependencies since it looks like with [this change](https://github.com/emacs-evil/evil-collection/pull/413), it isn't needed anymore.

### DIFF
--- a/modes/ripgrep/evil-collection-ripgrep.el
+++ b/modes/ripgrep/evil-collection-ripgrep.el
@@ -1,0 +1,45 @@
+;;; evil-collection-ripgrep.el --- Bindings for `ripgrep' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018 James Nguyen
+
+;; Author: Sid Kasivajhula <sid@countvajhula.com>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <mail@ambrevar.xyz>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, emacs, tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Bindings for `ripgrep'.
+
+;;; Code:
+(require 'evil-collection)
+(require 'ripgrep nil t)
+
+(defconst evil-collection-ripgrep-maps '(ripgrep-search-mode-map))
+
+;;;###autoload
+(defun evil-collection-ripgrep-setup ()
+  "Set up `evil' bindings for `ripgrep'."
+  (evil-collection-define-key 'normal 'ripgrep-search-mode-map
+    "n" 'evil-search-next
+    "\C-j" 'next-error-no-select
+    "\C-k" 'previous-error-no-select))
+
+
+(provide 'evil-collection-ripgrep)
+;;; evil-collection-ripgrep.el ends here
--- a/modes/ripgrep/evil-collection-ripgrep.el
+++ b/modes/ripgrep/evil-collection-ripgrep.el
@@ -37,8 +37,13 @@
   "Set up `evil' bindings for `ripgrep'."
   (evil-collection-define-key 'normal 'ripgrep-search-mode-map
     "n" 'evil-search-next
+    "\C-k" 'previous-error-no-select
     "\C-j" 'next-error-no-select
-    "\C-k" 'previous-error-no-select))
+    "gk" 'previous-error-no-select
+    "gj" 'next-error-no-select
+    "{" 'compilation-previous-file
+    "}" 'compilation-next-file
+    "i" 'wgrep-change-to-wgrep-mode))
 
 
 (provide 'evil-collection-ripgrep)
--- a/modes/ripgrep/evil-collection-ripgrep.el
+++ b/modes/ripgrep/evil-collection-ripgrep.el
@@ -1,7 +1,5 @@
 ;;; evil-collection-ripgrep.el --- Bindings for `ripgrep' -*- lexical-binding: t -*-
 
-;; Copyright (C) 2018 James Nguyen
-
 ;; Author: Sid Kasivajhula <sid@countvajhula.com>
 ;; Maintainer: James Nguyen <james@jojojames.com>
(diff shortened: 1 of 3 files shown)


modes/ripgrep/evil-collection-ripgrep.el
```diff
@@ -0,0 +1,45 @@
...
+
+;;;###autoload
+(defun evil-collection-ripgrep-setup ()
+  "Set up `evil' bindings for `ripgrep'."
+  (evil-collection-define-key 'normal 'ripgrep-search-mode-map
+    "n" 'evil-search-next
+    "\C-j" 'next-error-no-select
+    "\C-k" 'previous-error-no-select))
+
+
```
jojojames, line 43: Looks like these are the bindings.

``` (define-key map "p" 'compilation-previous-error)
    (define-key map "n" 'compilation-next-error)
    (define-key map "{" 'compilation-previous-file)
    (define-key map "}" 'compilation-next-file)
    (define-key map "k" 'ripgrep/kill-buffer)
    (define-key map "C-c C-p" 'wgrep-change-to-wgrep-mode)```

Are these just not bound in this compatibility layer?

Would like to see ```wgrep-change-to-wgrep-mode``` on 'i'.

Comment (countvajhula) [reply modes/ripgrep/evil-collection-ripgrep.el:48]: These bindings are present in the projectile wrapped version too.
I added `{` and `}` for file navigation, and `wgrep-change-to-wgrep-mode` on `i`.
`compilation-next-error` and `compilation-previous-error` don't seem to do very much, essentially equivalent to `j` and `k`.
Did you have something in mind for `ripgrep/kill-buffer`?

File: modes/ripgrep/evil-collection-ripgrep.el [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+;;;###autoload
+(defun evil-collection-ripgrep-setup ()
+  "Set up `evil' bindings for `ripgrep'."
+  (evil-collection-define-key 'normal 'ripgrep-search-mode-map
+    "n" 'evil-search-next
+    "\C-k" 'previous-error-no-select
+    "\C-j" 'next-error-no-select
+    "gk" 'previous-error-no-select
+    "gj" 'next-error-no-select
+    "{" 'compilation-previous-file
```
jojojames, line 44: Not super sure about this off the top of my head. I'd recommend looking at other files for inspiration. (If other files use these bindings, feel free to resolve.)

jojojames: re: { and }

modes/ripgrep/evil-collection-ripgrep.el
jojojames: Maybe 'q' if that's the right one to kill the buffer/similar to quit-window.

modes/ripgrep/evil-collection-ripgrep.el
```diff
@@ -0,0 +1,50 @@
+;;; evil-collection-ripgrep.el --- Bindings for `ripgrep' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2018 James Nguyen
```
Ambrevar, line 3: Set the year and the name to yours ;)

countvajhula: I actually don't claim copyright on [any of my work](https://github.com/countvajhula/symex.el/blob/6816417e71c4cb28f30c6d7c9917ca71f3d6550f/symex.el#L9-L23) and would prefer not to. Is it OK if I just leave the line as is and update the year? Another possibility is to eliminate the line entirely, as I do in my Emacs extensions.

modes/ripgrep/evil-collection-ripgrep.el
countvajhula: `q` does `quit-window` while in ripgrep mode. Looking at the value of `ripgrep-search-mode-map`, I see this:
```
...
 (normal-state keymap "Auxiliary keymap for Normal state"
			   (90 keymap
				   (81 . evil-quit)
				   (90 . quit-window))
			   (113 . quit-window)
...
```

113 is `q`.

Is this fine or do we want to replace the binding with `ripgrep/kill-buffer`?

modes/ripgrep/evil-collection-ripgrep.el
countvajhula: I lifted that one [from rg](https://github.com/emacs-evil/evil-collection/blob/5151be9b633a8b530d6050e55178a7b5056c10cc/modes/rg/evil-collection-rg.el#L62-L63).

modes/ripgrep/evil-collection-ripgrep.el
Ambrevar: As far as I understand it, the copyright line must appear for GPL work.  I'm not sure how it works if you want to surrender the rights to someone else, maybe it's not as straightforward as you did it here.  With the Free Software Foundation there is a piece of paper to sign, don't know if it would be the same here.

GPL is different from public domain in that it takes a political stance against proprietary licensing.

What this copyright line effectively does, if I understand correctly, is grant you the right to change the license to this piece of code, that's it.

countvajhula: Nowadays we no longer need to put an explicit copyright line for copyright to still apply (e.g. [this link](https://www.newmediarights.org/business_models/artist/do_i_need_put_copyright_notice_my_work#:~:text=No.,longer%20necessary%20to%20receive%20protection.) talks about this). It's taken as a given from a legal perspective that at a minimum, copyright always applies for any creative work (but not patents etc). [This link](https://www.copyright.gov/help/faq/faq-general.html) also makes that explicit.

So the main reason to have a copyright line seems to be to simply record the names of the authors and the year of authorship as it relates to copyright. But that is already accomplished with version control much more reliably, and with [cryptographically secure hashes](https://gist.github.com/masak/2415865) (well, almost) recording uniquely identifying information about each commit, so it's much more reliable than a manually maintained list of authors for every file (which is probably inaccurate at scale anyway). This line may at this point just be a relic from traditional media. WDYT?

countvajhula: Nowadays we no longer need to put an explicit copyright line for copyright to still apply (e.g. [this link](https://www.newmediarights.org/business_models/artist/do_i_need_put_copyright_notice_my_work#:~:text=No.,longer%20necessary%20to%20receive%20protection.) talks about this). It's taken as a given from a legal perspective that at a minimum, copyright always applies for any creative work (but not patents etc). [This link](https://www.copyright.gov/help/faq/faq-general.html) also makes that explicit.

So the main reason to have a copyright line seems to be to simply record the names of the authors and the year of authorship as it relates to copyright. But that is already accomplished with version control much more reliably, and with [cryptographically secure hashes](https://gist.github.com/masak/2415865) (well, almost) recording uniquely identifying information about each commit, so it's much more reliable than a manually maintained list of authors for every file (which is probably inaccurate at scale anyway). This line may at this point just be a relic from traditional media.

jojojames:
Thanks.

PR #418, opened 2020-12-11
magit: (visual) set "y" to magit-copy-section-value

qwerbzuio:
The behavior of magit-copy-section-value in visual-mode is slightly different
from a simple yank: if called with universal-argument, it removes any leading
+/- from diff regions, which is quite helpful. I think "y" should implement this
behavior.

This gives a minor difference: at the end of the copied region there is now no
trailing newline, where before there was one.

----

#


jojojames:
@justbur ?

Ambrevar:
Looks good to me!

We already have magit-copy-section-value bound to `y s` in normal mode, which seems useless to me, but maybe I'm missing the point.
If it is indeed useless, I suggest we remove the `y s` binding!

justbur:
I use that function in normal mode, but don't need it. It's an easy way to get the hash of a commit.

Ambrevar:
I see, it gets the commit in the magit-status buffer.
However in magit-revision-mode it yanks garbage like

(nil (113 6) (113 7))

We could unbind it in magit-revision-mode then.

justbur:
According to the docstring there seems to be a good reason for the value
yanked in magit-revision-mode. I'm not that familiar with using it, but I
wouldn't immediately assume it's useless.

qwerbzuio:
Is there anything left that needs clearance in order for this to get merged?

modes/magit/evil-collection-magit.el
```diff
@@ -319,6 +319,7 @@ moment.")
        (,states magit-status-mode-map "gh"  magit-section-up                       "^")
        (,states magit-diff-mode-map "gj" magit-section-forward)
        (,states magit-diff-mode-map "gd" magit-jump-to-diffstat-or-diff "j")
+       ((visual) magit-diff-mode-map "y" magit-copy-section-value)
```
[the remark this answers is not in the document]
jojojames: Should be ok I think..

jojojames, line 322: Do you have a simple set of instructions to get to magit-diff-mode?

modes/magit/evil-collection-magit.el
```diff
@@ -342,7 +343,7 @@ moment.")
            (,states magit-mode-map "yr"   magit-show-refs            "y")
            (,states magit-mode-map "ys"   magit-copy-section-value   "C-w")
            (,states magit-mode-map "yb"   magit-copy-buffer-revision "M-w")
-           ((visual) magit-mode-map "y"   evil-yank))
+           ((visual) magit-mode-map "y"   magit-copy-section-value))
```
jojojames, line 346: This seems like it might be overkill, I'm imagining in magit-mode, there might be reasons to use evil-yank (even in visual state). Maybe I'm misunderstanding everything magit-copy-section-value can do. As I understand, ys should be bound to that function right?

jojojames:
@qwerbzuio Sorry for the wait, I had been letting others comment on it. Didn't really look at the details until today. Left some comments.

modes/magit/evil-collection-magit.el
condy0919: `ys` is a motion key which is not proper in such a case

condy0919: > This seems like it might be overkill, I'm imagining in magit-mode, there might be reasons to use evil-yank (even in visual state). Maybe I'm misunderstanding everything magit-copy-section-value can do. As I understand, ys should be bound to that function right?

`ys` is a motion key which is not proper in such a case

condy0919 (commented):
Some comments left

modes/magit/evil-collection-magit.el
jojojames: What does "motion key" mean in this case? example?

For example:

```
y s runs the command magit-copy-section-value, which is an autoloaded
interactive Lisp function in ‘magit-extras.el’.

It is bound to y s, <normal-state> y s.

(magit-copy-section-value ARG)
```

modes/magit/evil-collection-magit.el
qwerbzuio: > Do you have a simple set of instructions to get to magit-diff-mode?

In a magit-status buffer, navigate to a commit line, then invoke `dd`. This will open a magit-diff buffer.

modes/magit/evil-collection-magit.el
qwerbzuio: Concerning the overkill-argument, `magit-copy-section-value` without a prefix-arg simply forwards to `copy-region-as-kill` if a transient-region is active (same as visual state?). However, `evil-yank` is different from `copy-region-as-kill`, so if we don't want to shadow it completely, we'd need to use a different binding, like for example the suggested `ys`, and keep `evil-yank` on `yy` or similar.

One point where `evil-yank` would be necessary is in block-mode, in order to properly yank the selected block. However, horizontal navigation with `h` and `l` is anyway not possible in magit-buffers (it is still possible with cursor keys...). So I'd raise the question if it is ok to accept the forwarding to `copy-region-as-kill` and shadow `evil-yank` completely.

But I'd also be happy with the first approach to have different bindings.

jojojames: OK, if that's true it might be OK to use magit-copy-section-value. Lemme find some time to pull your PR and play with it myself to see if there are any extra subtleties.

jojojames:
Thanks and sorry for the wait @qwerbzuio

qwerbzuio:
@jojojames Thank you very much for your review and the merge! And no worries for the delay.

PR #419, opened 2020-12-14
magit: make gj/gk behave consistently

qwerbzuio:
Both should do the same thing, which is magit-section-{forward,backward}-sibling.

----

#


Ambrevar:
I guess this was a bug.  @justbur Do you confirm?

justbur:
Yes, it should be fixed. I want to say that the backward version didn't exist at the time, but I can't remember

jojojames:
OK to merge @justbur ?

justbur:
Looks fine to me. By the way I barely have time for the repos I own so feel free to manage this one without me.

PR #417, opened 2020-12-10
Update notmuch binds to reflect changes in notmuch

meain:
I was able to track this change to commit [adfded9ed0a5a4b06886f462314cd4511cb72d47](https://git.notmuchmail.org/git?p=notmuch;a=commit;h=adfded9ed0a5a4b06886f462314cd4511cb72d47).

I am kinda new to elisp, let me know if there are any issues.
Fixes https://github.com/emacs-evil/evil-collection/issues/415

PR #420, opened 2020-12-14
Setup Github Actions instead of Traivs

meain:
When I submitted https://github.com/emacs-evil/evil-collection/pull/417 I had to wait a really long time to get any response back from Travis. That is why I thought I would rewrite the checks into Github actions. This should be much faster. Also, added `27.1` to the list of versions we are checking along with `26.3`.


meain:
I think this will have to be merged for the PR runs to start kicking in, but here is a sample run: https://github.com/meain/evil-collection/pull/1

jojojames:
This looks pretty cool, I'm out of the loop on this. Are github actions the new recommended approach to set up CI? (Especially for emacs projects?)

meain:
I have seen a lot of people switch over to Github actions recently in general. Not exactly sure about if this is the case for Emacs projects specifically, but [a search](https://github.com/search?l=&q=purcell/setup-emacs+++filename%3A*.yml&type=Code) for the "plugin" that lets you setup Emacs in a Github action shows some popular ones using it.`quelpa`, `org-ql`, `org-roam`, `dap-mode` and even `doom` is in the list.

Ambrevar:
No strong opinion myself, I've only used GitHub actions before, I've
never used Travis.

jojojames:
Sounds good to me, I'm not familiar with the github actions syntax. If it looks fine to you @Ambrevar , feel free to merge. Thanks @meain for working on this!

PR #421, opened 2020-12-17
edebug: "h" should behave the same as j, k, l

qwerbzuio:
j, k, l are used for navigation in edebug-mode and so should do h.
However, currently it performs "edebug-goto-here", which is also bound
to H. Therefore, "h" is free to be used for navigation.


Ambrevar:
Good catch, thanks!

PR #424, opened 2020-12-18
newsticker: add treeview bindings

condy0919:
``` emacs-lisp
k newsticker-previous-item
j newsticker-next-item
```

The bindings are removed

### DIFF
--- a/modes/newsticker/evil-collection-newsticker.el
+++ b/modes/newsticker/evil-collection-newsticker.el
@@ -35,28 +35,74 @@
 ;;;###autoload
 (defun evil-collection-newsticker-setup ()
   "Set up `evil' bindings for `newsticker'."
+  ;; plainview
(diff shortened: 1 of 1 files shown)


Ambrevar:
Are you saying the `newsticker-previous-item` and `newsticker-next-item` are commands that don't exist anymore?
If so, how do we go to the next/previous item now?  Are `k` and `j` doing the right thing out of the box?

Ambrevar:
Otherwise looks good to me!

condy0919:
> Are you saying the `newsticker-previous-item` and `newsticker-next-item` are commands that don't exist anymore?
> If so, how do we go to the next/previous item now? Are `k` and `j` doing the right thing out of the box?

Yes, `j` `k` works out of box.

Ambrevar:
Thanks!

PR #426, opened 2020-12-20
outline: Optionally enable keybindings in Outline minor mode

fishyfriend:
Hi,

This PR adds a setting that exposes the Outline mode bindings in Outline minor mode. It's turned on by default. As noted in the docstring, a user can override or exclude specific bindings with `evil-define-minor-mode-key`.

Thanks!

*EDIT*: simplified/corrected the new setting's docstring

### DIFF
--- a/modes/outline/evil-collection-outline.el
+++ b/modes/outline/evil-collection-outline.el
@@ -40,6 +40,18 @@ mode."
   :group 'evil-collection-outline
   :type 'boolean)
 
+(defcustom evil-collection-outline-enable-in-minor-mode-p t
+  "Enable Outline mode bindings in Outline minor mode.
+
+To override individual bindings, modify the minor mode keymap.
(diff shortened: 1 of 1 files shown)


Ambrevar:
Looks good to me.

Thoughts on this, @jojojames?

jojojames:
I don't use outline mode enough but it looks OK to me. @Ambrevar OK to merge.

yuhan0:
These are very intrusive keybindings, and I would strongly prefer if they were left off by default. 
Outline-minor-mode should be able to play well with all other programming major modes without hijacking keys like `M-RET` and `gj`/`gk` in the way that might make sense for the major Outline-mode.

jojojames:
> Outline-minor-mode should be able to play well with all other programming major modes without hijacking keys like M-RET and gj/gk in the way that might make sense for the major Outline-mode.

Could you elaborate in a way for someone that doesn't use outline-mode to understand?

PR #428, opened 2020-12-24
Put notmuch-common-keymap last

gagbo:
Since all the other maps in `evil-collection-notmuch-maps` 
have notmuch-common-keymap as parent
https://git.notmuchmail.org/git?p=notmuch;a=blob;f=emacs/notmuch-hello.el;h=fa31694ff0f79a7f2781849394390a673f75aade;hb=HEAD#l651

When `evil-collection-setup-hook` runs, if it changes keymaps
using a `dolist` over `evil-collection-notmuch-maps`, then a binding might
see 2 changes at once.

See discussion here https://github.com/hlissner/doom-emacs/pull/4403#issuecomment-750907188

### DIFF
--- a/modes/notmuch/evil-collection-notmuch.el
+++ b/modes/notmuch/evil-collection-notmuch.el
@@ -36,13 +36,20 @@
 (declare-function notmuch-search-tag "notmuch")
 (declare-function notmuch-tree-tag "notmuch-tree")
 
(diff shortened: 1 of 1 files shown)


Ambrevar:
Good catch!  Can you just add a comment in the code to explain this change?  Thanks!

Ambrevar:
Merged!

gagbo:
Thanks !

PR #430, opened 2021-01-01
xref: Add bindings for xref--transient-buffer-mode

condy0919:


modes/xref/evil-collection-xref.el
```diff
@@ -50,13 +52,24 @@
...
     "go" 'xref-show-location-at-point)
 
   (when (>= emacs-major-version 27)
     (evil-collection-define-key 'normal 'xref--xref-buffer-mode-map
-      "gr" 'xref-revert-buffer)))
+      "gr" 'xref-revert-buffer))
+
+  (when (>= emacs-major-version 28)
+    (evil-collection-define-key 'normal 'xref--xref-buffer-mode-map
+      "[[" 'xref-prev-group
```
jojojames, line 66: I wonder if this conflicts with unimpaired style bindings.

modes/xref/evil-collection-xref.el
```diff
@@ -50,13 +52,24 @@
 
     ;; open
     (kbd "RET") 'xref-goto-xref
+    (kbd "TAB") 'xref-quit-and-goto-xref
```
jojojames, line 55: This one might be unexpected by a user. Maybe another key?

condy0919: That's the default Emacs style binding:

``` elisp
(defvar xref--xref-buffer-mode-map
    (define-key map (kbd "n") #'xref-next-line)
    (define-key map (kbd "p") #'xref-prev-line)
    (define-key map (kbd "N") #'xref-next-group)
    (define-key map (kbd "P") #'xref-prev-group)
    (define-key map (kbd "r") #'xref-query-replace-in-results)
    (define-key map (kbd "RET") #'xref-goto-xref)
    (define-key map (kbd "TAB")  #'xref-quit-and-goto-xref)
    (define-key map (kbd "C-o") #'xref-show-location-at-point)
    ;; suggested by Johan Claesson "to further reduce finger movement":
    (define-key map (kbd ".") #'xref-next-line)
    (define-key map (kbd ",") #'xref-prev-line)
    (define-key map (kbd "g") #'xref-revert-buffer)
    map))
```

Agree that it's confused, what about changing to `S-RET`?
`xref-quit-and-goto-xref` has already two bindings (`o` and `go`) to invoke.

modes/xref/evil-collection-xref.el
condy0919: As far as I known, unimpaired won't/shouldn't use `[[` and `]]`. Those are used in vim motion state.

modes/xref/evil-collection-xref.el
jojojames: Maybe it's OK, I'm not sure if we have precedence yet for "action -> close current window".

condy0919:
@jojojames ping

PR #435, opened 2021-01-05
elisp-mode: Binds gz to ielm

condy0919:
`evil-collection-elisp-mode-ielm-repl` will open a new window while `ielm` use current window. Since it's a REPL, the former is more intuitional


jojojames:
*musing* I almost never use IELM, I wonder how many people use it over editing inline/straight in emacs.

PR #439, opened 2021-01-10
elfeed: Bind q, ZQ, ZZ to elfeed-search-quit-window

mssdvd:
elfeed binds `q` to `elfeed-search-quit-window` which saves the db and
quit the window. As today we only close the window which may lead to
data loss.


condy0919:
Thanks!

PR #441, opened 2021-01-12
Sly: Switch advice to `sly-last-expression`

jojojames:
sly-eval-print-last-expression is implemented like this:
```
(defun sly-eval-print-last-expression (string)
  "Evaluate sexp before point; print value into the current buffer"
  (interactive (list (sly-last-expression)))
  (insert "\n")
  (sly-eval-print string))
```

Since the arguments are already populated by the time the advice happens,
evaluating something like this:

(+ 1 1[)] ; bracket wrapping cursor representing cursor position.

will evalute to 1 when we expect (+ 1 1) to be evaluated. We advise
`sly-last-expression` with a before advice to move the cursor forward one before
grabbing the sexp.

@nnoodle I see you are a sly connoisseur, if you have some time can you see if this works for you and the solution is palatable (e.g. wouldn't give you a headache to see that -sly-last-expression- advised).

If we don't advise this function, another option is to re-implement the eval functions to account for evil.

@condy0919 PTAL

### DIFF
--- a/modes/sly/evil-collection-sly.el
+++ b/modes/sly/evil-collection-sly.el
@@ -32,6 +32,9 @@
 (require 'evil-collection)
 (require 'sly nil t)
 
+(declare-function sly-last-expression "sly")
+(declare-function sly-eval-print "sly")
+
 (defvar sly-connection-list-mode-map)
 (defvar sly-db-mode-map)
 (defvar sly-inspector-mode-map)
@@ -62,14 +65,29 @@
         (apply command args))
     (apply command args)))
 
+(defun evil-collection-sly-eval-print-last-expression (string)
+  "Evaluate sexp before point; print value into the current buffer.
+
+Evil version of `sly-eval-print-last-expression' that accounts for
+`evil-move-beyond-eol'."
+  (interactive
+   (list (progn
+           (when (and (not evil-move-beyond-eol)
+                      (or (evil-normal-state-p) (evil-motion-state-p)))
+             (unless (or (eobp) (eolp))
+               (forward-char)))
+           (sly-last-expression))))
+  (insert "\n")
+  (sly-eval-print string))
+
 ;;;###autoload
 (defun evil-collection-sly-setup ()
   "Set up `evil' bindings for `sly'."
   (unless evil-move-beyond-eol
     (advice-add 'sly-eval-last-expression :around 'evil-collection-sly-last-sexp)
     (advice-add 'sly-pprint-eval-last-expression :around 'evil-collection-sly-last-sexp)
-    (advice-add 'sly-eval-print-last-expression :around 'evil-collection-sly-last-sexp)
-    (advice-add 'sly-mrepl-return :around 'evil-collection-sly-last-sexp))
+    (advice-add 'sly-mrepl-return :around 'evil-collection-sly-last-sexp)
+    (advice-add 'sly-eval-print-last-expression :override 'evil-collection-sly-eval-print-last-expression))
 
   (evil-collection-define-key 'normal 'sly-db-mode-map
     [follow-link] 'mouse-face


nnoodle:
With mostly just the essentials loaded, evaluating `(+ 1 1`[`)`] with the cursor over the closing parenthesis using `sly-eval-last-expression` properly evaluates to 2, as well as when I had `sly-last-expression` advised with `evil-collection-sly-last-sexp` around instead.

An issue that does come up however is if I evaluate `(+ 1 2`[`)`] using `sly-eval-print-last-sexp` when it's advised around `evil-collection-sly-last-sexp`, it incorrectly prints and transforms to:
```lisp
(+ 1 1 ; => 1 (1 bit, #x1, #o1, #b1)
   )
```
I think it should be kept as is, except change `sly-eval-print-last-sexp`'s advice to `evil-collection-sly-move-cursor-forward`.

jojojames:
> An issue that does come up however is if I evaluate (+ 1 2[)] using sly-eval-print-last-sexp when it's advised around evil-collection-sly-last-sexp, it incorrectly prints and transforms to:

Hunh, that's what this change is supposed to fix and what I am seeing without this PR. I wonder if you had any extra advices stuck to those functions from previous emacs/evil-collection state.

This is `sly-eval-print-last-sexp''s implementation.

```
(defun sly-eval-print-last-expression (string)
  "Evaluate sexp before point; print value into the current buffer"
  (interactive (list (sly-last-expression)))
  (insert "\n")
  (sly-eval-print string))
```

Trying to form the flowchart in my head, if you add a before advice here to `sly-last-expression`, it should move the cursor forward and get the right sexp but if you add the advice to `sly-eval-print-last-expression`, it'll get the wrong STRING argument.

> I think it should be kept as is, except change sly-eval-print-last-sexp's advice to evil-collection-sly-move-cursor-forward.

Right, I'm expecting that not to work given the way advices are implemented. Both the advice and 'old function'/advised function take in the same arguments to their functions which means by the time old-function is called, even if we had moved the cursor forward, will still see the same STRING argument as if we didn't move the cursor.

Maybe I'm wrong on some details so hopefully more people can vouch for/against this change before I merge it.

condy0919:
Maybe we should fix upstream `sly-eval-print-last-expression` rather than advising?

``` elisp
;; I think it's similar to `sly-eval-print-last-expression'
(defun eval-to-comment (&optional arg)
    (interactive "P")
    (let ((start (point)))
      (eval-print-last-sexp arg)
      (save-excursion
        (goto-char start)
        (save-match-data
          (re-search-forward "[[:space:]\n]+" nil t)
          (insert ";;=> "))))))
```

jojojames:
> Maybe we should fix upstream sly-eval-print-last-expression rather than advising?

There's no real problem with the function. Sure, it can be written in a way to account for evil's idiosyncrasies but I don't expect every external package from evil to do that, so we should be comfortable with advice when it works. Note also, we've been using advice for these types of integrations since even evil-integration.el.

https://github.com/emacs-evil/evil/blob/cc9d6886b418389752a0591b9fcb270e83234cf9/evil-integration.el#L232

re: eval-to-comment 

The sly-eval functions seemingly do a little more than that:

```
(defun sly-eval-print (string)
  "Eval STRING in Lisp; insert any output and the result at point."
  (sly-eval-async `(slynk:eval-and-grab-output ,string)
    (lambda (result)
      (cl-destructuring-bind (output value) result
        (push-mark)
        (let* ((start (point))
               (ppss (syntax-ppss))
               (string-or-comment-p (or (nth 3 ppss) (nth 4 ppss))))
          (insert output (if string-or-comment-p
                             ""
                           " => ") value)
          (unless string-or-comment-p
            (comment-region start (point) 1)))))))
```

nnoodle:
>Hunh, that's what this change is supposed to fix and what I am seeing without this PR.

My bad, I meant that yeah I see that issue without the PR, and the PR does fix that specific problem, but I didn't see any problems with any of the other sly-*-last-sexp functions without the PR.

The other two functions that use `sly-last-expression` should be advised to return the cursor back to it's initial place.
```emacs-lisp
(defun evil-collection-sly-move-cursor-backward (&rest _args)
  "Move cursor backward once so for `sly-last-expression'."
  (when (and (not evil-move-beyond-eol)
             (or (evil-normal-state-p) (evil-motion-state-p)))
    (unless (or (bobp) (bolp))
      (backward-char))))
(advice-add 'sly-eval-last-expression :after 'evil-collection-sly-move-cursor-backward)
(advice-add 'sly-pprint-eval-last-expression :after 'evil-collection-sly-move-cursor-backward)
```
It isn't wrapped around save-excursion so it won't return back if there's an error.

>Right, I'm expecting that not to work given the way advices are implemented.

Ah, yeah my bad I see it now, I just tried it and advising around it didn't work. Other than advising, it could also be overridden since it's the only function that needs to be fixed for evil.
```emacs-lisp
(define-advice sly-eval-print-last-expression (:override (string) evil-collection-sly)
  "Evaluate sexp before point; print value into the current buffer"
  (interactive
   (list (progn
           (when (and (not evil-move-beyond-eol)
                      (or (evil-normal-state-p) (evil-motion-state-p)))
             (unless (or (eobp) (eolp))
               (forward-char)))
           (sly-last-expression))))
  (insert "\n")
  (sly-eval-print string))
```

jojojames:
> but I didn't see any problems with any of the other sly-*-last-sexp functions without the PR.

Right yeah, those are fine, but they use the same function to find the sexp so would be affected with this PR.

> Other than advising, it could also be overridden since it's the only function that needs to be fixed for evil.

Yeah, I agree, that's probably the better approach, reworked the advice similar to what you wrote. PTAL. @nnoodle

nnoodle:
LGTM @jojojames

PR #442, opened 2021-01-15
add kbd for magit-log-select-mode

tshu-w:
before:
Screen Shot 2021-01-15 at 2 24 22 PM
after:
Screen Shot 2021-01-15 at 2 23 21 PM

BTW, I'm not quite sure whether we should use `p` or `RET` to `magit-log-select-pick`

### DIFF
--- a/modes/sly/evil-collection-sly.el
+++ b/modes/sly/evil-collection-sly.el
@@ -32,6 +32,9 @@
 (require 'evil-collection)
 (require 'sly nil t)
 
+(declare-function sly-last-expression "sly")
+(declare-function sly-eval-print "sly")
+
 (defvar sly-connection-list-mode-map)
 (defvar sly-db-mode-map)
 (defvar sly-inspector-mode-map)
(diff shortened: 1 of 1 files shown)


jojojames:
What does magit-log-select-pick do? I'm guessing from the name RET would be appropriate.

tshu-w:
> What does magit-log-select-pick do? I'm guessing from the name RET would be appropriate.

For example, if I want to modify a commit, I can type `r`(rebase) `m`(modify a commit) and move the cursor to the commit I want to select, and then run `magit-log-select-pick`

condy0919:
If I remember correct, the default binding is <kbd>C-c C-c</kbd> to
fixup, <kbd>C-c C-k</kbd> to abort

tshu-w:
> If I remember correct, the default binding is C-c C-c to
> fixup, C-c C-k to abort

Yes, but also `e` and `q` if I don't use evil-collection. evil-collection change `magit-log-select-mode ` init mode to normal and somehow make `magit-log-select-quit` shortcut hint to stop appearing.

jojojames:
> evil-collection change magit-log-select-mode init mode to normal

Do you remember if that was true for evil-magit as well? Maybe there was a bug during the transition.

tshu-w:
@jojojames I don't really remember, I used to use spacemacs and had its own set of settings. But now I have the same problem with evil-magit.

tshu-w:
I close this pull request and create issue #444, once that was confirmed, I was happy to submit another pull request.

tshu-w:
> If I remember correct, the default binding is C-c C-c to
> fixup, C-c C-k to abort

You are right, default binding was not shown because of sis https://github.com/laishulu/emacs-smart-input-source/issues/28, so shall we configure more convenient kbd?

PR #447, opened 2021-01-23
consult-evil-mark

tpeacock19:
Jump to evil markers in current buffer

### DIFF
--- a/modes/consult/evil-collection-consult.el
+++ b/modes/consult/evil-collection-consult.el
@@ -30,11 +30,61 @@
 (require 'evil-collection)
 (require 'consult nil t)
 
+(defvar consult-line-numbers-widen)
+(declare-function consult--forbid-minibuffer "consult")
+(declare-function consult--fontify-all "consult")
+(declare-function consult--in-range-p "consult")
+(declare-function consult--line-with-cursor "consult")
+(declare-function consult--remove-dups "consult")
+(declare-function consult--add-line-number "consult")
+(declare-function consult--mark-candidates "consult")
+(declare-function consult-mark "consult")
+
 (defun evil-collection-consult-set-bindings ()
   "Set the bindings."
   (evil-set-command-property 'consult-outline :jump t)
   (evil-set-command-property 'consult-mark :jump t)
   (evil-set-command-property 'consult-line :jump t))
+
+(defun evil-collection-consult-evil-mark-ring ()
+  "Return alist of char & marker for evil markers in current buffer."
+  (sort (cl-remove-if (lambda (elem)
+                        (or (not (evil-global-marker-p (car elem)))
+                            (not (markerp (cdr-safe elem)))))
+                      evil-markers-alist)
+        #'car-less-than-car))
+
+(defun evil-collection-consult--mark-candidates ()
+  "Return alist of lines containing markers.
+The alist contains (string . position) pairs."
+  (consult--forbid-minibuffer)
+  (unless (evil-collection-consult-evil-mark-ring)
+    (user-error "No marks"))
+  (consult--fontify-all)
+  (let* ((max-line 0)
+         (candidates))
+    (save-excursion
+      (dolist (marker (evil-collection-consult-evil-mark-ring))
+        (let ((pos (marker-position (cdr marker))))
+          (when (consult--in-range-p pos)
+            (goto-char pos)
+            (let ((line (line-number-at-pos pos consult-line-numbers-widen)))
+              (setq max-line (max line max-line))
+              (push (list (cdr marker) line (format "%s: %s" (char-to-string (car marker))
+                                                    (consult--line-with-cursor (cdr marker))))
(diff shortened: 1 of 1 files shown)


modes/consult/evil-collection-consult.el
```diff
@@ -36,6 +36,45 @@
   (evil-set-command-property 'consult-mark :jump t)
   (evil-set-command-property 'consult-line :jump t))
 
+(defun evil-mark-ring ()
```
jojojames, line 39: This functionality seems pretty cool, I think we'd need to prefix these with evil-collection-consult though as to not pollute the global namespace which begs the question of where this functionality belongs.

tpeacock19: Right, I don't think the maintainers of consult want to include any evil configuration in their repo, which makes sense. So I figured this would be the best place to put it. However, I understand if you want to restrict it to just keybindings.

jojojames: Right, I was just thinking the usefulness of this function is partly also thanks to a short name, once you add the -collection- part, it becomes quite verbose as an interactive function. 

I'm guessing it's not a problem for you since you probably have the function bound somewhere else. If the maintainers don't want this, this should be an OK spot, I think.

jojojames:
OK to merge after fixing CI: e.g.

```
modes/consult/evil-collection-consult.el:63:49:Error: reference to free variable ‘consult-line-numbers-widen’
make: *** [Makefile:11: compile] Error 1
```

Will need to use a combo of defvar/declare-function to clear up the compile errors.

tpeacock19:
> OK to merge after fixing CI: e.g.

I'm unsure how to fix these as I'm not seeing them when I compile. Do I need to use a different version of emacs?

jojojames:
It's because you're probably compiling within your live emacs setup which already has the symbols defined, if you try to compile this file from a blank slate, it'll throw errors/warnings. You can try adding the defvar/declare-function symbols mentioned in the CI and then pushing the branch and checking if the CI passes or set up Cask and rerun the tests (check Makefile) from a terminal.

tpeacock19:
Okay, I understand. I am at the point where I believe the last change is to rename the current `evil-collection-mark-ring` function to `evil-collection-consult-mark-ring`. Perhaps its better to place this in `evil-collection.el` to keep it less verbose?

modes/consult/evil-collection-consult.el
```diff
@@ -36,6 +36,55 @@
...
   (evil-set-command-property 'consult-line :jump t))
 
+(defun evil-collection-evil-mark-ring ()
+  "Return alist of char & marker for evil markers."
+  (sort (cl-remove-if (lambda (elem)
+                        (not (markerp (cdr-safe elem))))
+                      evil-markers-alist)
+        #'car-less-than-car))
+
+(defvar consult-line-numbers-widen)
```
jojojames, line 46: Move these to the top of the file.

modes/consult/evil-collection-consult.el
```diff
@@ -36,6 +36,55 @@
...
+      (dolist (marker (evil-collection-evil-mark-ring))
+        (let ((pos (marker-position (cdr marker))))
+          (when (consult--in-range-p pos)
+            (goto-char pos)
+            ;; `line-number-at-pos' is a very slow function, which should be replaced everywhere.
+            ;; However in this case the slow line-number-at-pos does not hurt much, since
+            ;; the mark ring is usually small since it is limited by `mark-ring-max'.
+            (let ((line (line-number-at-pos pos consult-line-numbers-widen)))
+              (setq max-line (max line max-line))
+              (push (list (cdr marker) line (format "%s: %s" (char-to-string (car marker)) (consult--line-with-cursor (cdr marker))))
```
jojojames, line 75: Less than 80 columns formatting

jojojames:
> Perhaps its better to place this in evil-collection.el to keep it less verbose?
Probably not, it's not an interactive function and is only used here so far.

tpeacock19:
Okay I think that should cover it. I noticed there is `evil-global-marker-p` that could be used to emulate `consult-global-marks`, but I don't know if I've used any global marks in evil before anyway.

modes/consult/evil-collection-consult.el
```diff
@@ -30,12 +30,62 @@
...
+(declare-function consult--mark-candidates "consult")
+(declare-function consult-mark "consult")
+
 (defun evil-collection-consult-set-bindings ()
   "Set the bindings."
   (evil-set-command-property 'consult-outline :jump t)
   (evil-set-command-property 'consult-mark :jump t)
   (evil-set-command-property 'consult-line :jump t))
 
+(defun evil-collection-consult-evil-mark-ring ()
```
jojojames, line 49: Should this be internal too? e.g. evil-collection-consult--evil-mark-ring?

jojojames:
Merging, feel free to address comments in a separate PR.

PR #448, opened 2021-01-24
Add additional evil libvterm keybindings

mjlbach:
Draft, still working (and taking requests) on what to add.

@jixiuf All the bindings that depend on delete region don't seem to be working for me. Delete-region itself works, I'm not trying to delete anything outside the editable portion of the current prompt to my knowledge.

### DIFF
--- a/modes/vterm/evil-collection-vterm.el
+++ b/modes/vterm/evil-collection-vterm.el
@@ -33,6 +33,11 @@
 
 (defconst evil-collection-vterm-maps '(vterm-mode-map))
 
+(declare-function vterm-goto-char "vterm")
+(declare-function vterm--get-prompt-point "vterm")
+(declare-function vterm--get-end-of-line "vterm")
+(declare-function vterm-delete-region "vterm")
+
 (defvar vterm--process)
 
 (defun evil-collection-vterm-escape-stay ()
@@ -61,6 +66,106 @@ also uses `evil-mode'."
                    (if evil-collection-vterm-send-escape-to-vterm-p
                        "vterm"
                      "emacs"))))
+
+(defun evil-collection-vterm-insert ()
+  "Insert character before cursor."
+  (interactive)
+  (vterm-goto-char (point))
+  (call-interactively #'evil-insert))
+
+(defun evil-collection-vterm-insert-line ()
+  "Insert character at beginning of prompt."
+  (interactive)
+  (vterm-goto-char (vterm--get-prompt-point))
+  (call-interactively #'evil-insert))
+
+(defun evil-collection-vterm-append ()
+  "Append character after cursor."
+  (interactive)
+  (vterm-goto-char (1+ (point)))
+  (call-interactively #'evil-append))
+
+(defun evil-collection-vterm-append-line ()
+  "Append character at end-of-line."
+  (interactive)
+  (vterm-goto-char (vterm--get-end-of-line))
+  (call-interactively #'evil-append))
+
+(evil-define-operator evil-collection-vterm-delete (beg end type register yank-handler)
+  "Modification of evil-delete to work in vterm buffer. 
+Delete text from BEG to END with TYPE.
+Save in REGISTER or in the kill-ring with YANK-HANDLER."
+  (interactive "<R><x><y>")
+  (let* ((beg (max (or beg (point)) (vterm--get-prompt-point)))
+         (end (min (or end beg) (vterm--get-end-of-line))))
+    (unless register
+      (let ((text (filter-buffer-substring beg end)))
+        (unless (string-match-p "\n" text)
+          ;; set the small delete register
+          (evil-set-register ?- text))))
+    (let ((evil-was-yanked-without-register nil))
+      (evil-yank beg end type register yank-handler))
+    (cond
+     ((eq type 'block)
+      (evil-apply-on-block #'vterm-delete-region beg end nil))
+     ((and (eq type 'line)
+           (= end (point-max))
+           (or (= beg end)
+               (/= (char-before end) ?\n))
+           (/= beg (point-min))
+           (=  (char-before beg) ?\n))
+      (vterm-delete-region (1- beg) end))
+     (t
+      (vterm-delete-region beg end)))
+    ;; place cursor on beginning of line
+    (when (and (called-interactively-p 'any)
+               (eq type 'line))
+      (evil-first-non-blank))))
+
+(evil-define-operator evil-collection-vterm-delete-backward-char (beg end type register)
+  "Delete previous character."
+  :motion evil-backward-char
+  (interactive "<R><x>")
+  (evil-collection-vterm-delete beg end type register))
+
+(evil-define-operator evil-collection-vterm-delete-line (beg end type register yank-handler)
+  "Modification of evil-delete line to work in vterm bufer. Delete to end of line."
+  :motion nil
+  :keep-visual t
+  (interactive "<R><x>")
+  ;; act linewise in Visual state
+  (let* ((beg (or beg (point)))
+         (end (or end beg))
+         (visual-line-mode (and evil-respect-visual-line-mode
+                                visual-line-mode))
+         (line-end (if visual-line-mode
+                       (save-excursion
+                         (end-of-visual-line)
+                         (point))
+                     (line-end-position))))
+    (when (evil-visual-state-p)
+      (unless (memq type '(line screen-line block))
+        (let ((range (evil-expand beg end
+                                  (if visual-line-mode
+                                      'screen-line
+                                    'line))))
+          (setq beg (evil-range-beginning range)
+                end (evil-range-end range)
+                type (evil-type range))))
+      (evil-exit-visual-state))
+    (cond
+     ((eq type 'block)
+      ;; equivalent to $d, i.e., we use the block-to-eol selection and
+      ;; call `evil-collection-vterm-delete'. In this case we fake the call to
+      ;; `evil-end-of-line' by setting `temporary-goal-column' and
+      ;; `last-command' appropriately as `evil-end-of-line' would do.
+      (let ((temporary-goal-column most-positive-fixnum)
+            (last-command 'next-line))
+        (evil-collection-vterm-delete beg end 'block register yank-handler)))
+     ((memq type '(line screen-line))
+      (evil-collection-vterm-delete beg end type register yank-handler))
+     (t
+      (evil-collection-vterm-delete beg line-end type register yank-handler)))))
 
 ;;;###autoload
 (defun evil-collection-vterm-setup ()
@@ -101,7 +206,19 @@ also uses `evil-mode'."
     "[[" 'vterm-previous-prompt
     "]]" 'vterm-next-prompt
     "p" 'vterm-yank
-    "u" 'vterm-undo))
+    "a" 'evil-collection-vterm-append
+    "A" 'evil-collection-vterm-append-line
+    "d" 'evil-collection-vterm-delete
+    "D" 'evil-collection-vterm-delete-line
+    "x" 'evil-collection-vterm-delete-backward-char
+    (kbd "RET") 'vterm-send-return
+    "i" 'evil-collection-vterm-insert
+    "I" 'evil-collection-vterm-insert-line
+    "u" 'vterm-undo)
+
+  (evil-collection-define-key 'visual 'vterm-mode-map
+    "d" 'evil-collection-vterm-delete
+    "x" 'evil-collection-vterm-delete-backward-char))
 
 (provide 'evil-collection-vterm)
 ;;; evil-collection-vterm.el ends here


modes/vterm/evil-collection-vterm.el
```diff
@@ -62,6 +62,58 @@ also uses `evil-mode'."
                        "vterm"
                      "emacs"))))
 
+(defun vterm-evil-insert ()
```
jojojames, line 65: We'll want the prefix to go from vterm-evil- to evil-collection-vterm-.

jojojames, line 65: Nit: Might be good to have some comments on what this does. I'm guessing this gets us into insert mode at the actual point the cursor is on in the terminal instead of at the end of the line.

modes/vterm/evil-collection-vterm.el
```diff
@@ -62,6 +62,58 @@ also uses `evil-mode'."
...
+  (interactive)
+  (vterm-goto-char (point))
+  (call-interactively #'evil-insert))
+
+(defun vterm-evil-insert-line ()
```
jojojames, line 70: Having a hard time grokking what this does (or what someone would use it for).

modes/vterm/evil-collection-vterm.el
```diff
@@ -62,6 +62,58 @@ also uses `evil-mode'."
...
+(defun vterm-evil-append-line ()
+  (interactive)
+  (vterm-goto-char (vterm--get-end-of-line))
+  (call-interactively #'evil-append))
+
+(defun vterm-evil-delete ()
+  "Provide similar behavior as `evil-delete'."
+  (interactive)
+  (let ((inhibit-read-only t)
+        )
```
jojojames, line 89: Move ) up to match prior (

modes/vterm/evil-collection-vterm.el
```diff
@@ -62,6 +62,58 @@ also uses `evil-mode'."
...
+    (cl-letf (((symbol-function #'delete-region) #'vterm-delete-region))
```
jojojames, line 90: Wondering what the purpose of this let is?

modes/vterm/evil-collection-vterm.el
```diff
@@ -62,6 +62,58 @@ also uses `evil-mode'."
...
+      (call-interactively 'evil-delete))))
+
+(defun vterm-evil-delete-line ()
+  "Provide similar behavior as `evil-delete'."
+  (interactive)
+  (let ((inhibit-read-only t)
+        )
```
jojojames, line 97: Same here, move up to prior line

modes/vterm/evil-collection-vterm.el
```diff
@@ -62,6 +62,58 @@ also uses `evil-mode'."
...
+(defun vterm-evil-delete ()
+  "Provide similar behavior as `evil-delete'."
+  (interactive)
+  (let ((inhibit-read-only t)
+        )
+    (cl-letf (((symbol-function #'delete-region) #'vterm-delete-region))
+      (call-interactively 'evil-delete))))
+
+(defun vterm-evil-delete-line ()
+  "Provide similar behavior as `evil-delete'."
```
jojojames, line 94: s/evil-delete/evil-delete-line

modes/vterm/evil-collection-vterm.el
```diff
@@ -101,7 +153,22 @@ also uses `evil-mode'."
...
+    "i" 'vterm-evil-insert
+    "I" 'vterm-evil-insert-line
+    ;; "c" 'vterm-evil-change
+    "u" 'vterm-undo)
+
+  (evil-collection-define-key 'visual 'vterm-mode-map
+    ;; "d" 'vterm-evil-delete
+    ;; "x" 'vterm-evil-delete
+    ;; "c" 'vterm-evil-change
+    ;; "c" 'vterm-evil-replace
```
jojojames, line 171: c->r?

modes/vterm/evil-collection-vterm.el
```diff
@@ -101,7 +153,22 @@ also uses `evil-mode'."
...
+    ;; "D" 'vterm-evil-delete-line
+    ;; "x" 'vterm-evil-delete-char
+    (kbd "RET") 'vterm-send-return
+    "i" 'vterm-evil-insert
+    "I" 'vterm-evil-insert-line
+    ;; "c" 'vterm-evil-change
+    "u" 'vterm-undo)
+
+  (evil-collection-define-key 'visual 'vterm-mode-map
+    ;; "d" 'vterm-evil-delete
```
jojojames, line 168: These look okay to me.

modes/vterm/evil-collection-vterm.el
```diff
@@ -101,7 +153,22 @@ also uses `evil-mode'."
     "[[" 'vterm-previous-prompt
     "]]" 'vterm-next-prompt
     "p" 'vterm-yank
-    "u" 'vterm-undo))
-
+    "a" 'vterm-evil-append
+    "A" 'vterm-evil-append-line
+    ;; "d" 'vterm-evil-delete
```
jojojames, line 158: Would be cool if these vterm functions words, etc e.g. dw, dW etc etc, but that's probably out of scope.

These binds look ok to me.

modes/vterm/evil-collection-vterm.el
mjlbach: This is the equivalent to the I motion map in vim for going back to the beginning of the (editable) portion of the prompt. It's a little finnicky due to prompt detection with vterm (working on it), but I use it for changing the first command that I've entered.

modes/vterm/evil-collection-vterm.el
mjlbach: Just to inhibit the read only buffer so that delete-region can work. I took it from the PR adding the functions to vterm here: https://github.com/akermu/emacs-libvterm/pull/455

I'm having trouble getting this one to work, it works in edebug, but strangely not when called/mapped during normal usage.

modes/vterm/evil-collection-vterm.el
mjlbach: Yes! I'll add comments soon.

modes/vterm/evil-collection-vterm.el
jixiuf: maybe the `let` is not necessary , I tried it still words without `let`

jixiuf:
> @jixiuf All the bindings that depend on delete region don't seem to be working for me. Delete-region itself works, I'm not trying to delete anything outside the editable portion of the current prompt to my knowledge.

I can't reproduce this, Can you share your .bashrc/.zshrc or build a docker image to reproduce the bug.

mjlbach:
> > @jixiuf All the bindings that depend on delete region don't seem to be working for me. Delete-region itself works, I'm not trying to delete anything outside the editable portion of the current prompt to my knowledge.
> 
> I can't reproduce this, Can you share your .bashrc/.zshrc or build a docker image to reproduce the bug.

I'm having a little bit of trouble getting the prompt detection working on bash in the minimal example. Here it is: 

https://github.com/mjlbach/emacs-container

jixiuf:
> > > @jixiuf All the bindings that depend on delete region don't seem to be working for me. Delete-region itself works, I'm not trying to delete anything outside the editable portion of the current prompt to my knowledge.
> > 
> > 
> > I can't reproduce this, Can you share your .bashrc/.zshrc or build a docker image to reproduce the bug.
> 
> I'm having a little bit of trouble getting the prompt detection working on bash in the minimal example. Here it is:
> 
> https://github.com/mjlbach/emacs-container

Got 

```
docker build -t vterm .  
docker run -t -i vterm   

Warning: arch-dependent data dir '/usr/lib/emacs/27.1/x86_64-alpine-linux-musl/': Operation not permitted
Warning: arch-independent data dir '/usr/share/emacs/27.1/etc/': Operation not permitted
Warning: game dir '/var/games/emacs': Operation not permitted
Warning: Lisp directory '/usr/share/emacs/27.1/lisp': Operation not permitted
Error: /usr/share/emacs/27.1/etc/charsets: Operation not permitted
Emacs will not function correctly without the character map files.
Please check your installation!
```

mjlbach:
@jixiuf Sorry about that, it's due to `build-base` in alpine, not sure why it's causing an issue only on docker (and not podman) though.

I made a fedora branch of the repo that I sent you here: https://github.com/mjlbach/emacs-container/tree/fedora

I tested and confirmed it works with docker.

jixiuf:
I can reproduce with 
```
(defvar bootstrap-version)
(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

(straight-use-package 'use-package)

(setq evil-want-keybinding nil)
(setq vterm-always-compile-module t)

;; (add-to-list 'load-path "~/.emacs.d/straight/repos/evil")
;; (evil-mode)
(use-package evil
 :straight t
 :ensure t
 :config
 (evil-mode 1))

(use-package evil-collection
 :straight (:host github :repo "mjlbach/evil-collection"
            :no-byte-compile t
            :branch "feature/additional_libvterm_evil_bindings")
 :config
 (evil-collection-init)
 )

(use-package vterm
 :straight t
 :config
 )
``` 

But if I change 
```
(use-package evil
 :straight t
 :ensure t
 :config
 (evil-mode 1))

```
 to 
```
(add-to-list 'load-path "~/.emacs.d/straight/repos/evil")
(evil-mode)

```
it would works, I don't known why  right now.

mjlbach:
Making those changes to init.el didn't seem to fix anything for me. I added a zshrc and updated the Dockerfile to include hostname, still getting very odd behavior with prompt detection.

mjlbach:
What's weird is vterm-evil-delete seems to work fine when invoked manually

mjlbach:
This is really weird, the delete-region works now but only if I manually defined evil-delete-region before importing evil -collection (I didn't have to modify load-path) see:  https://github.com/mjlbach/emacs-container/tree/works

jixiuf:
init.el
```
;;
;;
(add-to-list 'load-path "~/.emacs.d/")

(defun my-evil-delete ()
  "Provide similar behavior as `evil-delete'."
  (interactive)
  (cl-letf (((symbol-function #'delete-region) #'(lambda (s e) (message "cl-letf works" ))))
    (call-interactively #'mock-evil-delete)))

(global-set-key (kbd "C-d") 'my-evil-delete)
(require 'my-evil)
```

my-evil.el
```
(defun mock-evil-delete()
  (interactive )
  (delete-region 1 2 )
  )

(provide 'my-evil)
```

When my-evil.el is not byte compiled , "cl-letf works" show in mini buffer,
But when my-evil.el is byte compiled , "cl-letf works" does not show in mini buffer


see https://www.gnu.org/software/emacs/manual/html_node/cl/Obsolete-Macros.html
maybe we should use advice instead of `cl-letf`

mjlbach:
I see advice has the same warning about byte-compilation as flet/cl-letf

```
Note that many primitives (e.g., +) have special byte-compile handling. Attempts to redefine such functions using flet, cl-letf, or advice will fail when byte-compiled.
```

mjlbach:
I'm just going to duplicate and substitute the requeste evil maps for now

mjlbach:
Everything works now, but I haven't added change and zsh is still finnicky (upstream issue though)

mjlbach:
Also `dd` isn't mapped, not sure how to do that one.

modes/vterm/evil-collection-vterm.el
mjlbach: Resolving because I removed this

modes/vterm/evil-collection-vterm.el
mjlbach: Added documentation :)

modes/vterm/evil-collection-vterm.el
mjlbach: I added some comments. Let me know if these aren't sufficient!

mjlbach:
@jojojames If you let me know how to fix the CI failures, I'm read for re-review :)

jojojames:
@mjlbach You can use declare-function, https://github.com/emacs-evil/evil-collection/blob/323bb7d85848a6a142ae14f39c3a073ce6423e20/modes/scroll-lock/evil-collection-scroll-lock.el#L32

mjlbach:
Good on my end to merge :)

jojojames:
Thanks, looks awesome to me!

jixiuf:
> Of note, zsh inserts a ~ on each deletion, see `[root@5eeb5f080a4d]/# echo te~~~~st`

`(vterm-send-delete)`. insert `~` in zsh does not delete char ,Maybe I should call `vterm-send-C-d` in `vterm-delete-region`

mjlbach:
It seems to work fine in my normal zsh prompt, not sure what happened in the test container

mjlbach:
dd is a little finnicky (I have a multi-line prompt)

jixiuf:
> It seems to work fine in my normal zsh prompt, not sure what happened in the test container

maybe you bind delete in your zshrc
```
 bindkey|grep delete
```

mjlbach:
Ah yeah, good catch 
```
❯ bindkey|grep delete
"^D" delete-char-or-list
"^H" backward-delete-char
"^[[3~" delete-char
"^?" backward-delete-char
```

cobac:
Hey, thanks everyone for your work.

Just a question out of ignorance.

Will this implementation work with other interpreters instead of shells? I'm thinking about interactive python/R/Julia sessions inside vterm.

arunkmv:
Hi, I have a suggestion based on a binding I use with vi emulation in zsh. Currently the j and k keys in normal mode just go up in the terminal buffer which may not be of much use most of the time. A more intuitive use of these keys could be to scroll up and down the shell history.

```
(evil-collection-define-key 'normal 'vterm-mode-map 
  "k" 'vterm-send-up
  "j" 'vterm-send-down)
```

This may be undesirable behaviour for some, but I've certainly found it useful.

mjlbach:
I frequently use j/k to navigate in normal mode in the terminal buffer in order to yank previous commands/output, so this is not a default I would support.

jojojames:
I'd agree with @mjlbach here, one reason to use emulators within something like emacs is the better output navigation support compared to traditional terminals and I think the number of times I'd want to scroll up or down a buffer is much higher than going through shell history.

arunkmv:
I guess I'm in the minority here, maybe my workflow could use some tweaking. Thanks for the feedback, Cheers!

BlueDrink9:
Was `cc`/`dd` ever implemented successfully? I'm using doom and these pairs are refusing to modify because of the read-only buffer

PR #452, opened 2021-01-27
fix: Stop inhibiting company popup when not in evil-mode

### LINKED ISSUE #451: Inhibiting company completions
I have a tiny problem while using this package. Whenever installed, it to inhibits company completions and I get the message `No completion found` whenever evil-mode is disabled. Once this package is deleted and I switch between evil-mode and normal emacs then company-mode works fine.

Here's my current configuration:

```emacs-lisp
(setq evil-want-keybinding nil)

(use-package evil
  :bind ("<C-tab>" . 'evil-mode))
  
  (use-package evil-collection
    :after evil
    :config
    (evil-collection-init))


```

I know it is not a critical bug, but it's super annoying given the fact that I switch between evil and normal emacs quite often. Anyone have any idea where I am going wrong?

Randy1Burrell:
The `evil-collection` package inhibits `company-mode` from showing completions when evil-mode
is switched off. This commit will fix it and only activate `evil-collection` advice for `company-mode`
when in `evil-mode`.

If applied, this commit will:
    * Allow `company-mode` to show completions when not in `evil-mode`.
    * Only activate `evil-collection` advice for `company-mode` when in `evil-mode`.
    * Fixes #451 .

### DIFF
--- a/modes/company/evil-collection-company.el
+++ b/modes/company/evil-collection-company.el
@@ -63,8 +63,7 @@ be set through custom or before evil-collection loads."
   "Return non-nil if `evil-state' is in supported states."
   (cond
    ((not (bound-and-true-p evil-mode)) t)
-   ((eq command 'prefix)
-    (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states)))
+   ((eq command 'prefix) (memq evil-state evil-collection-company-supported-states))
    (t t)))
 
 ;;;###autoload
--- a/modes/company/evil-collection-company.el
+++ b/modes/company/evil-collection-company.el
@@ -62,9 +62,9 @@ be set through custom or before evil-collection loads."
 (defun evil-collection-company-supported-p (command &rest _)
   "Return non-nil if `evil-state' is in supported states."
   (cond
+   ((not (bound-and-true-p evil-mode)) t)
    ((eq command 'prefix)
-    (or (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states))
-        (not (bound-and-true-p evil-mode))))
+    (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states)))
    (t t)))
 
 ;;;###autoload
--- a/modes/company/evil-collection-company.el
+++ b/modes/company/evil-collection-company.el
@@ -63,12 +63,9 @@ be set through custom or before evil-collection loads."
   "Return non-nil if `evil-state' is in supported states."
   (cond
    ((eq command 'prefix)
-    (memq evil-state evil-collection-company-supported-states))
+    (or (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states))
+        (not (bound-and-true-p evil-mode))))
    (t t)))
-
-(defun evil-collection-company-popup ()
-  "When `evil-mode' is active, make `company-mode' not show popup if not in supported state."
-  (advice-add 'company-call-backend :before-while 'evil-collection-company-supported-p))
 
 ;;;###autoload
 (defun evil-collection-company-setup ()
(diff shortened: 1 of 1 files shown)


modes/company/evil-collection-company.el
```diff
@@ -98,7 +102,8 @@ be set through custom or before evil-collection loads."
       (company-tng-configure-default)))
 
   ;; Make `company-mode' not show popup when not in supported state
-  (advice-add 'company-call-backend :before-while 'evil-collection-company-supported-p))
+  (add-hook 'evil-mode-hook 'evil-collection-company-popup))
```
jojojames, line 105: What happens if you toggle evil-mode on and off? It seems like the advice in this case is stick, e.g. won't go away.

modes/company/evil-collection-company.el
```diff
@@ -62,9 +62,13 @@ be set through custom or before evil-collection loads."
...
-    ((eq command 'prefix)
-     (memq evil-state evil-collection-company-supported-states))
-    (t t)))
+   ((eq command 'prefix)
+    (memq evil-state evil-collection-company-supported-states))
+   (t t)))
+
+(defun evil-collection-company-popup ()
+  "When `evil-mode' is active, make `company-mode' not show popup if not in supported state."
+  (advice-add 'company-call-backend :before-while 'evil-collection-company-supported-p))
```
jojojames, line 71: It might be better to modify evil-collection-company-supported-p to check if evil-mode is on first (e.g. (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states)

Randy1Burrell: This is a good solution, but it causes the same problem. It changes the company-call-backend when I enter `evil-mode` and doesn't change it back after `evil-mode` gets deactivated, so it still sticks. I will look at this with a fresh pair of eyes in the morning.

jojojames: Why does it need to change? The function itself should be handling the logic, no?

Randy1Burrell: Yes, the function should be handling the logic. What I meant is

```emacs-lisp
(and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states))
```
won't work as it is. There should be an `or` statement to ensure that if `evil-mode` is not active then this function doesn't return `nil`. Adding to what you suggested, something along the line:
```emacs-lisp
(or (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states))
 (not (bound-and-true-p evil-mode)))
```
looks to work. I am going to try this out and see what happens.

Randy1Burrell: I just pushed an update.

jojojames: You can just add an additional clause checking explicitly if evil-mode is active and return true if it's not active instead of making the current clause more complex.

Randy1Burrell: That could work too. I'm new to Emacs so thanks for the info, I appreciate it. I'll check if it works and push an update.

modes/company/evil-collection-company.el
```diff
@@ -62,9 +62,10 @@ be set through custom or before evil-collection loads."
 (defun evil-collection-company-supported-p (command &rest _)
   "Return non-nil if `evil-state' is in supported states."
   (cond
-    ((eq command 'prefix)
-     (memq evil-state evil-collection-company-supported-states))
-    (t t)))
+   ((not (bound-and-true-p evil-mode)) t)
+   ((eq command 'prefix)
+    (and (bound-and-true-p evil-mode) (memq evil-state evil-collection-company-supported-states)))
```
jojojames, line 67: You don't need this second bound-and-true check.

Randy1Burrell: I left this in for safe measure. I don't think an extra check will cause much computing power, but I will remove it for your preference.

jojojames: If you left it in, it defeats the purpose of adding the original clause, whose purpose was to simplify the conditional logic.

jojojames:
Nit, it looks like the commits weren't squashed together properly.

Randy1Burrell:
Must be something wrong with Magit from my end. I recently installed forge some maybe that have changed something. I will have to look into that.

PR #453, opened 2021-02-01
Add evil-collection-define-operator-key

jojojames:
Some context: https://github.com/emacs-evil/evil-collection/pull/91#issuecomment-366181047

https://github.com/emacs-evil/evil-collection/commit/6bdbcccf1e4f51c75f5cf54b3dc27419b32fe608

This is for attaching additional keys to operator commands.

e.g. Add "f" to prefix/operator (evil-yank) "y" to get

"yf"

while still maintaining the other operator commands like "yy".

Would be happy if you had time to take a look @noctuid, curious if general has something already for this use case.

@condy0919 Appreciate a look before I merge.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -352,32 +352,38 @@ to filter keys on the basis of `evil-collection-key-whitelist' and
           (push key filtered-bindings)
           (push def filtered-bindings))))
     (setq filtered-bindings (nreverse filtered-bindings))
-    (cond ((null filtered-bindings))
-          ((and (boundp map-sym) (keymapp (symbol-value map-sym)))
-           (condition-case-unless-debug err
-               (apply #'evil-define-key*
-                      state (symbol-value map-sym) filtered-bindings)
-             (error
-              (message "evil-collection: error setting key in %s %S"
-                       map-sym err))))
-          ((boundp map-sym)
-           (user-error "evil-collection: %s is not a keymap" map-sym))
-          (t
-           (let* ((fname (format "evil-collection-define-key-in-%s" map-sym))
-                  (fun (make-symbol fname)))
-             (fset fun `(lambda (&rest args)
-                          (when (and (boundp ',map-sym) (keymapp ,map-sym))
-                            (remove-hook 'after-load-functions #',fun)
-                            (condition-case-unless-debug err
-                                (apply #'evil-define-key*
-                                       ',state ,map-sym ',filtered-bindings)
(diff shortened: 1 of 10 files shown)


condy0919:
Sorry, it's beyond my level :'(

jojojames:
> Sorry, it's beyond my level :'(

Shouldn't be too bad :), and would be helpful to diff the two define functions to gain a better understanding of the delta between the two.

Better to read the commits separately, the first commit just does a simple refactor, moving the binding implementation out to a helper function.

https://github.com/emacs-evil/evil-collection/pull/453/commits/a82eb0110be4ac12e2481479953d03e78ed91661

The purpose is so I can reuse the code later.

The second commit:

https://github.com/emacs-evil/evil-collection/pull/453/commits/d1dd2cfa77b0943c3ab111b2f3e4900ec04ecb48#diff-25b530b1930b6d328a2e79d0628153175a9436a842ed2e37f2e1c7e6b69c34ceR353

is almost 1:1 with the original but I'm changing:

1. What keys I'm checking in the white/black list filter by concatenating a prefix and checking against that.

line 353:

```
(key-with-prefix (concat prefix key))
```

line 365:

```
(when (or (and whitelist (member key-with-prefix whitelist))
```

2. The definition of the function being bound:

You can see the original definition in modes/pass/evil-collection-pass.el, which is just a define-key to a menu-item.

```
(evil-collection-define-key 'operator 'pass-mode-map
    ;; Like `eww'.
    "f" '(menu-item
          ""
          nil
          :filter (lambda (&optional _)
                    (when (memq evil-this-operator
                                evil-collection-yank-operators)
                      (setq evil-inhibit-operator t)
                      'pass-copy-field))))
```

You can see I'm building the definition in the let and then using that for the binding later.

```
(def
              `(menu-item
                ""
                nil
                :filter
                (lambda (&optional _)
                  (when (or
                         (eq evil-this-operator (key-binding ,remap))
                         (memq evil-this-operator ,operators))
                    (setq evil-inhibit-operator t)
                    ',(pop bindings))))
```

jojojames:
Been using this for a bit now and it seems good enough for me, would appreciate a look when time frees.

noctuid:
No, currently general only has the generic `general-key-dispatch`. This seems like a useful shorthand to add. There might be issues if the operator is one besides delete and yank and still needs to be inhibited (though that would be a rare case).

PR #470, opened 2021-04-14
Magit: Keep bindings consistent with prefix key in magit-revert

MetaDark:
"O" is used to launch `magit-reset`, not `magit-revert`. "\_" and "-" should be used for the suffix commands to be consistent with the prefix key "\_".


jojojames:
Seems reasonable to me, I'll merge it now but may make further changes if other users raises issues around this.

PR #471, opened 2021-04-14
shell: Default to the normal state like in compilation-mode

MetaDark:
`shell-mode` derives from `comint-mode`, so it currently defaults to the insert state. Although, like `compilation-mode`, `shell-mode` is rarely used interactively and should default to the normal state.

### DIFF
--- a/modes/shell/evil-collection-shell.el
+++ b/modes/shell/evil-collection-shell.el
@@ -1,0 +1,39 @@
+;;; evil-collection-shell.el --- Evil bindings for `shell' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2021 Kira Bruneau
+
+;; Author: Kira Bruneau <kira.bruneau@pm.me>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <mail@ambrevar.xyz>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, shell, tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


modes/shell/evil-collection-shell.el
```diff
@@ -0,0 +1,39 @@
+;;; evil-collection-shell.el --- Evil bindings for `shell' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2021 Kira Bruneau
+
+;; Author: Kira Bruneau <kira.bruneau@pm.me>
+;; Maintainer: James Nguyen <james@jojojames.com>
+;; Pierre Neidhardt <mail@ambrevar.xyz>
```
jojojames, line 7: Can remove this line.

condy0919:
`shell-mode` is more like `term`/`ansi-term`  used as a REPL. Making a REPL immutable by default doesn't make sense.

jojojames:
What's an example to get into shell-mode? As far as I understand, it's used for other modes to build on top of.

KSSysLab:
> What's an example to get into shell-mode? As far as I understand, it's used for other modes to build on top of.

`M-x shell`

MetaDark:
Oh right sorry, I usually use `shell-mode` through commands like `shell-command` / `async-shell-command`, which act more like the `compile` command, but I forgot about the interactive case.

I wonder if there's anyway to distinguish this, because both cases use the same mode.

condy0919:
``` elisp
(setq proc
			  (start-process-shell-command "Shell" buffer command)))
		  (setq mode-line-process '(":%s"))
                  (shell-mode)
                  (setq-local revert-buffer-function
                              (lambda (&rest _)
                                (async-shell-command command buffer)))
                  (set-process-sentinel proc #'shell-command-sentinel)
```

Maybe by checking the `real-this-command`?

condy0919:
https://debbugs.gnu.org/cgi/bugreport.cgi?bug=47991

The buffer created by `shell-command` will be in `fundamental-mode` so the
`shell-mode-hook` won't be executed.

condy0919:
The bug ticket has been closed. So there is no way to `add-hook` for `M-!` (`shell-command`).

PR #473, opened 2021-04-22
Added help prompt for reftex-toc (with comments)

arifer612:
I noticed that the help message for refTeX-TOC did not specify the evil bindings and I suppose that it was because the refTeX-TOC prompt could not be changed from this add-on since it was hard-coded in the original file [here](https://github.com/emacs-mirror/emacs/blob/3e41eccc27586fd9e24369a2c82db483a97ad9eb/lisp/textmodes/reftex-toc.el#L267).

However, all but one of the keybindings in the prompt are the same in the evil-bindings so I do not see a point in not showing the help message with evil-bindings.

### DIFF
--- a/modes/reftex/evil-collection-reftex.el
+++ b/modes/reftex/evil-collection-reftex.el
@@ -162,7 +162,7 @@
     "zi" 'reftex-toc-display-index
     "x" 'reftex-toc-external
     "." 'reftex-toc-show-calling-point
-    "gs" 'reftex-toc-jump
+    "J" 'reftex-toc-jump
     (kbd "[[") 'reftex-toc-previous-heading
     (kbd "]]") 'reftex-toc-next-heading
     (kbd ">") 'reftex-toc-promote
--- a/modes/reftex/evil-collection-reftex.el
+++ b/modes/reftex/evil-collection-reftex.el
@@ -74,14 +74,20 @@
 (setq reftex-toc-help
       "                      AVAILABLE KEYS IN TOC BUFFER
                       ============================
-j / k      next-line / previous-line
-TAB        Goto the location and keep the TOC window.
-RET        Goto the location and hide the TOC window (also on mouse-2).
-q / ZZ     Hide/Kill *toc* buffer, return to position of reftex-toc command.
-l          Toggle display of  [l]abels
-f          Toggle follow mode
-r / gr     Reparse the LaTeX document     / Reparse entire LaTeX document.
-x          Switch to TOC of external document (with LaTeX package `xr').")
+    j / k      next-line / previous-line
+    go         Show the corresponding location of the LaTeX document.
+    TAB        Goto the location and keep the TOC window.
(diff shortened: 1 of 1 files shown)


arifer612:
I'd also like to ask what is the opinion on adding more keybindings for refTeX-TOC? I feel that there is too few bindings given the API for refTeX. If the response is positive I can it into this PR (or another one)

condy0919:
`substitute-command-keys` is better than hard code.

condy0919:
I'm not a refTeX user, no thoughts for adding more bindings. Sorry.

arifer612:
I completely agree that hard-coding is not preferred, but unless someone over at GNU changes that I doubt we can do anything about that. Nevertheless, I still think that instead of leaving the help message with all the standard Emacs keybindings, it should show evil bindings if the user has it on, thus the PR.

No thoughts is good! That means I can propose some of my changes then :D I'll get to them once I'm done with my work!

modes/reftex/evil-collection-reftex.el
```diff
@@ -121,6 +134,8 @@
 
   ;; This one is more involved, in reftex-toc.el, line 282 it shows the prompt
   ;; string with the keybinds and I don't see any way of changing it to show evil-like binds.
+  ;;;; Since bindings that are specified in the prompt are the same except for SPC, wouldn't
```
jojojames, line 137: Remove comment here.

jojojames:
Looks good otherwise, if you have other keybinding proposals, feel free to submit them.

modes/reftex/evil-collection-reftex.el
arifer612: Got it, I will remove the comments once I have settled on the other keybindings to include

arifer612:
@jojojames I have added in evil-ised keybindings that I think will be useful for others who use refTeX. There are some with the insertion keybindings "i", "c" etc, but it does not matter since the buffer is muted and cannot be edited, to begin with.

arifer612:
@jojojames Can I get some help with merging and closing up this pull request?

modes/reftex/evil-collection-reftex.el
```diff
@@ -119,23 +138,35 @@
...
-    "?" 'reftex-toc-show-help
+    "i" 'reftex-toc-toggle-index
+    "c" 'reftex-toc-toggle-context
+    "F" 'reftex-toc-toggle-file-boundary
+    "rl" 'reftex-toc-rename-label
+    "zi" 'reftex-toc-display-index
     "x" 'reftex-toc-external
-    ;; (kbd "SPC") 'reftex-toc-view-line
+    "." 'reftex-toc-show-calling-point
+    "gs" 'reftex-toc-jump
```
jojojames, line 165: Maybe "J", I believe we usually use that for jump.

arifer612: Makes sense. Fixed that!

jojojames:
Thanks!

Issue #474, opened 2021-04-23
[Bug/Request] Consider tuning <escape> in ivy-minibuffer, but not in general ivy-mode-map


ianpan870102:
In https://github.com/emacs-evil/evil-collection/issues/139 and https://github.com/emacs-evil/evil-collection/pull/142, a nice effort have been made to rebind `<escape>` to minibuffer-keyboard-quit in an ivy minibuffer.

However, the implementation raises one problem: If a user switches to Emacs mode/evil-emacs-state (quite useful for editing magit commit messages, or working in vterm/term), the `<escape>` key cannot work its classic Meta-functionalities, such as interpreting `<escape> f` as `M-f`, or `<escape> b` as `M-b`.

I inspected the code and realized that instead of binding `<escape>` to minibuffer-keyboard-quit in a ivy-minibuffer-mode-map, **the PR incorrectly bound it to ivy-mode-map**.

I have tested the following code in my own init.el, and solved the issue:

```emacs-lisp
(use-package evil-collection
  :after evil
  :config
  (setq evil-collection-company-use-tng nil)
  (evil-collection-init)
  (with-eval-after-load 'ivy
    (define-key ivy-mode-map (kbd "<escape>") nil)
    (define-key ivy-minibuffer-map (kbd "<escape>") #'minibuffer-keyboard-quit)))
```

This snippet preserves the functionality of quitting an ivy minibuffer with a single escape keystroke, but also gives the escape key back to Meta when a user enters emacs state (Emacs mode).

ianpan870102:
Closing issue, as PR #475 is merged.

PR #475, opened 2021-04-23
Fix escape in in emacs-state, yet preserving minibuffer quit functionality

ianpan870102:
This PR fixes issue #474 (also raised by me).

This fixes the problem: If a user switches to Emacs mode/evil-emacs-state (quite useful for editing magit commit messages, or working in vterm/term), the `<escape>` key cannot work its classic Meta-functionalities, such as interpreting `<escape> f` as `M-f`, or `<escape> b` as `M-b`.

For further details, please kindly refer to issue #474.

### DIFF
--- a/modes/ivy/evil-collection-ivy.el
+++ b/modes/ivy/evil-collection-ivy.el
@@ -37,7 +37,7 @@
 ;;;###autoload
 (defun evil-collection-ivy-setup ()
   "Set up `evil' bindings for `ivy-mode'."
-  (evil-collection-define-key nil 'ivy-mode-map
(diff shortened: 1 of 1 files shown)


condy0919:
Does it play well when `evil-collection-setup-minibuffer` is enabled?

ianpan870102:
> Does it play well when evil-collection-setup-minibuffer is enabled?

Personally I don't use evil-collection-setup-minibuffer. But I just set it to `t` and it seems fine at first glance. Is there any specific functionality you'd like me to test out?

condy0919:
Thanks, merged.

PR #476, opened 2021-04-24
Magit: Properly format rebase instructions when commentChar is "%"

### LINKED ISSUE #468: evil-magit rebase instructions are improperly formated with `%` commentChar
When rebasing with git `core.commentChar` set to `%`, the magit rebase buffer command descriptions will look something like this:

```
% Commands:
%-8s p
%-8s r
%-8s e
%-8s s
...
```

Whereas with a different comment character (e.g. `@`), they look like this (as expected):

```
@ Commands:
@ p        pick = use commit
@ r        reword = use commit, but edit the commit message
@ e        edit = use commit, but stop for amending
@ s        squash = use commit, but meld into previous commit
...
```

The culprit looks to be [here](https://github.com/emacs-evil/evil-collection/blob/master/modes/magit/evil-collection-magit.el#L511).

A potential solution would be to have `comment-start` be an argument to `format`, rather than be prepended to the format string directly.

nacl:
This change adjusts the formatting of the instructions in the `git-rebase`
buffer to allow them to be properly formatted when git's `core.commentChar` is
set to `%`.  In particular, lines will look something like this:

```
%-8s p
```

The root cause seems to be the how the lines are constructed: comment characters
are directly added to the format string.  When it is `%`, `format` appears to
proceed until it finds the next `%` character, (corresponding to the `%` in
`%-8s`) above and prints it directly to the string rather than interpreting it
something to format.

This is fixed by constructing the format string with the comment character as an
argument to `format` rather than being added directly to the format string.

Fixes #468.

### DIFF
--- a/modes/magit/evil-collection-magit.el
+++ b/modes/magit/evil-collection-magit.el
@@ -508,7 +508,8 @@ denotes the original magit key for this command.")
                (flush-lines (concat "^" (regexp-quote comment-start) ".+ = "))
                (dolist (cmd evil-collection-magit-rebase-commands-w-descriptions)
                  (insert
-                  (format (concat comment-start " %-8s %s\n")
+                  (format "%s %-8s %s\n"
+                          comment-start 
                           (if (and (car cmd)
                                    (eq (nth 1 cmd)
                                        (lookup-key aux-map (kbd (car cmd)))))


jojojames:
Thanks.

PR #480, opened 2021-05-03
Fix Jd binding

rgrinberg:
`prodigy-jump-dired` does not exist. `prodigy-jump-file-manager` is
likely the intended function.


jojojames:
Thanks!

PR #483, opened 2021-05-14
Add quit-window binding to vc-dir-mode

jellelicht:


condy0919:
Use `evil-collection-set-readonly-bindings` instead.

jellelicht:
Fixed ‘n pushed

PR #484, opened 2021-05-16
update evil-collection-consult--mark-candidates

tpeacock19:
fixing the function that collects mark candidates after an update to consult internal functions


jojojames:
Thanks.

PR #486, opened 2021-05-25
Tide/references mode: Go to reference at cursor

Tommos0:
Current implementation seems to only work for the first reference in a file. This patch should fix that.


condy0919:
Thanks.

PR #487, opened 2021-05-27
reflex: update documentation

dankessler:
A [recent commit](https://github.com/emacs-evil/evil-collection/commit/8244f8806ecb52574827ec30d3c9cda7d8a8e67c) remapped reflex-toc-jump from `gs` -> `J`, but the help didn't get updated (so it erroneously still suggests using `gs` to jump to section).


condy0919:
Good catch!

PR #488, opened 2021-06-02
[transmission] Fix keybinding

arifer612:
`transmission-delete` and `transmission-remove` have different purposes and
should both be included as bindings.

`transmission-delete`: Removes torrent from Transmission client and unlinks files from drive.
`transmission-remove`: Removes torrent from Transmission client.

`I` has been set to inputting labels for torrents using `transmission-label`.


condy0919:
thanks

PR #465, opened 2021-06-14
Feature request: add an option like evil-disable-insert-state-bindings

daanturo:
Can `evil-collection` exclude insert state bindings when `evil-disable-insert-state-bindings` is non-nil?

It's confusing since I have already set that variable but in some modes' insert state, evil bindings are still around.

Perhaps we can add an additional check in `evil-collection-define-key`?

```elisp
(defun evil-collection-define-key (state map-sym &rest bindings)
  ;; ...
  (when-let ((state
              (cond
               ((not evil-disable-insert-state-bindings) state) ; Unaffected, just move on
               ((eq 'insert state) nil)
               ((and (listp state)) (remove 'insert state)))))
    (let*
        ;; ...
        )))
```

Or to be less disruptive, let's define a variable named `evil-collection-disable-insert-state-bindings`?

Edit: clarified suggestion.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -388,6 +388,17 @@ binding in `annalist' as so."
     (setq filtered-bindings (nreverse filtered-bindings))
     (evil-collection--define-key 'operator map-sym filtered-bindings)))
 
+(defun evil-collection--filter-states (states)
+  "Filter STATES with respect to `evil-disable-insert-state-bindings'.
+If `evil-disable-insert-state-bindings' is non-nil:
+Return STATES without `insert': if it is `insert' -> nil, list -> list without `insert';
+Else return STATES as-is."
+  (cond
+   ((not evil-disable-insert-state-bindings) states)
+   ((eq 'insert states) nil)
+   ((listp states) (remove 'insert states))
+   (t states)))
+
 (defun evil-collection-define-key (state map-sym &rest bindings)
   "Wrapper for `evil-define-key*' with additional features.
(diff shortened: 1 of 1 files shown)


daanturo:
> There's something here: https://github.com/emacs-evil/evil/blob/d3cc85bf828f04256de4a36ddb7624064b0252c8/evil-maps.el#L400 that I wonder if we could leverage. I didn't look too closely at that function though.s at all.

About `evil-update-insert-state-bindings`: I think we can't utilize it as-is, since it was specifically designed for `evil-insert-state-map` only, but `evil-collection` revolves around various modes.

Or did you mean something like this (just a prototype, there should be definitely a better version)?

```lisp
(defun evil-collection-remove-insert-state-bindings (map-sym)
  "Remove all insert-state bindings from MAP-SYM."
  (setcdr   (evil-lookup-key (eval map-sym) [insert-state])   nil))
```
Then we create a custom variable like `evil-collection-disable-insert-state-bindings` and ran this on all `evil-collection-mode-list` when it is turned off, and re-run `evil-collection-init` when turned on?

> I think this may open up the question of whether or not we want a filter list similar to the black/white list we have now, but specifically for states to enable/disable. E.g. A user can set a a filter to not have insert state binds at all.

IMO this is completely feasible. But from my perspective, casually we only care about insert and normal: sometimes we want bindings in normal state but not in insert state (like `evil-disable-insert-state-bindings`'s author has predicted), and vice versa; on the other hand, normal but not visual/motion seems a bit rare.

condy0919:
Closed in favor of #493

PR #493, opened 2021-06-23
Make filtering states possible

daanturo:
Solves #465 by defining `evil-collection-state-passlist` and `evil-collection-state-denylist` analogous to `evil-collection-key-{black,white}list`.

If you find the new variable names alien to the latter's, please tell me to rename them. I also prefer the old naming scheme but have seen the trend such as `native-comp-deferred-compilation-deny-list`.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -410,7 +410,7 @@ binding in `annalist' as so."
   "Return a list states after filtering STATE (a single symbol or list of symbols).
 The return value adheres to `evil-collection-state-passlist' and
 `evil-collection-state-denylist'. When the STATE is `nil', which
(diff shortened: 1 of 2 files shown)


condy0919:
> evil-collection.el:764:1:Error: the following functions are not known to be defined: seq-difference, seq-intersection

Seems like Emacs 26.3 doesn't provide `seq-difference` and `seq-intersection`?

daanturo:
> Updated: `seq-intersection` and `seq-difference` in Emacs 26.3 are not `autoload`ed. So a `(require 'seq)` is required.

Done, thank you.

condy0919:
Thanks

Issue #492, opened 2021-06-22
evil-collection-magit-mode-map-bindings interprets evil-toggle-key in a wrong way


geza-herman:
`evil-toggle-key`'s usual value is `C-z`. This value is interpreted as a prefix key instead of the actual ctrl-z combination. Pressing `C` in `magit-status` starts a prefix sequence instead of `magit-clone`.

The line

```((,evil-collection-magit-state) magit-mode-map ,evil-toggle-key evil-emacs-state)```

should read something similar to this:

```((,evil-collection-magit-state) magit-mode-map ,(read-kbd-macro evil-toggle-key) evil-emacs-state)```

Note: I'm not sure that this line is needed, it seems that `C-z` still works for me if I remove this line. Perhaps it was added as a workaround for something?

**Comment by condy0919** [comment]
Agreed.

@jojojames Can we remove the "redundant" line?

**Comment by jojojames** [comment]
What about using:

`(kbd ,evil-toggle-key)


```
(read-kbd-macro evil-toggle-key)
returns: ""
`(kbd ,evil-toggle-key)
returns: (kbd "C-z")
```

Please PR, I think we want to make the change instead of removing the line.

condy0919:
I fix that

PR #495, opened 2021-06-28
Properly bind escape in magit-mode-map

progfolio:
Without using `kbd', each key in the string "\<escape>" is interpreted as a
prefix key.


progfolio:
Related to #492

condy0919:
Good catch

PR #496, opened 2021-07-01
[elfeed]: add missing keybind

NEXCydran:
The command that resets the search filter to it's default value, `elfeed-search-clear-filter`, was missing it's keybinding.
elfeed bind this to `c` by default.


condy0919:
thanks

PR #497, opened 2021-07-03
update evil-collection-consult--mark-candidates for upstream change

tpeacock19:
internal function needed to be updated inline with an upstream
change. The `markers` variable is ignored and only current-buffer evil
marks are shown.


tpeacock19:
I needed to use the variable in order to pass the check. I've updated the docstring instead.

condy0919:
Thanks

tpeacock19:
Ah sorry, I need to do one more update on this. I didnt realize `consult-mark` provides a default mark-ring.

PR #501, opened 2021-07-06
Fix up Info-mode

MaxCan-Code:


KSSysLab:
I'm curious about why `digit-argument` bindings are not working out of box.

When pressing `0`, Emacs will try to find the commands in this [order](https://www.gnu.org/software/emacs/manual/html_node/elisp/Searching-Keymaps.html). It can be confirmed that `emulation-mode-map-alists` will be used. And the first element is `evil-mode-map-alist`. Let me dig deeper.

``` elisp
;; eval under evil-normal-state
(mapcar 'car evil-mode-map-alist)
;;=> (t evil-normal-state-local-minor-mode outline-minor-mode evil-collection-unimpaired-mode t evil-normal-state-minor-mode t evil-motion-state-local-minor-mode visual-line-mode evil-collection-unimpaired-mode evil-motion-state-minor-mode)
```

Found that `evil-motion-state-minor-mode` is in this alist. And

``` elisp
;; eval under evil-normal-state
(describe-keymap (alist-get 'evil-motion-state-minor-mode evil-mode-map-alist))

;; ...
;; ,               evil-repeat-find-char-reverse
;; -               evil-previous-line-first-non-blank
;; /               evil-search-forward
;; 0               evil-digit-argument-or-evil-beginning-of-line
;; 1 .. 9          digit-argument
;; :               evil-ex
;; ;               evil-repeat-find-char
;; ?               evil-search-backward
;; B               evil-backward-WORD-begin
;; E               evil-forward-WORD-end
;; ...
```

[1, 9] have already bound to `digit-argument`.

So when I press `0` in evil normal state, Emacs should find it for us. However, Emacs runs `Info-nth-menu-item` which is bound in info.el.

KSSysLab:
```
evil-overriding-maps is a variable defined in ‘evil-vars.el’.

Its value is shown below.

  You can customize this variable.

Keymaps that should override Evil maps.
Entries have the form (MAP-VAR . STATE), where MAP-VAR is
a keymap variable and STATE is the state whose bindings
should be overridden. If STATE is nil, all states are
overridden.

Value:
((Buffer-menu-mode-map)
 (color-theme-mode-map)
 (comint-mode-map)
 (compilation-mode-map)
 (grep-mode-map)
 (dictionary-mode-map)
 (ert-results-mode-map . motion)
 (Info-mode-map . motion)
 (speedbar-key-map)
 (speedbar-file-key-map)
 (speedbar-buffers-key-map))

[back]
```

`Info-mdoe-map` is in `evil-overriding-maps`, so that the commands bound in evil motion state map will have no effects.

MaxCan-Code:
`Info-nth-menu-item` still overrides `digit-argument` after I delete `Info-mode-map` from `evil-overriding-maps`:
https://github.com/emacs-evil/evil-collection/blob/d8d8b13c3e73cdb499141340816cf1b141ab1ea8/modes/info/evil-collection-info.el#L37-L41
```elisp.diff
(defun evil-collection-info-setup ()
  "Set up `evil' bindings for `info-mode'."
  (evil-collection-set-readonly-bindings 'Info-mode-map)
+ (assoc-delete-all 'Info-mode-map evil-overriding-maps)
  (evil-set-initial-state 'Info-mode 'normal)
  (evil-collection-define-key 'normal 'Info-mode-map
```
Let me know if there's a better way to do this

condy0919:
If you're a `use-package` user, use `:custom` to customize the `evil-overriding-maps`.

condy0919:
FYI, I propose to change the default value of `evil-overriding-maps` https://github.com/emacs-evil/evil/issues/1488

MaxCan-Code:
When that change comes through, I can remove the manual evil overrides in
https://github.com/emacs-evil/evil-collection/blob/d8d8b13c3e73cdb499141340816cf1b141ab1ea8/modes/info/evil-collection-info.el#L97-L106
and
https://github.com/emacs-evil/evil-collection/blob/d8d8b13c3e73cdb499141340816cf1b141ab1ea8/modes/info/evil-collection-info.el#L115-L130

condy0919:
The PR is merged.

MaxCan-Code:
The mappings are fixed

condy0919:
Thanks

Issue #48, opened 2017-12-04
[Question] Is evil-make-overriding-map purposely avoided?


dieggsy:
Evil-integration does this to some extent, and it seems like the right approach as it avoids re-defining default keys that you want to stay the same, while you can override keys relevant to evil. This might also make some of the code neater and easier to follow IMO.

dieggsy:
For example, I see `evil-dired.el` uses a lot of the same bindings as are default in `dired-mode-map`, so it's not necessarily easy to tell what's new.

jojojames:
I've understood the dired integration in evil-integration to be pretty broken. Don't those overriding maps create duplicate menus? IIRC, I believe so.

I actually think it makes it more confusing than just using evil-define-key everywhere. At least with evil-define-key, you know the precedence order of the keymaps so it's somewhat easy to reason about.

I'd actually like to either request or make a patch to evil-mode to make that integration (evil-integration.el) optional and just port over the good stuff from that file to this project.

Tangentially, I think @noctuid's general library is pretty slick. I wish evil came with something like that directly instead of a separate library.

dieggsy:
@jojojames dired integration seems to work ok for me? but perhaps you're right in that regard, i haven't looked at the internals of `make-overriding-map` that closely. If the internals are simpler, we should probably keep the current system, though it'd still be nice to have comments on what was taken from the default and what was changed or added.

Agreed on making integration optional, and I use and love general.el as well :)

jojojames:
"Pretty broken" was a heavy comment. It doesn't have very good vim integration first and the duplicate menus are a pretty bad experience in my opinion.

dieggsy:
@jojojames what do you mean by "duplicate menus" here?

jojojames:
Last I checked, the menus in the top toolbar (menu) were duplicated.

Maybe it's fixed by now, but I doubt it.

dieggsy:
@jojojames oh, I always run emacs with `(menu-bar-mode -1)` so that doesn't affect me in any way, but it appears you're right.

noctuid:
+1 to not using `make-overriding-map`. I think it is much better for all keybindings to be made explicit, and `make-overriding-map` has extra effects that may not be expected/desired (e.g. if the user or some other package binds a key in `dired-mode-map` when it is an overriding map).

> I'd actually like to either request or make a patch to evil-mode to make that integration (evil-integration.el) optional and just port over the good stuff from that file to this project.

[...]

PR #506, opened 2021-07-17
evil-collection: respect evil-overriding-maps

condy0919:
Respect `evil-overriding-maps`.

If `dictionary-mode-map` is in `evil-overriding-maps`, `h`/`l`/`n`/`p` will not
be overridden by evil. Currently `evil-collection` overrides theses keys, see
https://github.com/emacs-evil/evil-collection/blob/master/modes/dictionary/evil-collection-dictionary.el#L42-L45

If `evil-overriding-maps` is `'()` (default now), `h`, `l`, `n` and `p` will be defined by evil as these are the basic keys. In such a case, `evil-collection` needn't to redefine. It will help us to maintain less keybindings.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -85,7 +85,7 @@ See https://github.com/emacs-evil/evil-collection/issues/60 for more details.")
 (declare-function org-table-align "org-table.el" nil)
 
 (defgroup evil-collection nil
-  "A set of keybindings for Evil mode"
+  "A set of keybindings for Evil mode."
   :group 'evil)
 
 (defcustom evil-collection-setup-minibuffer nil
@@ -422,7 +422,7 @@ means all states for `evil-define-key', return `nil'."
        states)
      evil-collection-state-denylist)))
 
-(defun evil-collection-define-key (state map-sym &rest bindings)
+(cl-defun evil-collection-define-key (state map-sym &rest bindings)
   "Wrapper for `evil-define-key*' with additional features.
 Unlike `evil-define-key*' MAP-SYM should be a quoted keymap other than the
 unquoted keymap required for `evil-define-key*'. This function adds the ability
@@ -431,8 +431,23 @@ to filter keys on the basis of `evil-collection-key-whitelist' and
   (declare (indent defun))
   (let* ((whitelist (mapcar 'kbd evil-collection-key-whitelist))
          (blacklist (mapcar 'kbd evil-collection-key-blacklist))
+         (overriding (assoc map-sym evil-overriding-maps))
          (states-to-bind (evil-collection--filter-states state))
          filtered-bindings)
+    ;; We take account of `evil-overriding-maps' first, as it's defined in evil,
+    ;; not evil-collection.
+    ;;
+    ;; `overriding' is non-nil, and cdr of `overriding' is nil, which means evil
+    ;; shouldn't define any keybindings in that keymap.
+    ;;
+    ;; `overriding' is non-nil, and cdr of `overriding' is non-nil, which means
+    ;; evil shouldn't define keybindings in that keymap in such state.
+    ;;
+    ;; `overriding' is nil, no constraints.
+    (when overriding
+      (if (null (cdr overriding))
+          (cl-return-from evil-collection-define-key)
+        (setq states-to-bind (seq-difference states-to-bind (list (cdr overriding))))))
     (when (or states-to-bind (null state))
       (while bindings
         (let ((key (pop bindings))


jojojames:
Hmnn, I'm very hesitant to support respecting this evil-override-map. IIRC, ideally we'd want to get away from override maps in the first place.

Secondly, I'm not sure all the implications of each comment here:


    ;; `overriding' is non-nil, and cdr of `overriding' is nil, which means evil
    ;; shouldn't define any keybindings in that keymap.
    
What if evil-collection wanted to define a,b,c,d,e,f but the overriding map was nil? Why would overriding be non nil but the cdr of overriding be nil? When would that happen?

    ;; `overriding' is non-nil, and cdr of `overriding' is non-nil, which means
    ;; evil shouldn't define keybindings in that keymap in such state.
    
What if evil-collection wanted to define a,b,c,d,e,f but the overriding map was a,b,c? What would happen to the d,e,f?

condy0919:
> Hmnn, I'm very hesitant to support respecting this evil-override-map. IIRC, ideally we'd want to get away from override maps in the first place.
>
> Secondly, I'm not sure all the implications of each comment here:
>
> ```
> ;; `overriding' is non-nil, and cdr of `overriding' is nil, which means evil
> ;; shouldn't define any keybindings in that keymap.
> ```
>
> What if evil-collection wanted to define a,b,c,d,e,f but the overriding map was nil? Why would overriding be non nil but the cdr of overriding be nil? When would that happen?

The cdr of `overriding` is a evil state whose evil bindings should be overridden
by the original keymap. When it's nil, it means all evil state bindings will be
overriden. e.g.

``` emacs-lisp
(evil-collection-define-key 'normal 'Info-mode-map
  "h" 'evil-backward-char)
```

But `h` will be bound to `Info-history` due to override.

>
> ```
> ;; `overriding' is non-nil, and cdr of `overriding' is non-nil, which means
> ;; evil shouldn't define keybindings in that keymap in such state.
> ```
>
> What if evil-collection wanted to define a,b,c,d,e,f but the overriding map was a,b,c? What would happen to the d,e,f?

evil-collection will skip a,b,c and evil-collection-define-key only in d,e,f.
It's easier to implement by just skipping instead of overriding by the original
keymap.

jojojames:
> But h will be bound to Info-history due to override.

Could you elaborate more here?

What do you expect to happen before this change and what do you expect to happen after this change, and why?

> evil-collection will skip a,b,c and evil-collection-define-key only in d,e,f.
It's easier to implement by just skipping instead of overriding by the original
keymap.

I'm not sure that's what we want.

condy0919:
> > But h will be bound to Info-history due to override.
>
> Could you elaborate more here?
>
> What do you expect to happen before this change and what do you expect to happen after this change, and why?
>
> > evil-collection will skip a,b,c and evil-collection-define-key only in d,e,f.
> > It's easier to implement by just skipping instead of overriding by the original
> > keymap.
>
> I'm not sure that's what we want.

Oh, I found the example is wrong. All modes should have hjkl, e.g. `h` is
`evil-backward-char` in normal state.

But if `Info-mode-map` is in `evil-overriding-maps` and the cdr is nil, `h` will
be `Info-history` at last due to the override mechanism.

So we must manually use `evil-collection-define-key` to redefine `h` to `evil-backward-char` in some keymaps (e.g. `dictionary-mode-map`).

But is it user expected? If a keymap is in `evil-overriding-maps`, it means
users would like to use the vanilla keybindings.

So that we try to respect the `evil-overriding-maps`. Currently, the value of
`evil-overriding-maps` is `'()`, so that this PR won't create issues for most people.

After this patch, we can delete tons of redundant keybdings.

If `dictionary-mode-map` isn't in `evil-overriding-maps`,

``` emacs-lisp
(evil-collection-define-key 'normal 'dictionary-mode-map
  ;; motion
  (kbd "l") 'evil-forward-char  ; otherwise bound to `dictionary-previous'
  (kbd "h") 'evil-backward-char ; otherwise bound to `dictionary-help'
  (kbd "p") 'ignore             ; otherwise bound to `backward-button'
  (kbd "n") 'evil-search-next   ; otherwise bound to `forward-button'
  ;; mouse
  [mouse-1] 'link-selected
  ;; misc
  "g?" 'dictionary-help ; normally under `h` which is rebound here
  (kbd "C-o") 'dictionary-previous) ; normally under `l` which is rebound here
```

l/h/p/n can be removed safely.

jojojames:
> But if Info-mode-map is in evil-overriding-maps and the cdr is nil, h will
be Info-history at last due to the override mechanism.

Yes, we want 'h' to be 'evil-backward-char'.

> So we must manually use evil-collection-define-key to redefine h to evil-backward-char in some keymaps (e.g. dictionary-mode-map).

That's expected and wanted. In this case, the reasoning is 'this mode binds 'h' to something random, lets change it back to what a user would expect'.

> But is it user expected? If a keymap is in evil-overriding-maps, it means
users would like to use the vanilla keybindings.

That's not necessarily true. It could be that another mode (say evil-vars) is binding it for them without their intention.

> After this patch, we can delete tons of redundant keybdings.

If I'm understanding what you're saying, this isn't true. We'd be removing bindings -and also- changing what those bindings do.

Just going by your example, all the 'redundant' binds of hjkl that we've added to various modes will become what evil-overriding-map has in its map right? In this case, the binds in evil-collection serves a purpose of rebinding to hjkl.

> The diff in #501 also reflects the benefit.

From my quick skim through there, I'd say the problems lie in evil using those override maps for those various modes.

Somewhat related: https://github.com/emacs-evil/evil-collection/issues/48

condy0919:
> From my quick skim through there, I'd say the problems lie in evil using those override maps for those various modes.

Do you mean using `evil-overriding-maps` is not recommended? If so, we can
suggest that set `evil-overriding-maps` to `'()` as `evil-want-keybinding`.

jojojames:
> Do you mean using evil-overriding-maps is not recommended? 

Yes.

> If so, we can
suggest users set evil-overriding-maps to '() as evil-want-keybinding.

Maybe that's not a bad idea, let me think on this further.

jojojames:
I think it may be worth looking at where evil uses evil-overriding-maps and see if we can maybe PR something where -evil-want-binding- gates those maps from being set. That's already what we're doing with the evil-keybindings file in evil.

e.g. https://github.com/emacs-evil/evil/blob/5c28294d830a5a79e9b9da2c32e7675d52d76720/evil.el#L141

condy0919:
`evil-overriding-maps` defaults to `'()` now, changed since https://github.com/emacs-evil/evil/pull/1494

jojojames:
Is there something to be done here @condy0919 ? I think we don't want to support this (evil-collection respecting evil-overriding-maps)?

condy0919:
> Is there something to be done here @condy0919 ? I think we don't want to support this (evil-collection respecting evil-overriding-maps)?

Yes.

But we could prompt a diagnostic message to encourage user to set `evil-overriding-maps` to `nil` (it's now the default) to reduce redundant keybindings (e.g. h/j/k/l, w/b/W/E).

``` emacs-lisp
(unless evil-overriding-maps
  (message "Make sure to set `evil-overriding-maps' to nil before loading evil \
or evil-collection."))
```

jojojames:
It's probably OK to not prompt them, maybe a documentation change is all we need.

If we do prompt them, we'd also need a way to disable the prompt for the users already aware but don't care for the message.

condy0919:
> It's probably OK to not prompt them, maybe a documentation change is all we need.
>
> If we do prompt them, we'd also need a way to disable the prompt for the users already aware but don't care for the message.

Ok. I will raise an another PR later.

PR #507, opened 2021-07-17
Modify telega keybinding for copy and reply

Vonfry:
1. yy and yl is conflicted with the evil yank commands.
2. Reply is more common than forword and r is reply in default of telega.

Advice from @cnsunyour @yangsheng6810 and do you have more suggestions for improvement?


condy0919:
Sorry, I'm not a telega user. :'(

PR #512, opened 2021-08-05
Wgrep support for grep for other package managers

daanturo:
Support is mentioned in the README but isn't implemented for the vanilla `grep-mode`.

### DIFF
--- a/modes/grep/evil-collection-grep.el
+++ b/modes/grep/evil-collection-grep.el
@@ -41,7 +41,7 @@
     "\C-k" 'previous-error-no-select)
 
   ;; `wgrep' integration
-  (when (featurep 'wgrep)
+  (when (fboundp 'wgrep-setup)
     (evil-collection-define-key 'normal 'grep-mode-map
       "i" 'wgrep-change-to-wgrep-mode)))
 


condy0919:
It's here https://github.com/emacs-evil/evil-collection/blob/master/modes/grep/evil-collection-grep.el#L44

daanturo:
But `featurep` isn't usable for `straight.el`, I didn't know that it was configured since `(featurep 'wgrep)` returns `nil` on Doom Emacs, which makes use of `straight`.

condy0919:
Maybe you're the victim of lazy load.

``` elisp
(use-package wgrep
  :ensure t
  :demand t)
```

daanturo:
```diff
   ;; `wgrep' integration
-  (when (featurep 'wgrep)
+  (when (fboundp 'wgrep-setup)
     (evil-collection-define-key 'normal 'grep-mode-map
       "i" 'wgrep-change-to-wgrep-mode)))
```

Can we at least do this instead? It's guaranteed to work across package managers since `wgrep-setup` is autoloaded.

condy0919:
That's maybe better.

jojojames:
> But featurep isn't usable for straight.el, I didn't know that it was configured since (featurep 'wgrep) returns nil on Doom Emacs, which makes use of straight.

I'm not super familiar with the subtle tradeoffs betwen the two (in terms of impact) but maybe this is setting a bad precedence...

daanturo:
> I'm not super familiar with the subtle tradeoffs betwen the two (in terms of impact) but maybe this is setting a bad precedence...

We can get rid of `featurep` and `fboundp` by simply move to `evil-collection-wgrep.el`.
[Commit](https://github.com/daanturo/evil-collection/commit/764a98cd6c39957694f3ccf3ea44571d58c83779)

condy0919:
Is it the right approach to third library integration?

PR #526, opened 2021-08-31
Change depricated cask installation method

meain:
This changes the deprecated cask install method to the new suggested method.


condy0919:
Minor change to follow https://github.com/marketplace/actions/setup-cask

PR #529, opened 2021-09-07
mu4e: update binding for attachment actions

afontaine:
As part of the move to the gnus viewer for emails, the command name has been
changed from `mu4e-view-attachment-action` to `mu4e-view-mime-part-actions`.

https://github.com/djcb/mu/issues/2101#issuecomment-903074743


condy0919:
Thanks

PR #530, opened 2021-09-11
selectrum: add initial support

balajisivaraman:
Hey,

First of all, thanks for the truly wonderful package. My jump to Evil Mode (again!) wouldn't be complete without it.

In this PR, I have added basic support for [Selectrum](https://github.com/raxod502/selectrum), which I found to be lacking here.

I have tried my best to mimic either Helm or Ivy keybindings with this PR. I don't see `C-j` and `C-k` being bound for navigating between candidates in Ivy or Helm. However, I have found it to be useful. I can remove this if it goes against normal Evil Collection semantics.

Open to any other suggestions or changes. Thanks again.

### DIFF
--- a/modes/selectrum/evil-collection-selectrum.el
+++ b/modes/selectrum/evil-collection-selectrum.el
@@ -1,0 +1,67 @@
+;;; evil-collection-selectrum.el --- Evil bindings for Selectrum -*- lexical-binding: t -*-
+
+;; Copyright (C) 2021 Balaji Sivaraman
+
+;; Author: Balaji Sivaraman <balaji@balajisivaraman.com>
+;; Maintainer: Balaji Sivaraman <balaji@balajisivaraman.com>
+;; Pierre Neidhardt <mail@ambrevar.xyz>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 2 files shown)


condy0919:
It should take account of `evil-collection-setup-minibuffer`.

balajisivaraman:
@condy0919, Thanks! I had originally missed that. I have now added all the keybindings behind `evil-collection-setup-minibuffer`.

modes/selectrum/evil-collection-selectrum.el
```diff
@@ -0,0 +1,67 @@
...
+;;;###autoload
+(defun evil-collection-selectrum-setup ()
+  "Set up `evil' bindings for `selectrum-mode'."
+
+  (defvar evil-collection-setup-minibuffer)
+  (when evil-collection-setup-minibuffer
+    (evil-collection-define-key '(insert normal) 'selectrum-minibuffer-map
+      (kbd "C-f") 'selectrum-previous-page
+      (kbd "C-m") 'selectrum-submit-exact-input
+      (kbd "C-v") 'selectrum-next-page
```
condy0919, line 45: `C-f` -> `selectrum-next-page`
`C-b` -> `selectrum-previous-page`

condy0919 (changes requested):

condy0919:
IIRC, `C-m` is the same as `RET` (the logical key), and Emacs will automatically translate `<return>` to `RET`.

balajisivaraman:
@condy0919, I have now fixed both. I have setup `M-RET` for selecting the current minibuffer input. Looking around all the other configs, I could not find one that exactly matched with this particular use-case, so I went with this, since it sort of made sense to me.

condy0919:
I found https://github.com/minad/vertico/ also recommend `M-RET`.

PR #532, opened 2021-09-12
comint: Bind comint-kill-input to C-u

leungbk:
In Vim insert mode, C-u deletes to the beginning of the line, so we
make an analogous binding for comint buffers.


condy0919:
Thanks

PR #537, opened 2021-09-22
beginend: evil-collection-beginend-maps -> [...]-modes, and iterate

leungbk:
I noticed that on the master branch, the dired-mode stuff wasn't working since it hadn't been manually typed into the setup function. So I think we should simply iterate over the defconst instead.

cc @balajisivaraman


balajisivaraman:
Ah, thank you @leungbk. I was trying to achieve something like this originally with my limited ELisp knowledge. Let me go through this and understand further. :)

condy0919:
Great work!

PR #544, opened 2021-10-14
use esc to quit vertico

tshu-w:


condy0919:
No objection from me.

PR #548, opened 2021-11-05
Adding snake support

wickedjargon:
adding support for snake

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -254,6 +254,7 @@ This will bind additional find-* type commands, e.g. usages, assignments, etc.."
     telega
     (term term ansi-term multi-term)
     tetris
+    snake
     ,@(when (>= emacs-major-version 27) '(thread))
     tide
     timer-list
--- a/modes/snake/evil-collection-snake.el
+++ b/modes/snake/evil-collection-snake.el
@@ -1,9 +1,10 @@
-;;; evil-collection-tetris.el --- Bindings for `tetris' -*- lexical-binding: t -*-
+;;; evil-collection-snake.el --- Bindings for `snake' -*- lexical-binding: t -*-
 
 ;; Copyright (C) 2021 Farzin Firouzi
 
 ;; Author: Farzin Firouzi
-;; Maintainer: Farzin Firouzi <farzinf@riseup.net>
+;; Maintainer: Farzin Firouzi
+;; Farzin Firouzi <farzinf@riseup.net>
 ;; URL: https://github.com/emacs-evil/evil-collection
 ;; Version: 0.0.1
 ;; Package-Requires: ((emacs "25.1"))
@@ -50,4 +51,4 @@
     (kbd "<escape>") 'quit-window))
 
 (provide 'evil-collection-snake)
-;;; evil-collection-snake.el ends here
+;;; evil-collection-tetris.el ends here
--- a/modes/snake/evil-collection-snake.el
+++ b/modes/snake/evil-collection-snake.el
@@ -28,7 +28,7 @@
 
 ;;; Code:
 (require 'evil-collection)
-(require 'snake nil t)
+(require 'snake)
 
 (defconst evil-collection-snake-maps '(snake-mode-map))
 
--- a/modes/snake/evil-collection-snake.el
+++ b/modes/snake/evil-collection-snake.el
@@ -1,0 +1,53 @@
+;;; evil-collection-tetris.el --- Bindings for `tetris' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2021 Farzin Firouzi
+
+;; Author: Farzin Firouzi
+;; Maintainer: Farzin Firouzi <farzinf@riseup.net>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
+;; Package-Requires: ((emacs "25.1"))
+;; Keywords: evil, emacs, tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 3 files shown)


modes/snake/evil-collection-snake.el
```diff
@@ -0,0 +1,54 @@
...
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Bindings for `snake'.
+
+;;; Code:
+(require 'evil-collection)
+(require 'snake nil t)
```
condy0919, line 31: `snake` is a builtin package, the `FILENAME` and `NOERROR` parameters can be omitted..

modes/snake/evil-collection-snake.el
```diff
@@ -0,0 +1,54 @@
...
+
+(defconst evil-collection-snake-maps '(snake-mode-map))
+
+;;;###autoload
+(defun evil-collection-snake-setup ()
+  "Set up `evil' bindings for `snake'."
```
condy0919, line 37: `(evil-collection-set-readonly-bindings 'snake-mode-map)` and the `q` and `ZZ` can be omitted.

modes/snake/evil-collection-snake.el
```diff
@@ -0,0 +1,54 @@
...
+    "j" 'snake-move-down
+    "k" 'snake-move-up
+    "l" 'snake-move-right
+
+    "gr" 'snake-start-game
+    "p" 'snake-pause-game
+    "q" 'quit-window
+    "ZQ" 'quit-window
+    "ZZ" 'snake-end-game
+    (kbd "<escape>") 'quit-window))
```
condy0919, line 51: The original `snake-mode-map` has no `ESC` binding, is there any reason to add an evil binding?

modes/snake/evil-collection-snake.el
```diff
@@ -0,0 +1,54 @@
...
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+;; Bindings for `snake'.
+
+;;; Code:
+(require 'evil-collection)
+(require 'snake nil t)
+
+(defconst evil-collection-snake-maps '(snake-mode-map))
```
condy0919, line 33: `snake-null-map` is missing.

condy0919 (changes requested):

modes/snake/evil-collection-snake.el
wickedjargon: @condy0919 copied the tetris dir and modified the appropriate files. Is there a reason why `(require 'tetris nil t)` is included in the tetris dir?

wickedjargon: Also, can you please provide the steps, or provide a resource I can use to test this code? This is my first pull request. I know the steps for evil-mode is to `cd` into the dir and run `make emacs`. How do I do this for evil-collection? I noticed a make file. Do I use that?

condy0919: > Also, can you please provide the steps, or provide a resource I can use to test this code? This is my first pull request. I know the steps for evil-mode is to `cd` into the dir and run `make emacs`. How do I do this for evil-collection? I noticed a make file. Do I use that?

Manually <kbd>M-:</kbd> `(evil-collection-snake-setup)`

condy0919: > @condy0919 copied the tetris dir and modified the appropriate files. Is there a reason why `(require 'tetris nil t)` is included in the tetris dir? here: https://github.com/emacs-evil/evil-collection/blob/master/modes/tetris/evil-collection-tetris.el

It should also be fixed.

modes/snake/evil-collection-snake.el
wickedjargon: Yes there is. The reason is to bind `ESC` now is bound to `quit-window`, similar `evil-collection/modes/tetris/evil-collection-tetris.el`, as shown here: https://github.com/emacs-evil/evil-collection/blob/master/modes/tetris/evil-collection-tetris.el

modes/snake/evil-collection-snake.el
wickedjargon: In checking other builtin packages other than `tetris` (example: dired, eww, and virtually all of them) it appears they too have a line equivalent to `line 30 and 31` above, despite it being a builtin package. I understand you're point, and while I am happy to go forward with changing all files in `/modes/` that are related to built-ins, it would be a good idea to check with others to see if this is a good idea. Does `(require '...)` even take that much time to eval/parse?

condy0919: > while I am happy to go forward with changing all files in /modes/ that are related to built-ins

It seems like only `tetris` is the one should be fixed.

> Does (require '...) even take that much time to eval/parse?

It depends. If you found all commands used in `evil-collection-xxx.el` are `autoload`ed, then the `(require '...)` should be removed for better user experience.

e.g. https://github.com/emacs-evil/evil-collection/blob/master/modes/consult/evil-collection-consult.el#L37

modes/snake/evil-collection-snake.el
condy0919: OK.

evil-collection.el
```diff
@@ -256,6 +256,7 @@ This will bind additional find-* type commands, e.g. usages, assignments, etc.."
     telega
     (term term ansi-term multi-term)
     tetris
+    snake
```
condy0919, line 259: Make the list ordered.

``` diff
simple
sly
+snake
speedbar
```

condy0919 (changes requested):

PR #550, opened 2021-11-07
Adds goto next and previous flymake bindings

jgarte:
WIP: I haven't tested these yet.

Testing help appreciated.

### DIFF
--- a/modes/flymake/evil-collection-flymake.el
+++ b/modes/flymake/evil-collection-flymake.el
@@ -39,6 +39,8 @@
   (evil-collection-define-key
     '(normal visual) 'flymake-diagnostics-buffer-mode-map
     "q" 'quit-window
+    (kbd "C-j") 'flymake-goto-next-error
+    (kbd "C-k") 'flymake-goto-prev-error
     (kbd "RET") 'flymake-goto-diagnostic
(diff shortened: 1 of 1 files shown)


condy0919:
If you use `unimpaired`, it can be done via `[l` and `]l`

jgarte:
> If you use `unimpaired`, it can be done via `[l` and `]l`

@condy0919 You mean this package? 

https://github.com/zmaas/evil-unimpaired

condy0919:
> > If you use `unimpaired`, it can be done via `[l` and `]l`
> 
> @condy0919 You mean this package?
> 
> https://github.com/zmaas/evil-unimpaired

https://github.com/emacs-evil/evil-collection/blob/master/modes/unimpaired/evil-collection-unimpaired.el#L240-L241

No, it's bundled with evil-collection.

jojojames:
C-j and C-k seem like they'll clash with a lot of common binds since flymake is a more "global" minor mode during editing, imo.

condy0919:
> C-j and C-k seem like they'll clash with a lot of common binds since flymake is a more "global" minor mode during editing, imo.

It defined keys in the diagnostic buffer.

PR #553, opened 2021-11-11
Mu4e: Fix name of command for saving attachments

### LINKED ISSUE #523: wrong type arguement: commandp; mu4e-view-save-attachment
mu4e-view mode is showing **mu4e-view-save-attachments** (extra s) as an option to save attachment instead of **mu4e-view-save-attachment**.

extheo:
Fixes #523


condy0919:
Nice catch.

PR #557, opened 2021-12-06
consult: implement jump-list command

northgaard:
This PR implements a consult command which allows you to browse the evil jump list with consult preview and completion.

In a separate commit, I've also simplified the implementation of evil-collection-consult-mark a bit by removing unnecessary usage of advice. This can be dropped or moved to a separate PR depending on maintainers' preference.

Thanks in advance.


condy0919:
Thanks

Issue #543, opened 2021-10-10
Remap of magit-submodule suffix in magit-dispatch transient prefix causes forge to print an error


sin-ack:
Hi. This line:
https://github.com/emacs-evil/evil-collection/blob/6709c1ec4118c8721df43ea6708ae45ebbc01fd3/modes/magit/evil-collection-magit.el#L574
remaps `magit-submodule`'s key in the `magit-dispatch` transient prefix from `o` to `'`. Since magit/forge@9fbd00e, forge will now print an error when a Magit buffer is first visited:
```
Cannot insert ("N" "Forge" forge-dispatch) into magit-dispatch; o not found
```
as it cannot find the `o` key to insert its own key before. Not quite sure how to approach fixing this, because Forge can technically do that just fine, but remapping the keys should be fine too.

jojojames:
Hmnn.. this one is tricky since it seems like those changes are at the top level in forge.el. Any suggestions @sin-ack ?

tshu-w:
This only happens if we run `(evil-collection-magit-setup)` before `(require 'forge)`, we should `(require 'forge)` before `(evil-collection-magit-init)`:
```emacs-lisp
  (defun evil-collection-magit-setup@override ()
    "Set up `evil' bindings for `magit'."

    (evil-collection-define-key 'normal 'magit-blame-mode-map
      "q" 'magit-blame-quit)
    (evil-collection-define-key 'normal 'magit-blame-read-only-mode-map
      "q" 'magit-blame-quit)

    (require 'forge)
    (evil-collection-magit-init))
```
or seperate forge setup from magit in evil-collections.

jojojames:
We don't want to forcer equire forge as not everyone uses it (me included).

jojojames:
Thanks @tarsius ! I made some initial changes here (https://github.com/emacs-evil/evil-collection/pull/549/files) and it looks okay to me but I don't really use forge so not sure if anything is missing. @sin-ack Can you pull @tarsius 's changes for forge on his no-defaults branch and evil-collection's forge branch and see if it works for you?

If not we can probably go with the current set of changes on both branches.

tshu-w:
> Can you pull @tarsius 's changes for forge on his no-defaults branch and evil-collection's forge branch and see if it works for you?

This works for me.

tarsius:
I have merged the new variables into the master branch.

[...]

PR #563, opened 2021-12-13
Prevent forge from binding default keys

justbur:
evil-collection handles this, and the default bindings produce an error message
on startup.

### DIFF
--- a/modes/forge/evil-collection-forge.el
+++ b/modes/forge/evil-collection-forge.el
@@ -44,9 +44,15 @@
   :group 'magit
   :type  'symbol)
 
+(defvar forge-add-default-bindings)
+
 ;;;###autoload
 (defun evil-collection-forge-setup ()
   "Set up `evil' bindings for `magit'."
+  (when forge-add-default-bindings
+    (message "Setting `forge-add-default-bindings' to nil in `evil-collection-forge-setup'.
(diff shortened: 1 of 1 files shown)


jojojames:
There are some comments @tarsius left behind on https://github.com/emacs-evil/evil-collection/issues/543 that I haven't had a chance to look through. Do we know if setting this here definitely skips the binding in Forge?

justbur:
Yes it should. forge.el is a little odd in that it uses `provide` triggering `eval-after-load` before it makes the bindings.

jojojames:
Sounds good.

modes/forge/evil-collection-forge.el
```diff
@@ -44,9 +44,15 @@
   :group 'magit
   :type  'symbol)
 
+(defvar forge-add-default-bindings)
+
 ;;;###autoload
 (defun evil-collection-forge-setup ()
   "Set up `evil' bindings for `magit'."
+  (when forge-add-default-bindings
```
luishgh, line 52: I think this should be changed to `(when (boundp 'forge-add-default-bindings)`, as [the latest release tag of forge](https://github.com/magit/forge/releases/tag/v0.3.0) does not include this variable yet.

condy0919: Agreed.

PR #566, opened 2021-12-16
unimpaired: turn functions into proper motions

dominikh:
Use evil-define-motion macro, which sets :motion repeat, set :jump t,
and add support for counts.


condy0919:
Thanks

PR #567, opened 2021-12-17
Add ebib-jump-to-entry

Vonfry:


modes/ebib/evil-collection-ebib.el
```diff
@@ -47,6 +47,7 @@
     "k" 'ebib-prev-entry
     "n" 'ebib-search-next
     "p" 'ebib-yank-entry
+    "gj" 'ebib-jump-to-entry
```
condy0919, line 50: IMHO, `J` is better than `gj`.

condy0919 (changes requested):

modes/ebib/evil-collection-ebib.el
Vonfry: `J` has been used by ebib for jump to other database.

condy0919: There is no `J` binding defined in `ebib-index-mode-map` and `ebib-entry-mode-map`.

Vonfry: [https://github.com/joostkremers/ebib/blob/37b9c4ec7a57ffdda53f3345cc99cf6819e94863/ebib.el#L966](source)

condy0919: OK

Issue #559, opened 2021-12-09
Feature Request: mpc


jgarte:
https://pspiagicw.github.io/posts/the-weirdest-mode-in-emacs-mpc-mode/

Keys mentioned in article:

```
(general-define-key
 :keymaps 'mpc-mode-map
 :states 'normal
 "j" 'move-mpc-down
 "k" 'move-mpc-up
 "t" 'mpc-toggle-play
 "r" 'mpc-toggle-repeat
 "s" 'mpc-toggle-shuffle
 "S" 'mpc-toggle-shuffle
 "c" 'mpc-toggle-consume
 "a" 'mpc-playlist-add
 "p" 'mpc-playlist
 ">" 'mpc-next
 "<" 'mpc-prev
 "R" 'mpc-playlist-delete
 "RET" 'mpc-select
 "x" 'mpc-play-at-point)
```

pspiagicw:
Restatement of #535. Should close one of them.

pspiagicw:
Can be closed now , PR merged

Issue #535, opened 2021-09-15
Support for inbuilt MPC mode.


pspiagicw:
Emacs has a inbuilt mode for controlling MPD (Music Player Daemon). Alternative to other packages (simple-mpc , mpdel ). It would make life a lot easier if evil-collection can support mpc mode  https://github.com/emacs-mirror/emacs/blob/master/lisp/mpc.el

pspiagicw:
Can be closed. PR merged

PR #568, opened 2021-12-18
Mpc mode

pspiagicw:
PR introduces evil-bindings for MPC mode a front-end to the MPD Music Server. Referenced in #535 and #559. It uses 2 custom function
`evil-collection-move-mpc-up` and `evil-collection-move-mpc-down`.

Missing features
- Volume control (As there is no Emacs way of doing it currently, only way is to use the `mpc` client with `async-shell-command`)

### DIFF
--- a/modes/mpc/evil-collection-mpc.el
+++ b/modes/mpc/evil-collection-mpc.el
@@ -1,0 +1,61 @@
+;;; evil-collection-mpc.el --- Bindings for `mpc-mode' -*- lexical-binding: t -*-
+;; Copyright (C) 2021 pspiagicw
+
+;; Author: pspiagicw <pspiagicw@gmail.com>
+;; Maintainer: pspiagicw <pspiagicw@gmail.com>
+;; pspiagicw <pspiagicw@gmail.com>
+;; URL: https://github.com/emacs-evil/evil-collection
+;; Version: 0.0.1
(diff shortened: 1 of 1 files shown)


modes/mpc/evil-collection-mpc.el
```diff
@@ -0,0 +1,60 @@
...
+  "Move the cursor up along with selecting the element"
+  (evil-previous-visual-line)
+  (mpc-select))
+(defconst evil-collection-mpc-mode-maps '(mpc-mode-map))
+
+;;;###autoload
+(defun evil-collection-mpc-mode-setup ()
+  "Setup up 'evil' bindings for 'mpc-mode'"
+  (evil-collection-define-key 'normal 'mpc-mode-map
+    "j" 'evil-collection-move-mpc-down
```
condy0919, line 46: `j` -> `C-j`, `k` -> `C-k`. Keeping it consistent with other modules, e.g. https://github.com/emacs-evil/evil-collection/blob/master/modes/compile/evil-collection-compile.el#L54

condy0919 (changes requested):

modes/mpc/evil-collection-mpc.el
pspiagicw: Done

PR #572, opened 2021-12-27
notmuch: use notmuch-tree-archive-thread-then-next in tree view

dominikh:
Default notmuch keymaps bind A to
notmuch-tree-archive-thread-then-next in both notmuch-show-mode and
notmuch-tree-mode.


condy0919:
Thanks

PR #575, opened 2022-01-04
Fix C-u scroll in magit config

xeruf:
This little mistake broke the native magit `C` binding because the keybinding was now `C - u` rather than `C-u` ^^


condy0919:
Good catch.

PR #580, opened 2022-01-13
Avoid binding escape to bury-buffer and quit-window

Stebalien:
Escape usually means to:

1. Return to normal mode.
2. Abort the minibuffer.

By binding it to quit-window (or equivalent), hitting escape a few times to exit a (potentially recursive) minibuffer, etc. can cause buffers to vanish.

It's also:

1. Inconsistently applied. These are the only two modes that have escape bound this way.
2. Redundant with "q" (which is consistently applied).


condy0919:
According to the rationale:

> Don’t bind anything to `:` nor `<escape>`.

You're right.

condy0919:
cc @jojojames Unbinding `<escape>` from magit is OK?

jojojames:
What's the new behavior without bury-buffer?

condy0919:
> What's the new behavior without bury-buffer?

Manually press `q` to quit magit buffer (C-x g).

condy0919:
If no strong objection, I will merge it in 2 weeks. @jojojames

PR #581, opened 2022-01-17
Set readonly bindings for compilation mode

mcfarljm:
Same idea that is used in, e.g., `evil-collection-help.el`.  In particular, it binds `q` to `quit-window`.


condy0919:
Thanks

PR #582, opened 2022-01-18
Use default v/V for magit revert action

### LINKED ISSUE #571: Bug: can't set flags on magit revert action
For magit, the revert action is mapped to `_`/`-`.

However, `-` is how you set flags. The result is that you are unable to set flags on the `revert` action because pressing `-` triggers the action instead of setting a flag.

Screen Shot 2021-12-21 at 11 07 04 AM

micahbf:
The `-` binding conflicts with setting arguments.

Fixes #571.

### DIFF
--- a/modes/magit/README.org
+++ b/modes/magit/README.org
@@ -11,7 +11,6 @@
 
 * Recent Changes (most recent first)
 
-  1. [2019-01-18] Use default =v= and =V= bindings for revert action
   1. [2019-09-04] Don't use =evil-next-visual-line= and
      =evil=previous-visual-line=. See [[https://github.com/emacs-evil/evil-magit/issues/70][Issue #70]].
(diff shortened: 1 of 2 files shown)


condy0919:
I cherry pick your changes. Removing 

```
-     (magit-revert "v" "-" magit-revert-no-commit)
```

is enough.

1. <kbd>C-x g</kbd> launch magit
2. <kbd>-</kbd> invoke magit-revert
3. <kbd>v</kbd> revert changes

In `magit-revert`, `_` is retained to revert commits.

micahbf:
Thank you. This does fix the issue, but FWIW I think it does make sense to also revert `_`, since magit generally has linked pairs of lowercase and uppercase letters that have similar functions with different levels of destructiveness.

PR #584, opened 2022-01-30
Bind arrow keys in Doc View mode.

okamsn:
I found that these keys were bound to the normal character/line Evil motions in normal state. This changes them to have the same bindings that they have in the Emacs state.

### DIFF
--- a/modes/doc-view/evil-collection-doc-view.el
+++ b/modes/doc-view/evil-collection-doc-view.el
@@ -38,13 +38,15 @@
   (evil-set-initial-state 'doc-view-mode 'normal)
   (evil-collection-define-key 'normal 'doc-view-mode-map
     "q" 'quit-window
+    [remap evil-next-line] 'doc-view-next-line-or-next-page
+    [remap evil-previous-line] 'doc-view-previous-line-or-previous-page
+    [remap evil-backward-char] 'image-backward-hscroll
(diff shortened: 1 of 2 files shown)


condy0919:
Remapping `evil-next-line` and `evil-previous-line` looks good, and write the function name  `doc-view-previous-line-or-previous-page` only once.

okamsn:
I've made the changes.

condy0919:
Seems like `evil-collection-image.el` (doc-view mode uses images to display pdf) should also use `[remap evil-next-line]`. Since the precedence of `[remap evil-next-line]` is lower than explicit `"j"`, pressing `"j"` invokes `image-next-line`.

okamsn:
I've added the re-mappings to the Image mode. The commands `image-forward-hscroll` and `image-backward-hscroll` were not bound in  the Doc View mode settings. They are remapped in these changes in the Doc View settings and now in the Image mode settings too.

Do you want the commands to only be remapped in the file for Image mode, since you say that affects Doc View mode, or would you like to list them in both for explicitness?

condy0919:
> Do you want the commands to only be remapped in the file for Image mode, since you say that affects Doc View mode, or would you like to list them in both for explicitness?

`image-mode` is the parent mode of `doc-view-mode`, IMHO, explicitly binding looks good.

Issue #224, opened 2019-02-02
Define a consistent binding for next-error and previous-error


hammerandtongs:
The next-error and previous-error commands work reasonably well (and are very handy) with flycheck working.

What should be the consistent keybind policy from evil-collection?

jojojames:
I've been using 

```
  (define-key flymake-mode-map (kbd "M-n") 'flymake-goto-next-error)
  (define-key flymake-mode-map (kbd "M-p") 'flymake-goto-prev-error)
```
for flymake. Would be open to something consistent across packages or just leave it to the user too.

hammerandtongs:
So it turns out 

`:cfirst
:cnext
:cprevious`

All work right with Evil and this is the Vim method these days I think.

I was hoping to find something consistent in the Vim world that evil-collection could use and this seems to be it?

jojojames:
Do you have further suggestions @hammerandtongs ?

hammerandtongs:
:cnext has been more comfortable for me then M-x next-error so I won't move into a keybinding.

This does call back to #64 

evil-unimpaired seems like a sensible way forward, is this viable these days? 

It's seems pretty limited compared to Tim Pope's version but has a lot of potential.

As I look at the issue it seems like adopting evil-unimpaired into evil-collection more generally might solve this issue and have some nice mappings all around... unimpaired is adopted by the Pope of Vim afterall ;)

jojojames:
Added something unimpaired-like for this. I added two basic functionalities to it (next-buffer and next-error). I'm not married to next-error's binding so feel free to PR if you think a better one exists.

I'd like people to improve that file (adding more appropriate bindings/functions) with PRs though since I'm not too versed in the usecases of unimpaired specific keys unimpaired uses. @bling @hammerandtongs @ajgrf please improve it if you see fit!

PR #589, opened 2022-02-05
unimpaired: add vanilla next-error navigation

aidan-hall:

Emacs comes with built-in error jumping commands:
`(next-error)` and `(first-error)` are the Emacs equivalents of `:cnext` and `:cfirst` in Vim, which jump between errors in the `*compilation*` buffer, among other things.

They can move between buffers, so the behaviour might be unpredictable.

https://github.com/aidan-hall/evil-collection

### Checklist

- [x] byte-compiles cleanly
- [x] `M-x checkdoc` is happy.

### DIFF
--- a/modes/unimpaired/evil-collection-unimpaired.el
+++ b/modes/unimpaired/evil-collection-unimpaired.el
@@ -68,7 +68,7 @@
          (fboundp 'flymake-goto-next-error))
     (flymake-goto-next-error count))
    (:default
-    (message "No linting modes are on."))))
+    (next-error count))))
 
 (evil-define-motion evil-collection-unimpaired-previous-error (count)
   "Go to previous error."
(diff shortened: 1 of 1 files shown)


condy0919:
https://github.com/hlissner/doom-emacs/blob/master/modules/editor/evil/config.el#L472

doom use `next-error` and `previous-error`, regardless of modes. Can we do the same in evil-collection?

aidan-hall:
While flycheck does support next-error/previous-error, flymake doesn't.
Personally, I barely ever use flymake, so hard-binding the keys like that would make more sense for me.
It looks like this has been discussed before... https://github.com/emacs-evil/evil-collection/issues/224

I'll consider suggesting a more radically simplified version at a later point.

condy0919:
Thanks, bro

PR #595, opened 2022-02-20
Make ibuffer bindings consistent with dired

Silex:
Hello,

1. I find it frustrating that `t` behaves differently between `dired` and `ibuffer`.
2. Ibuffer `* *` is the same as `U`, it's more interesting for it to mean `mark special buffers` (the temporary ones, which start or end with `*`). That way when we want to get rid of all these `*` buffer the mnemonic is easy: press `* * D RET`


condy0919:
Agreed.

PR #596, opened 2022-02-21
mu4e: use newly-renamed functions as aliases when appropriate

leungbk:
This accommodates everyone using a version of mu4e shipped with their
package manager's installation of mu.


leungbk:
So I'm not sure how best to comply with our linting process here. Should we just copy-paste the `defun`s for these two functions? That approach wouldn't fail the linting.

condy0919:
May a wrapper function be more proper.

``` emacs-lisp
(defun evil-collection-mu4e-
```

condy0919:
Thanks

PR #599, opened 2022-03-02
Add hint on passing a list to evil-collection-init

jgarte:
This adds a hint in the readme on passing a list to `evil-collection-init`.


jojojames:
Thanks!

PR #604, opened 2022-03-14
Use transient's definition macros by their new names

tarsius:
See https://github.com/magit/transient/issues/192.


jojojames:
Thanks @tarsius !

PR #606, opened 2022-04-01
enhance embark-collect-mode

tshu-w:
add more keybindings to embark-collect-mode.

Most of the keybindings are consistent with the embark-collect default and dired.


tshu-w:
@condy0919 Sorry I didn't get why the test failed, can you help me look at it?

PR #612, opened 2022-04-05
magit: fix stage not working in visual mode after 65b43.

### LINKED ISSUE #611: Bug: magit stage from visual state broken since magit@38cb9
Magit [has removed simple stage and unstage bindings](https://github.com/magit/magit/commit/38cb9372dc51df4767b30181db93f51f7ac856ab#diff-335d97ddde0824a60d53f17c711216f33e2a27860169fc2e02763781e2ae4779L1962-L1963)  in favour of a new menu system, specifically [here](https://github.com/magit/magit/commit/38cb9372dc51df4767b30181db93f51f7ac856ab#diff-335d97ddde0824a60d53f17c711216f33e2a27860169fc2e02763781e2ae4779L1962-L1963).  This breaks staging from visual mode in magit buffers.

- `\s` (i.e. emacs-state and then `s`) still works
- I've tried editing `evil-collection-magit.el` to explicitly add the binding back in for visual state, but I'm doing something wrong and can't seem to remap anything at all.

Someone who knows how the rebinding actually works can probably just add visual-state bindings for `s` and `u` explicitly.

2e0byo:
fixes  #611.

I'm not sure if this is the best place to set this binding.

### DIFF
--- a/modes/magit/evil-collection-magit.el
+++ b/modes/magit/evil-collection-magit.el
@@ -305,6 +305,7 @@ When this option is enabled, the stash popup is available on \"Z\"."
        (,states magit-mode-map "\C-b"   evil-scroll-page-up)
        (,states magit-mode-map ":"     evil-ex)
        (,states magit-mode-map "q"     magit-mode-bury-buffer)
+       ((visual) magit-hunk-section-map "s" magit-stage)
 
        ;; these are to fix the priority of the log mode map and the magit mode map
        ;; FIXME: Conflict between this and revert. Revert seems more important here


condy0919:
As the magit link tells, `magit-stage` was removed from `magit-diff-section-map`. It's still available in `magit-hunk-section-map`.

2e0byo:
> As the magit link tells, `magit-stage` was removed from `magit-diff-section-map`. It's still available in `magit-hunk-section-map`.

Sorry, I'm confused: I had to explicitly bind `magit-stage` to `s` in visual mode in `magit-hunk-section-map` otherwise it didn't exist.  Should I have bound it elsewhere?

condy0919:
> Sorry, I'm confused: I had to explicitly bind `magit-stage` to `s` in visual mode in `magit-hunk-section-map` otherwise it didn't exist. Should I have bound it elsewhere?

Could you provide a recipe for reproducing? e.g.

```elisp
(require 'package)
(setq package-archives
      '(("melpa"  . "https://melpa.org/packages/")
        ("gnu"    . "https://elpa.gnu.org/packages/")
        ("nongnu" . "https://elpa.nongnu.org/nongnu/")))

(package-initialize)

;; Bootstrap `use-package'
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))
(eval-and-compile
  (setq use-package-always-ensure nil)
  (setq use-package-always-defer nil)
  (setq use-package-always-demand nil)
  (setq use-package-expand-minimally nil)
  (setq use-package-enable-imenu-support t))
(eval-when-compile
  (require 'use-package))

(use-package evil
  :ensure t
  :hook (after-init . evil-mode)
  :custom
  (evil-want-integration t)
  (evil-want-keybinding nil))

(use-package evil-collection
  :ensure t
  :hook (evil-mode . evil-collection-init))

(use-package magit
  :ensure t)
```

and the following operations.

BTW, you may have interests in https://github.com/syl20bnr/spacemacs/issues/15448

2e0byo:
Bother, I'd looked but managed to miss a bug on spacemacs and determined (erroneously) that this was a problem with `org-collection` (specifically I thought---but didn't test---that `evil-surround-mode` was invoked by vanilla evil).

Confirm this isn't reproducible with vanilla evil, and is thus a spacemacs bug.  Moreover it's now fixed in spacemacs, although that's moot for here.

Thanks for your time.

PR #613, opened 2022-04-11
ebib: change mapping for "y" in ebib-entry-mode.

ahnafalrafi:
Change mapping for "y" in ebib-entry-mode to "ebib-copy-current-field-contents" instead of "ebib-copy-field-contents" since the latter is not a command.


jojojames:
Thanks!

Issue #602, opened 2022-03-06
Key translation doesn't work in eshell


rosetail:
`evil-collection-translate-key` does nothing in eshell-mode-map. I think this is related to #433, as I am affected by that issue too. I should note though that `evil-collection-translate-key` works just fine with comint-mode.

Here is the minimal config I used to confirm the issue:
```lisp
(require 'package)
(package-initialize)

(setq evil-want-keybinding nil)
(require 'evil)
(evil-mode 1)
(require 'evil-collection)
(defun my-hjkl-rotation (_mode mode-keymaps &rest _rest)
  (evil-collection-translate-key 'normal mode-keymaps
    (kbd "C-n") (kbd "C-j")
    (kbd "C-e") (kbd "C-k")
    "gn" "gj"
    "ge" "gk"))
(add-hook 'evil-collection-setup-hook #'my-hjkl-rotation)
(evil-collection-init)
```

condy0919:
Yes, confirmed.

condy0919:
`evil-collection-eshell-mode-map-normal-state-backup-map` is nil, `lookup-key` on it will always be failed. Let me digger.

rosetail:
I just noticed that translations don't work for emms either

condy0919:
....

jojojames:
fwiw, this works for me:

```
(evil-collection-translate-key 'normal 'eshell-mode-map
    "gj" "gk"
    "gk" "gj"
    :destructive t)
```

Can you try that first and if that works, can you provide the values being passed to: 

```
(evil-collection-translate-key 'normal mode-keymaps
    (kbd "C-n") (kbd "C-j")
    (kbd "C-e") (kbd "C-k")
    "gn" "gj"
    "ge" "gk")
```

jojojames:
So translations like this:

```
(evil-collection-translate-key 'normal 'eshell-mode-map
    "gj" "gk"
    "gk" "gj"
    :destructive t)
```

Will work but when using:

```
(add-hook 'evil-collection-setup-hook #'my-hjkl-rotation)
```

The hook for eshell is ran too early, the keybinds aren't mapped yet. Looks like we already have a FAQ section for this (which also affects key translation). https://github.com/emacs-evil/evil-collection#faq

```
[...]

PR #614, opened 2022-04-14
Delay running `evil-collection-setup-hook` for modes like eshell (#602)

jojojames:
@noctuid Would love your thoughts on this. Think introducing this abstraction for these wonky modes with really late keymap initializations is good considering the hook's original purpose was to allow setting these types of binds after the bind exists is and already mapped.

This shouldn't affect how user's use the hooks as it'll get called at some point..

@condy0919

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -106,6 +106,31 @@ This will bind additional find-* type commands, e.g. usages, assignments, etc.."
   :type 'boolean
   :group 'evil-collection)
 
+(defvar evil-collection--modes-with-delayed-setup
+  `(eshell)
+  "List of modes whose keybinds aren't completely set up after the mode is
+loaded. This can be a problem for cases where we're doing key translations
+using `evil-collection-setup-hook' which would result in an empty keymap.
+
+Normally we run `evil-collection-setup-hook' right away after the mode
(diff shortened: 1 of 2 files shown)


noctuid:
I think it's good as a workaround, but it shouldn't be necessary forever and should only be necessary for eshell I think. This isn't so much wonky as it is completely broken. Reminder for anyone about the relevant code that caused this issue (even eshell knew it was broken):
```elisp
;; these are only set to nil initially for the sake of the
;; byte-compiler, when compiling other files which `require' this one
(defvar eshell-mode nil)
(defvar eshell-mode-map nil)
...
;; FIXME: What the hell!?
(setq-local eshell-mode-map (make-sparse-keymap))
...
```
This is why you have to define keys with `eshell-mode-hook` (because the keymap is buffer-local).

The good news is that it looks like this is fixed in Emacs 29 (maybe earlier), so it should eventually no longer be an issue. To continue to support older versions of eshell, it makes sense to do something like this PR I think.

jojojames:
I think ediff and emms may be other modes with bindings potentially deferred.

condy0919:
> The good news is that it looks like this is fixed in Emacs 29 (maybe earlier)

28+

PR #615, opened 2022-04-15
Add bindings for the new help-mode commands

0x28:
This pull request adds keybindings for the new help-mode commands introduced in Emacs 28. See also: https://git.savannah.gnu.org/cgit/emacs.git/tree/etc/NEWS.28#n558


jojojames:
Wonder if we should check for emacs 28 and bind it separately?

0x28:
The xref configuration also has this check. I think it makes sense and I will add it.

PR #619, opened 2022-04-24
Add "g," as a keybinding for `Info-index-next'

msaher:
I have noticed that there is no equivalent to `Info-index-next` in evil-collection-Info. In vanilla emacs, it is bound to `,`. I believe that `g,` is the most logical evil-equivalent.


condy0919:
Thanks

PR #620, opened 2022-04-26
vterm: add vterm-substitute and vterm-substitute-line

FrostyX:
I baesd the implementation on `evil-substitute` and
`evil-change-whole-line`.


condy0919:
Thanks

PR #621, opened 2022-04-27
vterm: fix paste-before and add paste-after command

FrostyX:
The `vterm-yank` that is currently bound to `p` actually pastes before
the cursor, so I am rebinding it to `P` and adding a function for
pasting after the cursor.

### DIFF
--- a/modes/vterm/evil-collection-vterm.el
+++ b/modes/vterm/evil-collection-vterm.el
@@ -91,6 +91,13 @@ also uses `evil-mode'."
   (vterm-goto-char (vterm--get-end-of-line))
   (call-interactively #'evil-append))
 
+(declare-function vterm-yank "vterm")
+
+(defun evil-collection-vterm-paste-after (&optional arg)
+  (interactive "P")
+  (vterm-goto-char (+ 1 (point)))
+  (call-interactively #'vterm-yank arg))
+
 (evil-define-operator evil-collection-vterm-delete (beg end type register yank-handler)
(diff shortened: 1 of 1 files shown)


condy0919:
> modes/vterm/evil-collection-vterm.el:94:2: Error: Unused lexical argument `arg'

modes/vterm/evil-collection-vterm.el
```diff
@@ -91,6 +91,11 @@ also uses `evil-mode'."
   (vterm-goto-char (vterm--get-end-of-line))
   (call-interactively #'evil-append))
 
+(defun evil-collection-vterm-paste-after (&optional arg)
```
condy0919, line 94: `arg` -> `_arg`

condy0919 (changes requested):

modes/vterm/evil-collection-vterm.el
FrostyX: Thank you for the review.
I used it for `vterm-yank` like so

```
 (call-interactively #'vterm-yank arg))
```

I am not sure if it is useful but I think it's better than just ignoring the parameter.

condy0919: OK

condy0919:
> modes/vterm/evil-collection-vterm.el: Error: the function ‘vterm-yank’ is not known to be defined.

`declare-function` it.

FrostyX:
Fixed

PR #622, opened 2022-04-28
vterm: use ^ for jumping at the beginning of the command but after prompt

FrostyX:
We have `0` for jumping at the very first character of the current
line and `^` for jumping to the non-blank character. It is useful to
keep this distinction in `vterm` buffers. However, it doesn't make
much sense in the command part. Because we have a prompt, both of the
commands jump to the very first character. I think it would be more
useful to change `^` to jump to the first non-blank character after
the prompt.

### DIFF
--- a/modes/vterm/evil-collection-vterm.el
+++ b/modes/vterm/evil-collection-vterm.el
@@ -67,6 +67,17 @@ also uses `evil-mode'."
                        "vterm"
                      "emacs"))))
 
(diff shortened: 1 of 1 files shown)


condy0919:
> modes/vterm/evil-collection-vterm.el: Error: the function ‘vterm-beginning-of-line’ is not known to be defined.

`declare-function` it.

FrostyX:
Thank you for the review. Fixed.

condy0919:
Thanks

PR #625, opened 2022-04-30
compile: bind TAB to compilation-next-error

edwargix:
Currently, it's bound to evil-jump-forward, which isn't useful in compile-mode.

### Checklist

- [x] byte-compiles cleanly

### DIFF
--- a/modes/compile/evil-collection-compile.el
+++ b/modes/compile/evil-collection-compile.el
@@ -50,6 +50,9 @@
       "go" 'compilation-display-error
       (kbd "S-<return>") 'compilation-display-error
 
+      (kbd "TAB") 'compilation-next-error
+      (kbd "S-TAB") 'compilation-previous-error
+
       "gj" 'compilation-next-error
       "gk" 'compilation-previous-error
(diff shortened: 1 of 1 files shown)


edwargix:
`S-TAB` currently does `'compilation-previous-error`.  Perhaps it would also be good to explicitly set that?  Let me know if you think it would be worth specifying that.

condy0919:
> `S-TAB` currently does `'compilation-previous-error`. Perhaps it would also be good to explicitly set that? Let me know if you think it would be worth specifying that.

It would be better to explicitly bind `S-TAB` since `S-TAB` in evil normal state is undefined and fallback to `<backtab>`.

edwargix:
> > `S-TAB` currently does `'compilation-previous-error`. Perhaps it would also be good to explicitly set that? Let me know if you think it would be worth specifying that.
> 
> It would be better to explicitly bind `S-TAB` since `S-TAB` in evil normal state is undefined and fallback to `<backtab>`. That's why it works.

a6467cf

condy0919:
Thanks

PR #626, opened 2022-05-06
vterm: when deleting command with `dd`, reset cursor after prompt

FrostyX:
The original `evil-first-non-blank` isn't usable because the beginning
of the prompt is the first non-blank character. Strangely enough,
`vterm-beginning-of-line` didn't work either.

### DIFF
--- a/modes/vterm/evil-collection-vterm.el
+++ b/modes/vterm/evil-collection-vterm.el
@@ -112,7 +112,7 @@ after the prompt."
 (declare-function vterm-reset-cursor-point "vterm")
 
 (evil-define-operator evil-collection-vterm-delete (beg end type register yank-handler)
-  "Modification of evil-delete to work in vterm buffer. 
+  "Modification of evil-delete to work in vterm buffer.
 Delete text from BEG to END with TYPE.
 Save in REGISTER or in the kill-ring with YANK-HANDLER."
   (interactive "<R><x><y>")
--- a/modes/vterm/evil-collection-vterm.el
+++ b/modes/vterm/evil-collection-vterm.el
@@ -109,6 +109,8 @@ after the prompt."
   (vterm-goto-char (+ 1 (point)))
   (call-interactively #'vterm-yank arg))
 
+(declare-function vterm-reset-cursor-point "vterm")
+
 (evil-define-operator evil-collection-vterm-delete (beg end type register yank-handler)
   "Modification of evil-delete to work in vterm buffer. 
 Delete text from BEG to END with TYPE.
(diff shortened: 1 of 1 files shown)


condy0919:
cc vterm developer @jixiuf

jixiuf:
https://github.com/akermu/emacs-libvterm/blob/master/vterm.el#L1695
when the `point` is at the end of the prompt ,call (vterm-beginning-of-line) would move to column 0

FrostyX:
Ah, thank you for the explanation @jixiuf.
If you don't see any issue with using `vterm-reset-cursor-point` then it might not be necessary for this PR.

But in general, I think could be useful.

jojojames:
How do you repro this manually?

FrostyX:
> How do you repro this manually?

Ah, sorry, I didn't realize the question was meant for me. Can you please explain what you mean by that?

jojojames:
No worries, I meant something along the lines of -> How do i see the behavior of "before this PR" and how do I compare it with "after this PR"?

FrostyX:
I think it would be easiest to show you screenshots. Let's say you have some line in vterm prompt.

vterm-before-dd

Then you pres `dd`. Before this patch, the command is deleted but the cursor jumps here

vterm-dd-before-patch


This PR changes it, and jumps the cursor into more reasonable place

vterm-dd-after-patch

FrostyX:
> Closing for now but happy to reopen 

I don't have the reopen PR button here, so maybe you can reopen?

jojojames:
I only get this when I 'dd' the line.

```
vterm--delete-region: Buffer is read-only: #<buffer *vterm*> [3 times]
```

FrostyX:
I believe that is because you didn't tell Emacs, what part of the line is the prompt and what is the command. I have the following line in my configuration:

```lisp
(setq term-prompt-regexp "^[^#$%>\n]*[#$%>] *")
```

Which I took from the "Good choices" in the documentation

```
term-prompt-regexp is a variable defined in ‘term.el’.
Its value is "^[^#$%>
]*[#$%>] *"

Documentation:
Regexp to recognize prompts in the inferior process.
Defaults to "^", the null string at BOL.

Good choices:
  Canonical Lisp: "^[^> \n]*>+:? *" (Lucid, franz, kcl, T, cscheme, oaklisp)
  Lucid Common Lisp: "^\\(>\\|\\(->\\)+\\) *"
  franz: "^\\(->\\|<[0-9]*>:\\) *"
  kcl: "^>+ *"
  shell: "^[^#$%>\n]*[#$%>] *"
  T: "^>+ *"

This is a good thing to set in mode hooks.

```

jojojames:
I see, reasonable enough for me, thanks!

PR #630, opened 2022-05-15
Bind `H` to `realgud:cmd-until-here`

Dickby:
As the title sais.
Before both `H` and `U` both where bound to the command `realgud:cmd-until` and there was no key for 
the function `realgud:cmd-until-here`


Dickby:
I don't know why this test failed `seq-into` is part of `seq` which is loaded at the beginning of the file?
Any ideas?

condy0919:
The same in https://github.com/emacs-evil/evil-collection/runs/6438981099?check_suite_focus=true

condy0919:
Thanks

PR #638, opened 2022-05-27
Correct cursor position when evil-append in vterm

quangIO:
vterm: currently, typing `a` in normal mode moves the cursor 2 characters to the right. 

We either (+ 1 (point)) and call evil-insert or evil-append at (point).


jojojames:
Does this look reasonable @mjlbach

Was added in: https://github.com/emacs-evil/evil-collection/pull/448

PR #639, opened 2022-05-27
vterm: use x to delete char and X to delete backward char

quangIO:
The behaviour of evil and evil in vterm should be consistent


modes/vterm/evil-collection-vterm.el
```diff
@@ -93,7 +93,7 @@ after the prompt."
 (defun evil-collection-vterm-append ()
   "Append character after cursor."
   (interactive)
-  (vterm-goto-char (1+ (point)))
```
jojojames, line 96: Could wait until the other change is merged or remove this from this PR.

PR #656, opened 2022-07-15
Changed some keys, so user can navigate the cursor during present.

funk443:
I think it's better to have j, k and w keys to remain their function inside org-present view, so people can navigate their cursor more easily.


condy0919:
Agreed.

PR #657, opened 2022-07-16
smerge: Add keybinding table

jgarte:
Hi, I added a README.org and keybinding table for smerge-mode.


condy0919:
The commit message should be like "smerge: ..."

PR #660, opened 2022-07-18
mu4e: replace obsolete mu4e-view-save-attachment-multi

tshu-w:
`mu4e-view-save-attachment-multi` does not exist anymore. I'm not sure whether we should delete this keybinding directly.

### DIFF
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -220,7 +220,7 @@ with older release versions of `mu4e.'"
      "cf" mu4e-compose-forward
      "cr" mu4e-compose-reply
      "p" mu4e-view-save-attachments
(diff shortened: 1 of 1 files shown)


condy0919:
It's OK, the master branch is always up to date.

tshu-w:
I want to say that `mu4e-view-save-attachments` is bound to `p` and `P` simultaneously. Should we delete `P` bindings? In the past, the `p` and `P` binding functions saved one attachment and multiple attachments, respectively. But now there is only this one function.

condy0919:
> I want to say that `mu4e-view-save-attachments` is bound to `p` and `P` simultaneously. Should we delete `P` bindings?

Please do. :-)

> In the past, the `p` and `P` binding functions saved one attachment and multiple attachments, respectively. But now there is only this one function.

PR #661, opened 2022-07-19
mu4e: delete the redundant keybinding

tshu-w:
delete the redundant keybinding as discussed in https://github.com/emacs-evil/evil-collection/pull/660#issuecomment-1188723037


condy0919:
Maybe `git pull origin master --rebase` first.

PR #670, opened 2022-08-10
mu4e: remove the inaccurate description

tshu-w:
-------

`revert-buffer-function` is set to `mu4e--main-view-real` in mu4e main mode. It's not the same as `mu4e-update-mail-and-index`(u).


condy0919:
Confirmed. tql

https://github.com/djcb/mu/blob/39d7096bba656cc64ec3ca8cc21ea3e6de9c2056/mu4e/mu4e-main.el#L125

PR #678, opened 2022-09-08
eglot: add binding for xref-find-definitions-other-frame

jgarte:

This adds a binding to eglot for xref-find-definitions-other-frame

### DIFF
--- a/modes/eglot/evil-collection-eglot.el
+++ b/modes/eglot/evil-collection-eglot.el
@@ -36,6 +36,8 @@
   "Set up `evil' bindings for `eglot'."
   (evil-collection-define-key 'normal 'eglot-mode-map
     "gd" 'xref-find-definitions
+    "gD" 'xref-find-definitions-other-window
+    "g5" 'xref-find-definitions-other-frame
     (kbd "C-t") 'xref-pop-marker-stack
     "K" 'eldoc-doc-buffer)
 


modes/eglot/evil-collection-eglot.el
```diff
@@ -36,6 +36,7 @@
   "Set up `evil' bindings for `eglot'."
   (evil-collection-define-key 'normal 'eglot-mode-map
     "gd" 'xref-find-definitions
+    "gD" 'xref-find-definitions-other-frame
```
condy0919, line 39: `g5`? Since other-frame commands usually are under C-x 5 prefix.

jgarte: That could work to. I prefer the ergonomics of pressing `gD`  than having to reach for `5` on my keyboard but I realize that this may be subjective/personal preference.

jgarte: @condy0919 WDYT

condy0919: I prefer to leave it in `g5` and make `gD` for the other-window version. Frames are rare to use when compared with windows.

jgarte: Sounds good to me. I'll send an update today.

PR #691, opened 2022-12-06
Bring `arc-mode` bindings closer to defaults

deuill:
This commit changes the normal-mode binding for `C` to `archive-copy-file` (which extracts, or copies, the file under cursor to the filesystem) from `archive-chown-entry` (which changes the file owner), and adds a new binding for the latter on `O`.

Understandably, `O` has a pre-existing meaning in Vi for normal mode, but is quite useless in `arc-mode`, since these buffers are declared read-only by default, and having them be read-write does nothing.

The intention behind this commit providing a straightforward way for file extraction to happen in `arc-mode` under Evil, a (presumably) common operation. Additionally, these new bindings now mirror those used in Emacs for `arc-mode`, as well as Emacs and Evil bindings for `tar-mode`; these improvements are likely worth the user-facing change made.

[Please write a quick summary of the package.]

https://github.com/your/awesome_package

### Checklist

- [ ] byte-compiles cleanly

### DIFF
--- a/modes/arc-mode/evil-collection-arc-mode.el
+++ b/modes/arc-mode/evil-collection-arc-mode.el
@@ -58,9 +58,10 @@
     "d" 'archive-flag-deleted
     "r" 'archive-rename-entry
     "x" 'archive-expunge
+    "C" 'archive-copy-file
     "M" 'archive-chmod-entry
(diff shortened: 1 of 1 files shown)


condy0919:
LGTM

PR #693, opened 2022-12-09
Fix evil clobbering active company-mode keymap

### LINKED ISSUE #664: `company-active-map` overridden when using company doc-buffer
When using company completion, if you open the doc buffer (`C-h`) then`company-active-map` will loose priority to `evil-insert-state-map`. This means that keys like `C-n`, `C-p`, and `C-k` no longer work for navigating the popup (interestingly `C-j`  still seems to work in this broken mode).

I was able to trace back this issue to this line in `company-doc-buffer`

https://github.com/company-mode/company-mode/blob/16ffeba5ef96c4c8e0cd39860b5402e25e304601/company.el#L2831

## to reproduce:
1. Begin company completion 
2.  `C-h` to open the doc buffer 
3. `C-n` will no longer work and will instead call `evil-complete-next` (The insert state binding)

CeleritasCelery:
fixes #664

- [ x ] byte-compiles cleanly
- [ x ] `M-x checkdoc` is happy.
- [ x ] define `evil-collection-mpc-setup` with `defun`
- [ x ] define `evil-collection-mpc-mode-maps` with `defconst`
- [ x ] All functions should start with `evil-collection-mpc-`

### DIFF
--- a/modes/company/evil-collection-company.el
+++ b/modes/company/evil-collection-company.el
@@ -33,6 +33,7 @@
 (declare-function company-tng-mode "company-tng")
 (declare-function company-grab-line "company")
 (declare-function company-begin-backend "company")
+(declare-function company-ensure-emulation-alist "company")
 
 (defgroup evil-collection-company nil
   "Evil bindings for `company-mode'."
@@ -43,6 +44,13 @@
   :type '(repeat symbol))
 (defcustom evil-collection-want-company-extended-keybindings nil
   "The \='evil-company-extended' keybindings should be requested"
+  :type 'boolean)
+
+(defcustom evil-collection-company-enable-keymap-protection t
+  "Prevent evil from breaking company completion keymap.
+When non-nil, prevents evil from overriding `company-active-map'
+after calling `company-doc-buffer'. If disabled, the completion
+keymap will be in a broken state."
   :type 'boolean)
 
 (defvar company-active-map)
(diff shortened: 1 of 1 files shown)


jojojames:
This looks good. Would request a boolean to not turn off this workaround though (default it to enabled).

CeleritasCelery:
I will add a boolean toggle. I am curious though, is there any reason someone would *not* want the fix enabled?

condy0919:
> Asked another way, is it really worth adding a toggle for something that almost no one would want disabled? As opposed to just using the Emacs builtin way to turn it off?
> 
> ```lisp
> (remove-hook 'evil-local-mode-hook #'evil-collection-company-protect-keymap)
> ```

It's inconvenient for users.

CeleritasCelery:
Thanks. Added the toggle.

condy0919:
> Thanks. Added the toggle.

squash your commits. The commit message should be like "company-mode: ..." or something else.

CeleritasCelery:
done

modes/company/evil-collection-company.el
```diff
@@ -112,6 +126,9 @@ C-x C-l."
     (kbd "M-k") 'company-select-previous
     (kbd "<escape>") 'company-search-abort)
 
+  ;; https://github.com/emacs-evil/evil-collection/issues/664
+  (add-hook 'evil-local-mode-hook #'evil-collection-company-protect-keymap)
```
jojojames, line 130: I'll probably move the toggle around this add-hook instead at some point if I remember. Thanks!

PR #696, opened 2022-12-16
Support new `python-ts-mode`

meliache:
## Support Emacs 29 `python-ts-mode` (tree-sitter  mode for python)

The yet unreleased Emacs 29 can be compiled with built-in tree sitter support, and includes several variants of programming language major modes which do font-locking via tree-sitter. Usually they end with `...-ts-mode` and are defined in the same elisp package as the original major mode. They have separate keymaps and hooks, so to support them those would also need to be used.

I looked supporting `python-ts-mode` via `evil-collection-python-setup`. Actually, the new `python` package also defines a new `python-base-mode` from which both `python-mode` and `python-ts-mode` inherit. In my personal configuration I therefore just use `python-base-mode-hook`  and `python-base-mode-map`, but this would not be backward compatible to older emacs versions, therefore I suggest just adding to both the normal and ts-modes the keybindings and hooks.

This is something should also be addressed for other programming major modes in the feature and I'm not sure if this has been discussed anywhere. One could also think of defining support for those modes in separate files, but not sure if that's necessary since the ts-mode and original-modes are also defined in the same files, and the setup will be probably identical for both.

I'm definetly not an expert when  it comes to tree-sitter and sadly this PR doesn't have a general solution for all modes, so I'm free if this will be rejected. But this is the fork I use now and at least I'd like to get a discussion started about the ts-modes.

### Checklist

- [x] byte-compiles cleanly
- [x] define `evil-collection-python-setup` with `defun`
- [x] define `evil-collection-python-mode-maps` with `defconst`
- [x] All functions should start with `evil-collection-python-`

### DIFF
--- a/modes/python/evil-collection-python.el
+++ b/modes/python/evil-collection-python.el
@@ -30,7 +30,9 @@
 (require 'evil-collection)
 (require 'python)
 
-(defconst evil-collection-python-maps '(python-mode-map))
+(defconst evil-collection-python-maps '(python-mode-map
+                                        python-ts-mode-map
+                                        python-base-mode-map))
 
 (defun evil-collection-python-set-evil-shift-width ()
   "Set `evil-shift-width' according to `python-indent-offset'."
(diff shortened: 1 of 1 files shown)


jojojames:
Do the new tree sitter modes have additional functions that may or may not be bound or they are strictly the same with the exception of the modes being different now?

If you only use python-ts-mode, does this binding layer even load? DOes python-ts-mode (require 'python)?

meliache:
Yes, `python-ts-mode` requires the `'python` package, i.e. it is provided by  it. It's also just defined in `progmodes/python.el` (see [here](https://github.com/emacs-mirror/emacs/blob/4d1e4a48938be8a08e294b4a5711a2c8ce941a5a/lisp/progmodes/python.el#L6627-L6686)). Therefore, I assume it will be loaded with python. The keymaps also seem to be identical, the code currently has
``` emacs-lisp
(defvar python-ts-mode-map (copy-keymap python-mode-map))
```

But this is not necessarily be the case for all tree sitter modes. `c-ts-mode` is defined in its own `progmodes/c-ts-mode.el`, and you need to require it via `(require 'c-ts-mode)`. So this approach here would not work for all built-in tree-sitter major modes. Also I can't guarantee that python-ts-mode will never be seperated into its own subpackage, especially as this is still unreleases (though already in the emacs 29 release branch and not just master).

condy0919:
LGTM.

PR #702, opened 2023-01-22
vterm: fix the `x' key, currently it does nothing

FrostyX:
Currently, pressing the `x` key in normal mode does nothing. I don't know much about `:motion` but AFAIK `evil-delete-char` doesn't exist. Using `evil-forward-char` fixes the issue.


FrostyX:
Thank you @jojojames :-)

PR #705, opened 2023-02-10
Rename smerge directory

northgaard:
The directory is called "smerge" while the configuration file uses "smerge-mode". This inconsistency means that you cannot use evil-collection-require to load the feature.


condy0919:
Good catch.

PR #710, opened 2023-02-15
fix(color-rg): keymap `override`

mumu-lhl:


condy0919:
Does it really need `override` keymap bindings?

mumu-lhl:
Yes, it does. Because I found that there will be a warning if it is not added. And the color-rg keybindings need to use `override` keymap bindings (eg `C`).

mumu-lhl:
It seems I made a misunderstanding. I will close this pull request.

PR #717, opened 2023-03-30
vterm: bypass line wraps in the prompt line

FrostyX:
When the command in the prompt line is so long that it wraps to multiple lines, movements such as `h`, `l`, `^`, and `$` IMHO behave unexpectedly. In normal terminal emulators outside of Emacs, a long prompt line contains only soft-wraps, i.e. being at the last character before a wrap and pressing `<right>` moves the cursor to the first character of the next line. This is also how vterm behaves without Evil. I am changing `h` and `l` to behave the same way.

While doing so, I am doing the same for `^` and `$`.

### DIFF
--- a/modes/vterm/evil-collection-vterm.el
+++ b/modes/vterm/evil-collection-vterm.el
@@ -40,6 +40,15 @@
 
 (defvar vterm--process)
 
+(defcustom evil-collection-vterm-prompt-line-continuation t
+  "Make movements within prompt behave like it was one line.
+This option becomes relevant when the command in prompt is so long that it wraps
+to multiple lines. If the value is `t', movements such as `h', `l', `^', `$',
+etc allows to move between lines (like it would be in any other terminal
+emulator outside of Emacs or in Vterm without Evil)."
+  :type 'boolean
+  :group 'vterm)
+
 (defun evil-collection-vterm-escape-stay ()
   "Go back to normal state but don't move cursor backwards.
 Moving cursor backwards is the default vim behavior but
@@ -74,9 +83,17 @@ also uses `evil-mode'."
   "Move the cursor to the first non-blank character
 after the prompt."
   :type exclusive
-  (if (vterm-cursor-in-command-buffer-p (point))
+  (if (and evil-collection-vterm-prompt-line-continuation
+           (vterm-cursor-in-command-buffer-p (point)))
       (vterm-beginning-of-line)
     (evil-first-non-blank)))
+
+(evil-define-motion evil-collection-vterm-end-of-line (count)
+  "Move the cursor to the end of the current line."
+  :type inclusive
+  (if evil-collection-vterm-prompt-line-continuation
+      (vterm-end-of-line)
+    (evil-end-of-line count)))
 
 (defun evil-collection-vterm-insert ()
   "Insert character before cursor."
@@ -223,6 +240,28 @@ But don't allow the cursor to move bellow the last prompt line."
   (when (> (count-words (point) (point-max)) 0)
     (evil-next-line count)))
 
+(evil-define-motion evil-collection-vterm-forward-char
+  (count &optional crosslines noerror)
+  "Move cursor to the right by COUNT characters, bypassing line wraps."
+  :type exclusive
+  (if (and evil-collection-vterm-prompt-line-continuation
+           (get-text-property (1+ (point)) 'vterm-line-wrap))
+      (forward-char 2)
+    (evil-forward-char count crosslines noerror)))
+
+(evil-define-motion evil-collection-vterm-backward-char
+  (count &optional crosslines noerror)
+  "Move cursor to the left by COUNT characters, bypassing line wraps."
+  :type exclusive
+  ;; Check if one character to the left is a vterm's line wrap.
+  ;; The `bobp' part is a special case for the first row and first character
+  ;; of the buffer - `evil-backward-char' would throw an error otherwise.
+  (if (and evil-collection-vterm-prompt-line-continuation
+           (get-text-property (1- (point)) 'vterm-line-wrap)
+           (not (bobp)))
+      (forward-char -1)
+    (evil-backward-char count crosslines noerror)))
+
 ;;;###autoload
 (defun evil-collection-vterm-setup ()
   "Set up `evil' bindings for `vterm'."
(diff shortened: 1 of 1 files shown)


FrostyX:
Just to easily illustrate behavior. When I press `l` without this patch, it does nothing. With this patch, it moves the cursor to the next line.

Screenshot_2023-03-30_02-37-02

Similarly `h` goes to the previous line

Screenshot_2023-03-30_02-37-54

This works in the visual mode as well.

modes/vterm/evil-collection-vterm.el
```diff
@@ -223,6 +223,22 @@ But don't allow the cursor to move bellow the last prompt line."
...
+  :type exclusive
+  (if (get-text-property (1+ (point)) 'vterm-line-wrap)
+      (forward-char 2)
+    (evil-forward-char count crosslines noerror)))
+
+(evil-define-motion evil-collection-vterm-backward-char
+  (count &optional crosslines noerror)
+  "Move cursor to the left by COUNT characters, bypassing line wraps."
+  :type exclusive
+  (if (and (not (bobp)) (get-text-property (1- (point)) 'vterm-line-wrap))
```
jojojames, line 238: Can you add a comment on what this is checking for.

FrostyX: Thank you for the reply @jojojames,

I got the idea from `vterm--get-beginning-of-line`. Basically `(get-text-property (1- (point)) 'vterm-line-wrap)` tells us whether one character to the left is a vterm's line wrap. The `(not (bobp))` is a special case for the first row and first character of the buffer. Because one character to the left is nothing and `evil-backward-char` would throw `if: Args out of range: 0, 0`.

Please let me know what part of this is interesting to you, and I'll add as a code comment.

jojojames: Up to you, what you wrote here seems pretty good to me.

FrostyX: Done

modes/vterm/evil-collection-vterm.el
```diff
@@ -282,8 +299,15 @@ But don't allow the cursor to move bellow the last prompt line."
     "G" 'vterm-reset-cursor-point)
 
   (evil-collection-define-key 'visual 'vterm-mode-map
+    "x" 'evil-collection-vterm-delete-backward-char)
+
+  (evil-collection-define-key '(normal visual) 'vterm-mode-map
     "d" 'evil-collection-vterm-delete
-    "x" 'evil-collection-vterm-delete-backward-char))
+    "l" 'evil-collection-vterm-forward-char
+    "h" 'evil-collection-vterm-backward-char
```
jojojames, line 307: One last thing, can you add a flag (can default to true), to conditionally enable this.

Something like:

if flag:
 bind l h $ ^ to -> new feature
else:
 bind ^ (whatever other key needs to be deleted from thsi CL)

This cl is moving some lines to other parts so hard to tell what's missing/gained easily. (Not taking too close a look)

FrostyX: I agree, that sounds like a good idea. Please take a look.

jojojames: I think we should create 2 separate flags (one for ^ $ and one for h l) and then check the flags only in evil-collection-vterm-setup to conditionally bind the functions.

One use case I'm thinking (which is one I'd be using) is to only bind this for h & l movements but keep ^ and $ movements alone.

modes/vterm/evil-collection-vterm.el
```diff
@@ -265,12 +304,10 @@ But don't allow the cursor to move bellow the last prompt line."
     "P" 'vterm-yank
     "a" 'evil-collection-vterm-append
     "A" 'evil-collection-vterm-append-line
-    "d" 'evil-collection-vterm-delete
```
jojojames, line 268: Why are we removing this line?

modes/vterm/evil-collection-vterm.el
FrostyX: > I think we should create 2 separate flags (one for ^ $ and one for h l)
> One use case I'm thinking (which is one I'd be using) is to only bind this for h & l movements but keep ^ and $ movements alone.

It doesn't really make sense to me, separating the configuration into two options. In my case, I either want the prompt to behave like one long line or not :-). That being said, if you are interested in having this behavior just for h, k and not ^, $, maybe others will as well. So sure, no problem with me. Do you have any preference for the option names?

> check the flags only in evil-collection-vterm-setup to conditionally bind the functions.

I am trying to find some other key bind that is done this way but not seeing any. Have you seen this approach somewhere else in `evil-collection`? I'd like to be consistent with it.

modes/vterm/evil-collection-vterm.el
FrostyX: I moved it from 

```lisp
  (evil-collection-define-key 'normal 'vterm-mode-map
```

to

```
  (evil-collection-define-key '(normal visual) 'vterm-mode-map
```

since I was creating this section, and it makes sense for `d` to work in `visual` mode as well.

modes/vterm/evil-collection-vterm.el
FrostyX: *bump* @jojojames

jojojames: https://github.com/emacs-evil/evil-collection/blob/5092f5e04cfd0a31b1c5c46cb6b792a9e917c728/modes/elpaca/evil-collection-elpaca.el#L105

FrostyX: It won't be as easy as we thought. Or at least as I thought.

We would have to define two variants for every movement. I am proposing `hj^$` in this PR but I forgot `X`, `0`, operations till the end of the line like `C`, `S`, `D`, `Y`, jumps like `t<char>`, `T<char>`, or line operations like `yy`, `dd`, and whatever else. That would be too many vterm specific key binds. But more importantly too many key binds for this line-continuation feature. Imagine how much would the number of key binds blow up if we want to add some other feature with conditional keybinds. There must be a smarter way of doing this but I am struggling to find it.

I am playing with overriding something like `evil-line-expand` but without much success. It may be a dead end as well.

You have much more experience with Evil than me, do you have any idea @jojojames, please?

jojojames: I think it'd be better to group a set of binds together than to consider them each individually, that way the amount of new flags won't balloon out of control. h&j would be in a separate group from $&0 and whatever else.

The grouping can be decided reasonably, I'd assume.

jojojames:
Sorry took a while to respond, I'm assuming you may be uninterested in pursing this further so closing for now, but feel free to reopen as it looks reasonable to add.

PR #720, opened 2023-04-03
Adapt evil-collection-mu4e for latest mu4e development version

### LINKED ISSUE #695: evil-collection-mu4e update required
There are some changes in the development version of mu4e which break evil-collection-mu4e.el In particular, the function, mu4e~view-quit-buffer no longer exists. This means that mu4e-view-mode-map entry for 'q' needs to be changed. Instead of binding q to mu4e~view-quit-buffer, I found binding to quit-window works. 

Note that this only affect the current 1.9.x version (1.9.7 specifically). It does not affect earlier versions of mu4e. 

As it stands now, when the user hits q to leave the view buffer (*mu4e-article* window), an error is thrown - wrong argument commandp as the function mu4e~viewquit-buffer does not exist.

meliache:
See the discussion in #695, which this PR should resolve. I just took the code provided by @theophilusx in https://github.com/emacs-evil/evil-collection/issues/695#issuecomment-1437577481 and made it into a branch. Also set him as the author of the first commit.

The latest mu4e version takes care of updating the keybindings in the menu when rebinding keys, therefore we can remove plenty of code related to that, and just focus on the bindings. This also makes this much more stable, as the menu was often a source of problems in the past with mu4e-updates.

mu4e relies on the `mu` program to be compiled, therefore many users use the version from their distro, which still might be v.1.8 or older, e.g. see https://packages.ubuntu.com/kinetic/mu4e. The menu might not work for those anymore, but one can live without a menu as long as general bindings work. 

So I'm thinking whether I should maybe provide backwards-compatible keybindings at least. Some function names changed, so i could add some code which either checks the mu4e version or just use the first bound function, like 

``` elisp
    "q" (or mu4e-view-quit mu4e~view-quit-buffer)
```

Another issue is now that several toggle-functions have been obsoleted since 1.9.5 and all been replaced by `mu4e-headers-toggle-property`, which invokes a transient menu in which we can toggle to which the old toggle different properties via single keys, which works in evil out-of-the-box. Due to the aliases, all the following keybindings now effectively open the same menu:

- `mu4e-headers-toggle-include-related` (`z r`), 
- `mu4e-headers-toggle-skip-duplicates` (`z d`)
- `mu4e-headers-toggle-threading` (`z t`)

This key-overlap is not necessarily bad, the overlap helps with muscle-memory when coming from the previous version and helps with backwards-compatibility, but those aliases are deprecated and might be gone in future versions, so we might try calling the property menu directly if a new version of mu4e is used for forward-compatibility.

Also, if we decide to only bind a single key to the property-menu, then some keys might be freed to do other things.

### TODO's:
  
- [ ] Maybe add backwards-compatible bindings?
- [ ] Maybe bind `mu4e-headers-toggle-property` directly instead of relying on aliases to obsolete functions for forward-compatibility.
    - [ ] Bind it to a single key or keep key overlap? The latter is simpler for backwards compatiblity. If we decide to only keep a single key for the toggle menu, which should it be?
    - [ ] Update table of keybindings in package commentary

### Required checks:
- [x] byte-compiles cleanly
- [x] define `evil-collection-mu4e-setup` with `defun`
- [x] define `evil-collection-mu4e-maps` with `defconst`
- [x] All functions should start with `evil-collection-mu4e-`

Resolves #695

### DIFF
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -89,9 +89,11 @@
 ;; putting the user into insert-state. The below code, together with the hooks
 ;; set in evil-collection-mu4e-setup fixes this issue.
 (defun evil-collection-mu4e-org-set-header-to-normal-mode ()
+  "Set initial state in `mu4e-compose-mode' to \='normal."
   (evil-set-initial-state 'mu4e-compose-mode 'normal))
 
 (defun evil-collection-mu4e-org-set-header-to-insert-mode ()
+  "Set initial state in `mu4e-compose-mode' to \='insert."
   (evil-set-initial-state 'mu4e-compose-mode 'insert))
 
 (defvar evil-collection-mu4e-mode-map-bindings
@@ -257,9 +259,10 @@
 
 
 
+
 (defun evil-collection-mu4e-setup ()
   "Initialize evil-mu4e if necessary.
-If mu4e-main-mode is in evil-state-motion-modes, initialization
+If `mu4e-main-mode' is in `evil-state-motion-modes', initialization
 is already done earlier."
     (evil-collection-mu4e-set-state)
     (evil-collection-mu4e-set-bindings))
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -68,7 +68,7 @@
 (require 'evil-collection)
 (require 'mu4e nil t)
 
-(if (version< mu4e-mu-version "1.10")
+(if (version< mu4e-mu-version "1.9")
     (require 'evil-collection-mu4e
              (expand-file-name "modes/mu4e/evil-collection-mu4e-1.8" evil-collection-base-dir))
 
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -76,7 +76,7 @@
 ;; continue supporting loading 1.8 code because mu4e has a C-dependency and is
 ;; often installed from linux distro package repositories, and therefore can be
 ;; severely out-of-date.
-(if (and (boundp mu4e-mu-version)       ; avoid evil-collection unit-tests failing from unbound variable
+(if (and (boundp 'mu4e-mu-version)       ; avoid evil-collection unit-tests failing from unbound variable
          (version< mu4e-mu-version "1.9"))
     (require 'evil-collection-mu4e
              (expand-file-name "modes/mu4e/evil-collection-mu4e-1.8" evil-collection-base-dir))
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -68,7 +68,15 @@
 (require 'evil-collection)
 (require 'mu4e nil t)
 
-(if (and (boundp mu4e-mu-version)
+;; Load compatibility code for Mu4e versions 1.8 or older.
+;; Mu4e version 1.10 introduced a new backwards-incompatible interface. In
+;; addition to new function names and some functions hidden in menus, 1.10
+;; reduces what we need to implement in evil-collection because it dynamically
+;; adapts the keys in the mu4e menu-buffer to the actually bound keys. We
+;; continue supporting loading 1.8 code because mu4e has a C-dependency and is
+;; often installed from linux distro package repositories, and therefore can be
+;; severely out-of-date.
+(if (and (boundp mu4e-mu-version)       ; avoid evil-collection unit-tests failing from unbound variable
          (version< mu4e-mu-version "1.9"))
     (require 'evil-collection-mu4e
              (expand-file-name "modes/mu4e/evil-collection-mu4e-1.8" evil-collection-base-dir))
@@ -206,7 +214,7 @@
        "cr" mu4e-compose-reply
        "p" mu4e-view-save-attachments
        "O" mu4e-headers-change-sorting
-       "A" mu4e-view-mime-part-action ; Since 1.6, uses gnus view by default
+       "A" mu4e-view-mime-part-action   ; Since 1.6, uses gnus view by default
        "a" mu4e-view-action
        "J" mu4e~headers-jump-to-maildir
        "[[" mu4e-view-headers-prev-unread
@@ -217,7 +225,7 @@
        "\C-k" mu4e-view-headers-prev
        "x" mu4e-view-marked-execute
        "&" mu4e-view-mark-custom
-       "*" mu4e-view-mark-for-something   ; TODO: Don't override "*".
+       "*" mu4e-view-mark-for-something ; TODO: Don't override "*".
        "m" mu4e-view-mark-for-move
        "r" mu4e-view-mark-for-refile
        "D" mu4e-view-mark-for-delete
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -68,204 +68,209 @@
 (require 'evil-collection)
 (require 'mu4e nil t)
 
-(defconst evil-collection-mu4e-maps '(mu4e-main-mode-map
-                                      mu4e-headers-mode-map
-                                      mu4e-view-mode-map
-                                      mu4e-compose-mode-map
-                                      mu4e-search-minor-mode-map))
-
-(defun evil-collection-mu4e-set-state ()
-  "Set the appropriate initial state of all mu4e modes."
-  (dolist (mode '(mu4e-main-mode
-                  mu4e-headers-mode
-                  mu4e-view-mode
-                  mu4e-org-mode))
-    (evil-set-initial-state mode 'normal))
-  (evil-set-initial-state 'mu4e-compose-mode 'insert))
+(if (version< mu4e-mu-version "1.10")
+    (require 'evil-collection-mu4e
+             (expand-file-name "modes/mu4e/evil-collection-mu4e-1.8" evil-collection-base-dir))
 
 
-;; When using org-mu4e, the above leads to an annoying behaviour, because
-;; switching from message body to header activates mu4e-compose-mode, thus
-;; putting the user into insert-state. The below code, together with the hooks
-;; set in evil-collection-mu4e-setup fixes this issue.
-(defun evil-collection-mu4e-org-set-header-to-normal-mode ()
-  "Set initial state in `mu4e-compose-mode' to \='normal."
-  (evil-set-initial-state 'mu4e-compose-mode 'normal))
+  (defconst evil-collection-mu4e-maps '(mu4e-main-mode-map
+                                        mu4e-headers-mode-map
+                                        mu4e-view-mode-map
+                                        mu4e-compose-mode-map
+                                        mu4e-search-minor-mode-map))
 
-(defun evil-collection-mu4e-org-set-header-to-insert-mode ()
-  "Set initial state in `mu4e-compose-mode' to \='insert."
-  (evil-set-initial-state 'mu4e-compose-mode 'insert))
-
-(defvar evil-collection-mu4e-mode-map-bindings
-  `((mu4e-main-mode-map
-     "J" mu4e~headers-jump-to-maildir
-     "j" next-line
-     "k" previous-line
-     "u" mu4e-update-mail-and-index
-     "gr" revert-buffer
-     "b" mu4e-search-bookmark
-     "B" mu4e-search-bookmark-edit
-     "N" mu4e-news
-     ";" mu4e-context-switch
-     "H" mu4e-display-manual
-     "C" mu4e-compose-new
-     "cc" mu4e-compose-new
-     "x" mu4e-kill-update-mail
-     "A" mu4e-about
-     "f" smtpmail-send-queued-mail
-     "m" mu4e~main-toggle-mail-sending-mode
-     "s" mu4e-search
-     "q" mu4e-quit)
-
-    (mu4e-headers-mode-map
-     "q" mu4e~headers-quit-buffer
-     "J" mu4e~headers-jump-to-maildir
-     "C" mu4e-compose-new
-     "E" mu4e-compose-edit
-     "F" mu4e-compose-forward
-     "R" mu4e-compose-reply
-     "cc" mu4e-compose-new
-     "ce" mu4e-compose-edit
-     "cf" mu4e-compose-forward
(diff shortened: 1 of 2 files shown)


theophilusx:
Hi Michael,

I agree with all of your points. Many of which are why I had not found
the time to put a formal PR together. Thank you for doing that.

BTW I don't feel I need an author acknowledgement. I didn't really author
anything, just removed code no longer required given internal mu4e
changes.

I also agree with your points regarding some of the bindings being
obsolete given changes within mu4e. I also came to the same conclusion
i.e. useful for muscle memory for existing users, but really needs to be
marked as obsolete and flagged for removal in the future.

The question of backward version compatibility is the main one why I had
not pulled a formal PR together. I just wasn't sure what was the best
route to take. As you noted, the code is much smaller, cleaner and more
likely to remain stable without the code necessary for backwards
compatibility. At the same time, many distributions, especially ones
like debian stable, will not come with this new version of mu for some
time. Whether it is better to require them to pin evil-collection to an
older version (as many of their other packages are likely older as well)
or include it within evil-collection-mu4e module (debian makes it more
complex because they also include a lot of elisp dpkg packages for
things many people now just grab from ELPA repositories).

At any rate, your PR certainly moves things forward and I appreciate you
taking the pressure off me right now as I have a full plate for the next
few weeks.

Tim

jojojames:
> mu4e relies on the mu program to be compiled, therefore many users use the version from their distro, which still might be v.1.8 or older, e.g. see https://packages.ubuntu.com/kinetic/mu4e. The menu might not work for those anymore, but one can live without a menu as long as general bindings work.

Do we know which version of mu the current version of evil-collection-mu4e works for? Maybe we can just add a FAQ entry telling users to go to [insert revision of evil-collection-mu4e where it will still work for them] and have the user manually pull it in. We could add some compatibility code but I'm afraid it'll fossilize this file and people will be hesitant to make changes to it for fear of breaking something. I'd be happy to just go completely towards the latest version.

>  Due to the aliases, all the following keybindings now effectively open the same menu:

Up to you.

meliache:
> Do we know which version of mu the current version of evil-collection-mu4e works for?

The latest stable release was 1.8 for which the current mu4e-collection works if I remember correctly. The 1.9 release series number was just the development release, it introduced the new changes slowly and was never published as a release, and the new 1.10 is the new stable release. See https://github.com/djcb/mu/releases

milanglacier:
> We could move the current evil-collection-mu4e to something like evil-collection-mu4e-1.8.el (if that name works) and create a new evil-collection-mu4e. Then create a a defcustom that lets the user define which mu they are using and we can conditionally switch based on that. (That's iff you want to do any compatibility stuff).

I think we don't need such `defcustom`, there is a variable `mu4e-mu-version` defined in `mu4e`. So we just need to check the value of `mu4e-mu-version` and conditionally load `eveil-collection-mu4e.el` or `evil-collection-mu4e-1.8.el`

jojojames:
Wouldn't that cost a system call? Seems slow-er, imo.

milanglacier:
> Wouldn't that cost a system call? Seems slow-er, imo.

No. This is hard coded in `mu4e-config.el`. So as long as the user uses the mu4e shipped with the mu binary, this variable should reflect the mu version correctly.

hokomo:
Perhaps it would also make sense to immediately break off a new file such as `evil-collection-mu4e-latest.el` rather than having all of the top-level definitions nested under the conditional for the version check in the main file. The main file would then just contain common definitions, perform the version check and load the appropriate version.

This way, in the future when mu/mu4e comes out with another release, we can just create additional files for new versions.

milanglacier:
> Just wanted to link to a comment I posted regarding backwards compatibility in the recent mu4e issue ([#695 (comment)](https://github.com/emacs-evil/evil-collection/issues/695#issuecomment-1502465893)), since it seems discussion has been going on in multiple places.
> 
> @milanglacier I've taken a look at your branch. I like the approach of having both a new file and a version check to load the appropriate file automatically. I'm assuming the differences between 1.8.x and 1.10.x and big enough to warrant a separate file? Otherwise, a single file could be enough.


I think @jojojames prefers the 2 files approach which is why I did so. You can check the commit in this PR: `8aa75087987766f4196da94610aebcd6f4680be8` to see how much difference between 1.8 and 1.10.

I think the change is not that much, a lot of code are cleaned, but if we want to keep them in single file, then what we need to do is just that: 1. don't call 1-2 functions 2. change 1-2 function rename 3. don't load some hooks conditioning on mu4e version. We can leave those function definitions necessary for mu4e 1.8 alone.

But the decision that make the code clean and maintainable is also a good decision.

> I would just make the version checks a little bit more explicit: (1) check for 1.8.x and load the compatibility version, (2) check for lower than 1.8.x and raise an error/warning rather than silently failing, and (3) load the latest version otherwise.

There's no need to do (2). Has anyone reported compatability issue wieth mu4e 1.8 and mu4e 1.8-? If not, we can just treat mu4e 1.8- and mu4e 1.8 the same.

> Perhaps it would also make sense to immediately break off a new file such as evil-collection-mu4e-latest.el rather than having all of the top-level definitions nested under the conditional for the version check in the main file. The main file would then just contain common definitions, perform the version check and load the appropriate version.

Shouldn't `evil-collection-mu4e.el` naturally correspond to the latest API? I think an additional `evil-collection-mu4e-latest.el` is just unintuitive to me.

Does emacs lisp provide a mechanism to early stop loading a file (like `finish` in vimscript)? I do agree nest our main code in a conditionals is ugly? But this scenerio should  fit better to the early return clause.

nielsk:
I tried the file and for some reason every movement in the headers-view opens directly the mail. I hit D which marks a mail for deletion and opens the next one, "j" or "k" moves up or down but directly opens the mail.

hokomo:
> I think the change is not that much, a lot of code are cleaned, but if we want to keep them in single file, then what we need to do is just that: 1. don't call 1-2 functions 2. change 1-2 function rename 3. don't load some hooks conditioning on mu4e version. We can leave those function definitions only necessary for mu4e 1.8 alone.
> 
> But the decision that make the code clean and maintainable is also a good decision.

Agreed, the change doesn't look that big and it's nice that we can get rid of the buffer hacking. All in all, I think I'd prefer the multiple files route since it will probably "scale" better as newer versions of mu are released.

> > I would just make the version checks a little bit more explicit: (1) check for 1.8.x and load the compatibility version, (2) check for lower than 1.8.x and raise an error/warning rather than silently failing, and (3) load the latest version otherwise.
> 
> There's no need to do (2). Has anyone reported compatability issue wieth mu4e 1.8 and mu4e 1.8-? If not, we can just treat mu4e 1.8- and mu4e 1.8 the same.

Yes, I have (see #672), before I realized I could just avoid the issue by upgrading to mu 1.8.x. The current code (which assumes 1.8.x) is definitely not compatible with 1.6.x. This is why I'd like to give a heads-up to anyone who finds themselves in a similar situation, now or in the future, by explicitly enumerating what we support and what we don't (rather than just praying it sort of works).

> Shouldn't `evil-collection-mu4e.el` naturally correspond to the latest API? I think an additional `evil-collection-mu4e-latest.el` is just unintuitive to me.

Agreed, it's probably not the greatest name. I guess being explicit and just using the concrete version would be better, i.e. `...-1.10.el` instead of `...-latest.el`.

> Does emacs lisp provide a mechanism to early stop loading a file (like `finish` in vimscript)? I do agree nesting our main code in conditionals is ugly. But this scenerio should fit better to the early return clause.

Sadly, no, and this is how the loading process works in every Lisp-like language -- all of the top-level definitions are processed in order (unless you want to start doing hacks like intentionally throwing a dummy error, but that's even worse than the nesting). The only clean way to avoid the nesting (which I think would be nice) would be to have a separate file for each version.

milanglacier:
> The only clean way to avoid the nesting (which I think would be nice) would be to have a separate file for each version.

I agree with you. Let's see @jojojames's opinion about the final decision.

> Yes, I have (see https://github.com/emacs-evil/evil-collection/pull/672), before I realized I could just avoid the issue by upgrading to mu 1.8.x. The current code (which assumes 1.8.x) is definitely not compatible with 1.6.x. This is why I'd like to give a heads-up to anyone who finds themselves in a similar situation, now or in the future, by explicitly enumerating what we support and what we don't (rather than just praying it sort of works).

mu 1.8 has been there for a long while now... I don't think we need to too conservative about the backward compatability. Given a warning is fine, but I do not want to find the old versioned file and create a "evil-collection-mu4e-1.6" anymore...

hokomo:
> mu 1.8 has been there for a long while now... I don't think we need to too conservative about the backward compatability. Given a warning is fine, but I do not want to find the old versioned file and create a "evil-collection-mu4e-1.6" anymore...

I agree, of course not. That's why I mentioned I would only start this backwards-compatibility stuff with 1.8.x. Version 1.6.x (or rather, any version less than 1.8.x) we can essentially drop support for and just issue a small warning as a heads-up.

meliache:
I'm happy about the activity on this PR, had been a bit quiet due to being a bit busy the last weeks. I created this PR because nobody else did first and I thought something is better than nothing. I'm fine with using something like @milanglacier's branch, it sounds good to me. I could also merge that into my PR branch to continue review here or I'm fine  with a new PR and this one being closed.a

jojojames:
2 file approach sounds good.

Prefer one evil-collection-mu4e-1.8 (or whatever) and one evil-collection-mu4e. If we need to update again for whatever reason, we can just cp evil-collection-mu4e to some evil-collection-mu4e-xx.el and update the latest file.

PR wise, also ok to pull in whatever is more appropriate.

milanglacier:
I created a PR to @meliache's mu4e branch [https://github.com/meliache/evil-collection/pull/1](https://github.com/meliache/evil-collection/pull/1) for the two files approach.

Since this PR has been referenced to so many other discussions/PRs. I think it is better we keep making the progress in this PR rather than create a new PR.

meliache:
> I created a PR to @meliache's mu4e branch [https://github.com/meliache/evil-collection/pull/1](https://github.com/meliache/evil-collection/pull/1) for the two files approach.
> 
> Since this PR has been referenced in so many other discussions/PRs. I think it is better we keep making the progress in this PR rather than create a new PR.
> 
> And if we are going to use the two files approach, I don't want us to be too conservative about the backward compatibility. When mu4e 1.12 is released, I prefer to dropping support for mu4e 1.8. 

Merged

edgar-vincent:
Should we bind `mu4e-headers-{previous next}-thread` to something? They are bound to `{`  and `}` by default. I think it makes sense to remap `evil-{forward backward}-paragraph`.

Thanks for this PR!

milanglacier:
> Should we bind `mu4e-headers-{previous next}-thread` to something? They are bound to `{` and `}` by default. I think it makes sense to remap `evil-{forward backward}-paragraph`.
> 
> It might also make sense to bind the new `mu4e-view-detach` to a key.
> 
> Thanks for this PR!

It seems `mu4e-headers-{previous next}-thread` stuffs are added in `mu4e 1.10`. It makes totally sense to remap `evil-{forward backward}-paragraph` to them.

And what key do you think about `mu4e-view-detach` is appropriate? To be honest, I don't know what does this command at all... And in conjunction with `mu4e-view-detach`, there is now a `mu4e-view-attach`, the docstring of these two functions are ambiguous, would you mind talk a little bit about how these two functions do?

edgar-vincent:
In the headers view, when the user opens an email with `mu4e-headers-view-message`, it is displayed in a new window below the current one. If a different email is subsequently opened, it is displayed in the same window, replacing the previous email. If `mu4e-view-detach` is called in a buffer with `mu4e-view-mode` enabled (an email, that is), then when `mu4e-headers-view-message` is called again, the email it displays will be opened in a new window, leaving the current email visible. In that case, two emails will be displayed, each one in its own window.

Calling `mu4e-view-attach` makes the window available for new emails: any email it contains will be replaced by the new one instead of remaining visible - hence restoring the default behaviour.

Perhaps we could bind them to `gd`[etach] and `gD`[attach] (or `ga`[ttach])?

peterhoeg:
As side comment, I would argue that getting support for 1.10.x merged ASAP is more important - quality-of-life improvements and support for new functionality could then be done separately/after.

edgar-vincent:
I agree. Just thought I'd mention the new features here so we can centralise the information pertaining to the new release. The features don't have to be implemented straight away (although, in this case, it would amount to exactly two lines of code).

modes/mu4e/evil-collection-mu4e.el
```diff
@@ -68,349 +68,209 @@
...

-;;; Initialize evil-collection-mu4e
-
-;;;###autoload
-(defun evil-collection-mu4e-setup ()
-  "Initialize evil-mu4e if necessary.
-If mu4e-main-mode is in evil-state-motion-modes, initialization
+(if (version< mu4e-mu-version "1.10")
+    (require 'evil-collection-mu4e
+             (expand-file-name "modes/mu4e/evil-collection-mu4e-1.8" evil-collection-base-dir))
```
meliache, line 73: 1. [ ] Unit tests seem to break due to `mu4e-mu-version` being a void variable. Not sure how this can be fixed. mu4e needs to be loaded to know its version. We could of course cheat with something like
   ``` elisp
   (if (and (boundp mu4e-mu-version) (version< mu4e-mu-version "1.10")) ...
   ```
   But this feels like not a proper solution. Maybe something with `eval-after-load` or something like that?

2. [ ] I What about users of 1.9, which started the transition to the new interface? In my opinion that's officially not a stable version but a development-version, so I wouldn't load any backwards-compatibility code for that. As a non-stable release, people who used that are likely to be users of bleeding-edge mu4e versions and will update soon.

3. [ ] I would add a comment above that `if` explaining why this is being done

meliache:
@milanglacier LGTM, I'm pretty busy so didn't check the progress here often. I put some commends regarding the version-related file loading above: https://github.com/emacs-evil/evil-collection/pull/720#discussion_r1171972906.

The most important remaining issue are the failing unit tests. Initially they failed pointed due to the unbound `mu4e-mu-version`. I added a `boundp` condition as a workaround, but I'm not sure if that's good and whether this might prevent the compatibility code from being loaded if we load evil-collection to early. But despite that hotfix, now other tests file.

I like the suggestions from @edgar-vincent, would be fine with adding them, but I agree we can try to merge this ASAP and then have those as separate PR's.

jojojames:
Lets make sure the compile errors are taken care of before I take another look. Might have to declare a few functions/variables before using it. Can take a look at other examples in this repo.

milanglacier:
I will create a new PR for solving the issues since I want to test how to fix those CI errors.

PR #733, opened 2023-05-29
minibuffer: fix typo

vhqr0:
`evil-collection-minibuffer-maps` is defined but not replaced in `evil-collection-minibuffer-setup`.


condy0919:
LGTM

PR #736, opened 2023-06-14
Fix typo in ement.el

daviwil:
I noticed this typo in the new ement.el module just after it got merged.  I apologize for not actually looking at the code at the time I commented on the PR!


jojojames:
All good, thanks!

PR #738, opened 2023-07-02
evil-collection-elpaca-setup: update command name

progfolio:
elpaca-ui-mark-install has been replaced upstream by elpaca-ui-mark-try.

### Checklist

- [x] byte-compiles cleanly


progfolio:
I've defined an obsolete function alias upstream as well, so nothing should break for users of the previous name.

jojojames:
Thanks!

PR #740, opened 2023-07-12
mu4e: add support for visual actions in header view

Sebatyne:
Hi,

In the headers-view of mu4e, the action functions  "mu4e-headers-mark-for-X" can work on selections. This allows to select many emails fast, and mark them with one key.

I've written a patch to make these actions work in evil visual mode.

Regards,

Nicolas


Sebatyne:
Thanks @condy0919 for the merge !

PR #741, opened 2023-07-24
Set local-buffer value for evil-shift-width

fredericgiquel:
Since commit https://github.com/emacs-evil/evil/commit/b27997b693dca40a08bfea00ae961f10bf839c0c, `evil-shift-width` is no longer buffer-local by default.

So with current versions of `evil` and `evil-collection`, this leads to inconsistent `evil-shift-width` value in some cases. For example:
- open a YAML file (`evil-shift-width` is globally set to `yaml-indent-offset` ; 2 by default)
- then open a Python file (`evil-shift-width` is now globally set to `python-indent-offset` ; 4 by default)
- switch back to the YAML buffer: the global value of `evil-shift-width` is used (here 4) so the shifting operators (> and <) don't work as expected

This PR fix the problem for all modes that set `evil-shift-value`.

### DIFF
--- a/modes/lua-mode/evil-collection-lua-mode.el
+++ b/modes/lua-mode/evil-collection-lua-mode.el
@@ -35,7 +35,7 @@
 
 (defun evil-collection-lua-mode-set-evil-shift-width ()
   "Set `evil-shift-width' according to `lua-indent-level'."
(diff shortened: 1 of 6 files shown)


condy0919:
LGTM

Issue #750, opened 2023-08-31
Possible regression when calling `evil-collection-*-setup`


gagbo:
It seems that de0b62b604e87ce43e2adc9e09bcd5174e8b877a brought a regression

I triggered this by requiring `'profiler`, which in turn triggered the `evil-collection-setup` call, which then triggered either the key definition hook or just the normal setup?

This error gets triggered by anything that's handled by evil-collection it seems.

```
Debugger entered--Lisp error: (cl-no-applicable-method seq-empty-p (normal))
  signal(cl-no-applicable-method (seq-empty-p (normal)))
  cl-no-applicable-method(#s(cl--generic :name seq-empty-p :dispatches nil :method-table nil :options nil) (normal))
  apply(cl-no-applicable-method #s(cl--generic :name seq-empty-p :dispatches nil :method-table nil :options nil) (normal))
  seq-empty-p((normal))
  seq-reduce(#f(compiled-function (acc elt) #<bytecode -0x542a0d6d15ae114>) (normal) nil)
  seq-difference((normal) nil)
  evil-collection--filter-states(normal)
  evil-collection-define-key(normal profiler-report-mode-map " " scroll-up-command [33554464] scroll-down-command [delete] scroll-down-command "j" profiler-report-next-entry "k" profiler-report-previous-entry [tab] profiler-report-toggle-entry "o" profiler-report-ascending-sort "O" profiler-report-descending-sort "c" profiler-report-render-calltree "C" profiler-report-render-reversed-calltree "i" profiler-report-describe-entry "=" profiler-report-compare-profile "\15" profiler-report-find-entry "gr" revert-buffer "q" quit-window "ZQ" evil-quit "ZZ" quit-windw)
  evil-collection-profiler-setup()
  #f(compiled-function () #<bytecode 0x88847b3ea7f1ffb>)()
  eval-after-load(profiler #f(compiled-function () #<bytecode 0x88847b3ea7f1ffb>))
  evil-collection-init((profiler))
```

In any case, I don't have the issue anymore when I check out 1ad283f5b7ac9320ac3d41bccfc71a52f714563a so I assume it's a quoting issue in the change to `evil-collection-translate-minor-mode-key` ? Couldn't dig more into that, I'm just sticking to a commit that works for now.

jojojames:
Was there a need for https://github.com/emacs-evil/evil-collection/commit/de0b62b604e87ce43e2adc9e09bcd5174e8b877a @condy0919 ?

[...]

PR #755, opened 2023-09-17
Fix: Revert de0b62b604e87ce43e2adc9e09bcd5174e8b877a (evil-with-delay change)

gagbo:
EDIT: This PR now fully reverts the problematic commit, while adding some glue to make CI green.

## Old version

This solution is incomplete, and might still fail because of the
impossibility to expand the hook-name dynamically, see the
discussion/screenshots in the PR

### Brief summary of what the PR does

Try to address the issues in #750

### Checklist

- [ ] byte-compiles cleanly

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -676,6 +676,16 @@ modes in the current buffer."
        nil "^[^.]")))))
   (find-file (evil-collection--mode-file mode "README.org")))
 
+(defun evil-collection--delay (condition form hook &optional append local name)
+  "Execute FORM when CONDITION becomes true, checking with HOOK.
+NAME specifies the name of the entry added to HOOK.  If APPEND is
+non-nil, the entry is appended to the hook.  If LOCAL is non-nil,
+the buffer-local value of HOOK is modified.
+
+This is a backport of `evil-delay' without the deprecation notice to deal with CI until migration can be done.
+Ref: https://github.com/emacs-evil/evil-collection/issues/750"
+  (eval `(evil-with-delay ,condition (,hook ,append ,local ,name) ,form) t))
+
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


gagbo:
I mostly tried to solve it by trying different quoting strategies and then using `macroexpand-step` to look at the result.

Before the PR (current state on master):
Screenshot_20230917_113458


Current state as the PR is made:
Screenshot_20230917_120301


# Important notes

- I still didn't test the fixes locally, it's a bit painful to do so yet, I mostly wanted to start a conversation about what we should expect the macro to expand to.
- Note that in the before screenshot, the name of the hook is unique (`evil-delay-in-{backquote}`), which might hint that the issue is indeed that the quoting is off with the `master` version of `evil-collection-translate` calls.
- _Also_ note that I couldn't find a way to properly expand the value of `hook-name` in the resulting expansion, so I don't really know how to unlock this part. In a scratch buffer, calling `macroexp-quote` with a dynamic var seems to work:

image

So I'm not sure how I can go further and fix that. Is the change of `evil-mode` actually breaking after all? Or is there a quoting pattern that would fix this issue?

jojojames:
I think we should have a reverse commit of the original commit that caused the regression and then we add your fixes on top of the reverted commit. This way, we can evaluate the necessary changes from the original (deprecated) method and your new changes.

gagbo:
Probably. I just made the revert commit this way as a compromise to keep the CI green. If it's not important, just reverting is probably going to be enough.

I'm not well-versed in macro magic, but I had a lot of issues trying to get the custom `evil-collection-translate-in-%s` name in the new `evil-with-delay` macro.

condy0919:
If there is no further discussion, I will merge 2 days later. :-)

PR #758, opened 2023-10-05
diff-hl: fix name incompatible

JiaweiChenC:
Issue: evil keybinding doesn't work with diff-hl popup


JiaweiChenC:
The maintainer of the diff-hl package changed the variable name from diff-hl-show-hunk--inline-popup-map to diff-hl-show-hunk-map. We should follow the name.
image
Link: https://github.com/dgutov/diff-hl/commit/2ded28eacca0df527c55759838357b3e56d9b711#diff-3eb89b676c56131c137e621b931595b62274106922bdcbb718141cf8a0fa2969

PR #760, opened 2023-10-22
feat(cider): add force-out, inspect-prompt, trace 

aisamu:
This is not a new package, just additions to `cider--debug-mode-map`'s map

#### feat(cider): add force-out, inspect-prompt, trace <sup>c70a83c</sup>
Matches cider's existing keybindings on `cider--debug-mode-map`

#### refactor(cider): match cider-debug.el command order <sup>89a317a</sup>
Since it's already not intentionally ordered (e.g alphabetical),
matching upstream's order helps when eyeballing additions.

Thank you for evil collection 🥳!


aisamu:
(The combined diff is subpar - the per-commit diffs linked on the description are way more useful!)

PR #765, opened 2023-12-18
fix bufler RET keybind

### LINKED ISSUE #764: Bufler RET binding is incorrect
This is an issue with a recent change. You can see here that the `RET` key is incorrectly bound: https://github.com/emacs-evil/evil-collection/blob/d58233be7dd51342970da3dad0428ea0e7af304c/modes/bufler/evil-collection-bufler.el#L53.

Should be (kbd "RET") , instead of "RET", which binds a sequence of R E T keys.

FYI @Vonfry.

Vonfry:

fix #764 

https://github.com/your/awesome_package

### Checklist

- [ ] byte-compiles cleanly


yoav-za (approved):
Thank you!

PR #767, opened 2023-12-25
Fix the new `corfu--setup` signature (backward compatible)

abougouffa:
This PR fixes this issue reported in `corfu`'s repository:

https://github.com/minad/corfu/issues/403

It is caused by a change of signature of the `corfu--setup` function, causing the advice defined by `evil-collection-corfu` to call the advised function with the wrong number of arguments.

This PR fixes this while staying backward compatible with the previous `corfu--setup` signature.


abougouffa:
Thank you for the info @minad 

I've opened another PR to fix this (#769)

minad:
Abdelhak Bougouffa ***@***.***> writes:

> Thank you for the info @minad

You're welcome.

> I've opened another PR to fix this (#769)

Thanks!

PR #770, opened 2024-01-03
Add bindings for new vundo diff operations.

Dickby:
Adds bindings for the new vundo commands `vundo-diff` (d), `vundo-diff-mark`(m) and `vundo-diff-unmark`(u).
`vundo--debug` was bound to (d), because it's an intern debug function, i changed that binding to (D).


condy0919:
LGTM

Issue #776, opened 2024-02-05
CIDER: Use conditionaly use `xref-find-definitions` in the goto definitions keybind 


toniz4:
Starting in at version 1.2.0, CIDER's main method of finding definitions is using the builtin emacs `xref-find-definitions`, unless the variable `cider-use-xref` is nil, the default being `t`.

It would be nice if we could mirror this functionality in evil collection.  

https://github.com/emacs-evil/evil-collection/blob/29212d123f1d2622332c02ca55b3a81faee1af76/modes/cider/evil-collection-cider.el#L135

If it's something you wish to add, I will be happy to make a PR.

condy0919:
PR welcome.

PR #782, opened 2024-03-01
CIDER: Only use cider-find-var when cider-use-xref is nil

toniz4:
Closes: #776

Follows the behavior of `cider-use-xref` in CIDER. When `cider-use-xref` is non nil (the default), it doesn't bind `cider-find-var` nor `cider-pop-back`.

Documentation of this feature in the CIDER docs: https://docs.cider.mx/cider/usage/misc_features.html#xref-integration


condy0919:
Thanks

PR #783, opened 2024-03-03
corfu: bind up/down in insert state

LemonBreezes:
Corfu already binds up/down in emacs state, but not in insert state. Normally this is not a problem, but when using corfu in comint-mode like with chatgpt-shell, up/down moves in the history instead of moving in the corfu completions.


condy0919:
LGTM

Issue #781, opened 2024-02-27
`mu4e-compose-goto-top` and `mu4e-compose-goto-bottom` have been removed from `mu4e`


LemonBreezes:
Hello. These commands `mu4e-compose-goto-top` and `mu4e-compose-goto-bottom` have been removed in the latest release of `mu4e` but are still referenced by the package.

Also referencing a Doom Emacs PR re-adding these functions: https://github.com/doomemacs/doomemacs/pull/7696.

condy0919:
PR welcome.. I'm not using mu4e for a long time.

PR #785, opened 2024-03-05
mu4e: re-add removed mu4e-compose-begin/end-commands

LemonBreezes:
The commands `mu4e-compose-begin` and `mu4e-compose-end` were removed in commit
https://github.com/djcb/mu/commit/85bfe763362b95935a3967f6585e14b3f9890a70 of version 1.12.0 of mu.  This commit re-adds these commands to the evil collection for mu4e.

### DIFF
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -93,6 +93,31 @@
       (evil-set-initial-state mode 'normal))
     (evil-set-initial-state 'mu4e-compose-mode 'insert))
 
+  ;; These functions were removed from mu4e 1.12.0, specifically this commit:
+  ;; https://github.com/djcb/mu/commit/85bfe763362b95935a3967f6585e14b3f9890a70
+  (when (version<= "1.12.0" mu4e-mu-version)
+    (defun mu4e-compose-goto-top ()
+      "Go to the beginning of the message or buffer.
+Go to the beginning of the message or, if already there, go to the
+beginning of the buffer."
+      (interactive)
+      (let ((old-position (point)))
+        (message-goto-body)
+        (when (= (point) old-position)
+          (goto-char (point-min)))))
+
+    (defun mu4e-compose-goto-bottom ()
+      "Go to the end of the message or buffer.
+Go to the end of the message (before signature) or, if already there, go to the
+end of the buffer."
+      (interactive)
+      (let ((old-position (point))
+            (message-position (save-excursion (message-goto-body) (point))))
+        (goto-char (point-max))
+        (when (re-search-backward message-signature-separator message-position t)
+          (forward-line -1))
+        (when (= (point) old-position)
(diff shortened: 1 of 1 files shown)


dschrempf:
Thanks!

Apparently an interactive functin needs to be changed to a non-interactive one:
```
modes/mu4e/evil-collection-mu4e.el:119:30:Error: ‘beginning-of-buffer’ is for interactive use only; use ‘(goto-char (point-min))’ instead.
```

LemonBreezes:
> Thanks!
> 
> Apparently an interactive functin needs to be changed to a non-interactive one:
> 
> ```
> modes/mu4e/evil-collection-mu4e.el:119:30:Error: ‘beginning-of-buffer’ is for interactive use only; use ‘(goto-char (point-min))’ instead.
> ```

Ok! I just fixed it.

meliache:
Great, I stumbled upon the same issue and cool that you fixed it already. This PR fixes #781, which you certainly know since you opened it, but it's lot linked to the issue, so I hereby mention it so that people who find the issue first find a link to this PR.

The removal seems to have been a by-product of trying to follow more closely Gnus and its message-mode, which has its own functions to go to the signature, body, etc, so maybe these mu4e-specific functions were deemed unnecessary. Thinking of that, we might even consider porting these functions to  evil-collections-gnus. But for now, I think it's fine keeping it here in mu4e only as it's consistent with its old behavior.

modes/mu4e/evil-collection-mu4e.el
```diff
@@ -93,6 +93,31 @@
...
+
+    (defun mu4e-compose-goto-bottom ()
+      "Go to the end of the message or buffer.
+Go to the end of the message (before signature) or, if already there, go to the
+end of the buffer."
+      (interactive)
+      (let ((old-position (point))
+            (message-position (save-excursion (message-goto-body) (point))))
+        (end-of-buffer)
+        (when (re-search-backward "^-- $" message-position t)
```
meliache, line 117: Why not use `message-goto-signature`, like how you used `message-goto-body` above? It also takes into account that `message-signature-separator` can be different from `--`.

LemonBreezes: Yes, I copied the code for those Mu4e commands verbatim from the original source. Now I have replaced that `^-- $` with `message-signature-separator`. It's similar to `message-goto-signature` except it goes to the line above the signature rather than the line below and it also does not push the mark. This seems like better behavior to me for Evil but I am interested in what you think about it now.

meliache: Haven't tested it but seems fine to me :+1: 

Btw, I was thinking, then doesn't `message-goto-body` used in `mu4e-compose-goto-top` also push the mark? Checked the source code and it has a special clause that it only pushes the mark if called interactively, but `message-goto-signature` doesn't have that, which tbh. seems inconsistent behavior to me...

condy0919:
Good job

Issue #683, opened 2022-10-06
Improve pdf-selection behavior


dalanicolai:
I would like to notify you about some small improvement regarding selection in pdf-tools (see here https://github.com/vedang/pdf-tools/issues/137).

In summary: please add the 'mouse bindings' for the 'pdf-view-mode-map' 'normal-state' also to its 'visual-state' map (see https://github.com/syl20bnr/spacemacs/pull/15762/files).

Thanks!

condy0919:
Thanks

condy0919:
repost link https://github.com/syl20bnr/spacemacs/pull/15762

jojojames:
@dalanicolai Looking at the links in this thread, it seems you would be the most appropriate person to add those bindings. I am not sure exactly which bindings you'd prefer on which keys. 

PR will definitely be appreciated here, thanks for the investigation.

PR #794, opened 2024-03-16
pdf: disable visual-mode

45mg:
Evil's visual-mode does not know how to select text within a PDF, and instead selects the entire pdf image, which is useless and counterintuitive - see doomemacs/doomemacs#6286, emacs-evil/evil-collection#683. Instead, rely on pdf-view to handle the region.

### DIFF
--- a/modes/pdf/evil-collection-pdf.el
+++ b/modes/pdf/evil-collection-pdf.el
@@ -98,6 +98,18 @@ Consider COUNT."
      (or reg ?\")
      (mapconcat #'identity txt nil))))
 
+(defun evil-collection-pdf-disable-visual-mode ()
+  "Don't enter visual-mode when the mark is activated.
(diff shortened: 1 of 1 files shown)


condy0919:
I think it has conflicts with #795 . What's your thought?

45mg:
It does. I had initially intended to submit #794 and #795 as a single PR, but they're separate changes and I thought you might want to consider them separately.
I don't really know what one generally does in these situations. I've attached a patch file to indicate how the conflict should be resolved; let me know if you'd like me to address this another way.
[794_795.patch.txt](https://github.com/emacs-evil/evil-collection/files/14650476/794_795.patch.txt)
(GitHub won't let me upload a .patch for some reason)

condy0919:
Let's fix the conflicts.

45mg:
Ok, I rebased onto master after you merged #795, so this is ready to merge.

PR #798, opened 2024-03-23
Corfu: evil-repeat support

aaronjensen:
Support for `evil-repeat` with Corfu.

See:

https://github.com/minad/corfu/pull/169
https://github.com/minad/corfu/pull/225

### DIFF
--- a/modes/corfu/evil-collection-corfu.el
+++ b/modes/corfu/evil-collection-corfu.el
@@ -38,8 +38,9 @@
 (defvar corfu-preselect-first)
 (defvar corfu--index)
 
-(declare-function "corfu-insert" "corfu")
-(declare-function "corfu-reset" "corfu")
+(declare-function corfu-insert "corfu")
+(declare-function corfu-reset "corfu")
+(declare-function corfu-quit "corfu")
 
 (defconst evil-collection-corfu-maps '(corfu-map))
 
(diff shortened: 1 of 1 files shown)


condy0919:
```
modes/corfu/evil-collection-corfu.el:157:8: Error: the function ‘corfu-quit’ is not known to be defined.
```

aaronjensen:
Yeah, that's odd. Any ideas?

aaronjensen:
By the way, the Makefile doesn't actually work. It expects evil to be installed, but doesn't install it. Do you have instructions for running tests locally? That would allow me to iterate locally.

aaronjensen:
There, that should work. The `declare-function` forms were incorrect.

asymmetric:
The two listed corfu PRs were not merged, how does this affect whether this PR should be merged?

aaronjensen:
It's the reason that I'm putting the code here. The author would prefer a different technique to support evil-repeat, but I do not know of any additional techniques. In other words, it not being merged there, is additional justification for it being merged here. I don't like having to patch one of corfu's methods, but there is no other option yet (aside from not supporting evil-repeat, of course).

modes/corfu/evil-collection-corfu.el
```diff
@@ -89,7 +107,7 @@ This key theme variable may be refactored in the future so use with caution."
   ;; https://github.com/minad/corfu#tab-and-go-completion
   (when (memq 'tab-n-go evil-collection-corfu-key-themes)
     (setq corfu-cycle t
-          corfu-preselect-first nil)
+          corfu-preselect 'prompt)
```
aaronjensen, line 110: This changed a bit ago, so `corfu-preselect-first` didn't actually exist anymore.

aaronjensen:
By the way, if there's any other way to do this, either by modifying evil-repeat or another way to make corfu compatible, I'd be interested. I started a thread on Emacs-devel, but haven't heard anything yet.

aaronjensen:
@condy0919 updated

PR #803, opened 2024-04-06
corfu: don't suppress popup when evil is disabled

45mg:
PR #800 made it impossible to use Corfu whenever Evil wasn't initialized (eg. in the minibuffer when `evil-want-minibuffer` is nil). This fixes that.

Amend: #800


45mg:
Never mind, duplicate of https://github.com/emacs-evil/evil-collection/pull/802.

PR #804, opened 2024-04-09
ci: Use Eask to test macos and windows

jcs090218:
Cask is deprecated and doesn't support Windows. [Eask](https://github.com/emacs-eask/cli) is the new replacement for Cask.

This patch does the following:

1. Add test for macOS and Windows
2. Clean up most `Makefile` logic to Eask-file

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,34 +7,47 @@ on:
     branches:
       - master
 
+concurrency:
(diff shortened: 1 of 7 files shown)


jojojames:
Looks good to me. Thanks @jcs090218 !

jcs090218:
Can someone re-run the tests? It seems like there is a network error. 🤔

PR #805, opened 2024-04-14
replace: Bind q to quit-window when not editing

bcc32:
q is bound by default to dismiss the window, which is useful when
occur is being used only to navigate occurrences and not edit them.

### DIFF
--- a/modes/replace/evil-collection-replace.el
+++ b/modes/replace/evil-collection-replace.el
@@ -58,6 +58,7 @@
     (kbd "C-k") 'previous-error-no-select
     "r" 'occur-rename-buffer
     "c" 'clone-buffer
+    "q" 'quit-window
     (kbd "C-c C-f") 'next-error-follow-minor-mode)
 
   (evil-collection-define-key 'normal 'occur-edit-mode-map


condy0919:
Occur is derived from special-mode, and evil-collection has already bound q to quit-window in normal state.

bcc32:
The original motivation for this PR was to be able to use this mode in Spacemacs.  Spacemacs currently only enables a small subset of the modes in `evil-collection-mode-list`, and certainly doesn't have `q` bound to quit-window in occur buffers at the moment.  (I worry that enabling `simple` in that list would have lots of other possibly-undesirable side-effects since Spacemacs already sets up its own key bindings for many modes that derive from special-mode).

Perhaps `evil-collection-set-readonly-bindings` should be used instead, would you recommend using that here?

condy0919:
Could you suggest enabling `simple` for Spacemacs? Or tons of modes derived from special-mode should have an extra `evil-collection-set-readonly-bindings` .

bcc32:
That's a good point, thanks.  I've incorporated it into this Spacemacs [PR](https://github.com/syl20bnr/spacemacs/pull/16371), in case you're curious.

PR #813, opened 2024-05-15
Add .nosearch to modes/ to not litter subdirs.el

knazarov:
If you install evil-collection to site-lisp/, it's getting picked by Emacs and added to `subdirs.el` in order to populate the `load-path` at startup. To quote the official documentation:

> For each directory in load-path, Emacs then checks to see if it contains
> a file subdirs.el, and if so, loads it. The subdirs.el file is created when
> Emacs is built/installed, and contains code that causes Emacs to add any
> subdirectories of those directories to load-path. Both immediate subdirectories
> and subdirectories multiple levels down are added. But it excludes subdirectories
> whose names do not start with a letter or digit, and subdirectories named RCS
> or CVS, and subdirectories containing a file named .nosearch.

There are legitimate reasons why one may add evil-collection to site-lisp/: Linux distributions packaging it as native distro packages may lead to this for example.

In my case, I'm using Nix and "Emacs Overlay", which is more or less a standard way to manage Emacs and its configuration under Nix. I've noticed that my Emacs starts slower than needed (0.6 seconds with configuration under Nix, and 0.2 seconds with configuration not managed by Nix). I've pin-pointed the issue to an explosion of `load-path`, which is one of the biggest performance hogs on Emacs startup.

It seems that:

- Many other packages add `.nosearch` to their subdirs to prevent automatic addition to load-path
- There is no risk in introducing this to evil-collection, as it loads modules by absolute paths from the basedir of the main file.

I've patched evil-collection and on my system it keeps working, while achieving a nice load speedup (almost 0.4 seconds). Please consider merging this.


jojojames:
Could you expound on the implications of .nosearch on other packages. As far as I'm aware some packages explicitly skip directories with nosearch in them (for ex, if I were to try to search/grep over the evil-collection directory). It's been a while since I looked at nosearch in relation to these search tools and I'm a little hesitant due to (don't fix what's not broken).

This change isn't offensive to me though, I'm just a little wary as it might break existing workflows (for me).

knazarov:
What I do know is that quite a few other packages in melpa with lots of subdirs do this. Take yasnippet for example:
https://github.com/AndreaCrotti/yasnippet-snippets/blob/master/snippets/.nosearch

So this seems to be a common practice. 

ripgrep and fd don't seem to care about the presence of .nosearch (I've just tested on my local system)

jojojames:
Sounds good.

Issue #810, opened 2024-04-18
`d` in diff-mode + visual selection does not kill text


Hi-Angel:
Before trying to send a PR to fix that I decided to create this issue to discuss.

The problem case: I was editing a diff file. I have selected a large part of the diff that covers multiple files and pressed <kbd>d</kbd> to remove it. Then I saved and applied it. Now, can you imaging my surprise when I found that literally nothing has changed. The files I removed were still there, in git. Upon closer inspection turns out that `evil-collection` [has bound this key to some mode-specific action `diff-hunk-kill`](https://github.com/emacs-evil/evil-collection/blob/5886bab852dc9e31959e70384d535473e44504ad/modes/diff-mode/evil-collection-diff-mode.el#L125)

I see 2 problems with that:

1.That seems to kind of beat the purpose of the plugin because from my understanding the plugin is trying to help with various annoying modes that override common Evil keybindings. But in this case it is the plugin that overrides it.
2. Even setting `1` aside, why <kbd>d</kbd> in visual mode? It is more confusing than it is helpful, because visual mode implies you've selected some region and so `diff-hunk-kill` should do the action over the region. Which `diff-hunk-kill` does not support.

Should the hotkey perhaps be removed?

Hi-Angel:
CC: @Ambrevar the commit author

jojojames:
Feel free to submit a PR to discuss better keybindings. What you're suggesting doesn't seem like a bad idea at an initial glance.

Hi-Angel:
@jojojames I am confused. There was a PR that is closing this issue that has undergone multiple changes and was stuck on discussion. You closed it because it's currently not ready to merge, okay. But then, the only thing I can do here is to open a PR that simply removes the `d` keybinding from the code, because the better way which is  "only remove it for visual state but leave it for normal" just doesn't work for whatever reason. So what do I need to do here?

jojojames:
> So what do I need to do here?

You can reopen the issue in the PR. I closed this issue because the topic of this issue is no longer relevant (e.g. can be handled in the PR)

The PR being closed can be reopened as long as there's further interest.

(I also added a change here https://github.com/emacs-evil/evil-collection/commit/22e45c6939c96b4da4cc2aba628efdfedfa887e9 that might fix (or worsen) some things for you.)

Hope you can try it out and we can go from there.

Hi-Angel:
> (I also added a change here 22e45c6 that might fix (or worsen) some things for you.)

[...]

PR #815, opened 2024-05-24
diff-mode: disable single-letter cmds w/o respect to selection in "visual"

Hi-Angel:
As mentioned in the original issue, the keybinding is confusing and problematic. There has been no opposing comments for 1.5 months, so let's go ahead and ~~remove~~ make it to only work in `normal`.

Fixes: https://github.com/emacs-evil/evil-collection/issues/810

### DIFF
--- a/modes/diff-mode/evil-collection-diff-mode.el
+++ b/modes/diff-mode/evil-collection-diff-mode.el
@@ -103,6 +103,13 @@ current file instead."
 
     "q" 'quit-window
 
+    "a" 'diff-apply-hunk
+    "*" 'diff-refine-hunk
+    "D" 'diff-file-kill
+    "d" 'diff-hunk-kill
+    "s" 'diff-split-hunk
+    "c" 'diff-test-hunk
+
     "\\" 'read-only-mode) ; magit has "\"
 
   (evil-collection-define-key 'motion 'diff-mode-map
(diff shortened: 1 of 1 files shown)


condy0919:
@jojojames What's your thoughts?

jojojames:
Wouldn't you want to move the key somewhere else than to remove it entirely?

Hi-Angel:
Sorry, I didn't know what modes `motion` includes. Is it just `normal` and `visual`? If that's the case, then indeed your idea is a good one, moreover, I think it's better to move every single-letter command that does not respect selection to under the "normal".

Hi-Angel:
It's just that I don't know, the `motion` state does not seem to be documented. There's [this answer](https://emacs.stackexchange.com/questions/30813/differences-between-evil-normal-state-map-and-evil-motion-state-map) which implies it is completely unrelated to "normal" and "visual" and instead it is for read-only buffers like `*Help*`. But that doesn't make sense to me because from the bug we can clearly see that it influences visual state in an editable buffer/major-mode.

jojojames:
motion -> normal is fine.

Hi-Angel:
@jojojames done, please see if it looks okay now

Hi-Angel:
Actually, wait, I see another command that may be confusing when used in "visual"

Hi-Angel:
@jojojames okay, so, I'm afraid your idea with moving stuff under the "normal" is not working due to a bug somewhere. When I'm testing the current changes, while pressing <kbd>d</kbd> in "visual" I still see the hunk getting deleted even though it is no longer under "motion".

jojojames:
Visual state is different from normal state. The changes here seem ok for fixing normal state&motion state for now. If you want to later fix visual state, we can figure that out separately.

Closing this PR, but feel free to reopen if you think this is in a good state to merge.

Issue #637, opened 2022-05-26
help menu in magit is not displayed properly


cs50Mu:
Press ? in `magit-status` gives the following in which some keybindings are wrong:
<img width="819" alt="image" src="https://user-images.githubusercontent.com/4456781/170499675-a1e7e365-9ec8-4b48-8474-104ac59e5504.png">

for example, Discard should be `x` instead of `k`; Reset should be `O` instead of `X`

`magit-version` gives:  

<img width="757" alt="image" src="https://user-images.githubusercontent.com/4456781/170500467-d111aea3-39a8-44ad-8943-f7da4ad70652.png">

jojojames:
I see 'x' for discard and 'O' for Reset. Can you provide an emacs-q recipe?

cs50Mu:
> I see 'x' for discard and 'O' for Reset. Can you provide an emacs-q recipe?

I'm an emacs noob, did you mean run emacs with `emacs -Q` in terminal?

jojojames:
Yeah, I added a section here:

https://github.com/emacs-evil/evil-collection#submitting-issues

emacs -Q in terminal, then you can probably eval that snippet and test from there.

cs50Mu:
> Yeah, I added a section here:
> 
> https://github.com/emacs-evil/evil-collection#submitting-issues
> 
> emacs -Q in terminal, then you can probably eval that snippet and test from there.

I've reproduced the issue with the following steps:

- start emacs with `emacs -Q`
- in the scratch buffer, paste the code snippet you just provided: [submitting-issues](https://github.com/emacs-evil/evil-collection#submitting-issues)
- add this line in the buffer: `(use-package magit)`
- M-x and run the command: eval-buffer
- open a file in a git project
- press `C-x g` to show the magit-status panel
- press `?` to show the help menu
- and it still show the wrong keybindings

<img width="657" alt="image" src="https://user-images.githubusercontent.com/4456781/170832470-090a3dd2-1b52-489d-85cc-bfcc2747b14e.png">

emacs version: I installed emacs from [emacsformacosx](https://emacsformacosx.com/)  
operating system: macOS Monterey 12.2.1

jojojames:
Can you paste the exact emacs -Q recipe you used?

e.g. what you eval'ed

cs50Mu:
> Can you paste the exact emacs -Q recipe you used?
> 
> e.g. what you eval'ed

```
[...]

Issue #510, opened 2021-07-27
Evil-collection bindings for magit-repolist-mode don't work until magit is fully loaded


fredericgiquel:
## What happened?

In my Emacs configuration, I defer the loading of `magit` with `use-package` (very common practice). If I open the magit repositories buffer (with `magit-list-repositorities` command) just after the launch of Emacs, here is the situation :
- the evil state of the buffer is `normal` (instead of `motion`)
- `<RET>` is bound to `evil-ret` (and not to `magit-repolist-status`)

I must first use `magit-status` with a git repo to have the intended bindings for `magit-repolist-mode`.

## Steps to reproduce

I tried to make a minimal init.el to reproduce the problem :
```emacs-lisp
;; Initialize package sources
(require 'package)

(setq package-archives '(("melpa" . "https://melpa.org/packages/")
                         ("elpa" . "https://elpa.gnu.org/packages/")))

(package-initialize)
(unless package-archive-contents
  (package-refresh-contents))

(unless (package-installed-p 'use-package)
  (package-install 'use-package))

(require 'use-package)
(setq use-package-always-ensure t)

;; Evil + evil-collection
(use-package evil
  :init
  (setq evil-want-integration t) ;; This is optional since it's already set to t by default.
  (setq evil-want-keybinding nil)
  :config
  (evil-mode 1))

(use-package evil-collection
  :after evil
  :config
  (evil-collection-init))

;; Deferred load of magit
(use-package magit
  :init
  (setq magit-repository-directories '(("~/.emacs.d" 0)))
  :bind ("C-x g" . magit-status))
```
2 important things to note :
- the `magit-repository-directories` must be initialized with a correct value. I used "~/.emacs.d" just for the example (and because it's a git repo for many Emacs users).
- the `:bind ("C-x g" . magit-status)` line is very important. Without it, the package load is not deferred and the problem does not occur.

Action to do : Launch Emacs with this init file and do `M-x magit-list-repositorities` (before anything else)

## Other informations

The problem can be reproduced either with melpa packages or with last versions from git repositories (using straight.el).

fredericgiquel:
I solved the problem in my configuration by adding `(with-eval-after-load 'magit-repos (evil-collection-magit-setup))` or `(add-to-list 'evil-collection-mode-list '(magit magit-repos))` before `(evil-collection-init)`.

Maybe something similar can be integrated directly into the package to solve the problem for everyone.

jojojames:
[...]

PR #819, opened 2024-06-08
magit: Separate mode for magit-repos

fredericgiquel:
The fix for https://github.com/emacs-evil/evil-collection/issues/637 reintroduced the problem reported in https://github.com/emacs-evil/evil-collection/issues/510.

This PR extracts magit-repos bindings from magit mode by creating a new separate mode. This should make possible to load correctly magit-repos bindings without interfering with magit main bindings loading process.


condy0919:
Thanks

PR #820, opened 2024-07-08
notmuch: fix `RET` keybinding in `notmuch-hello-mode`.

milanglacier:
In the previous version, pressing "RET" (i.e., calling `evil-collection-notmuch-hello-ret`) on a recent searches to enter a search mode buffer, then pressing `q` to quit the search buffer and return to the hello buffer, would result in all evil state keybindings in `notmuch-hello-mode-map` being deactivated. Rewriting the `evil-collection-notmuch-hello-ret` function resolves this issue.

### DIFF
--- a/modes/notmuch/evil-collection-notmuch.el
+++ b/modes/notmuch/evil-collection-notmuch.el
@@ -36,6 +36,10 @@
 (declare-function notmuch-search-tag "notmuch")
 (declare-function notmuch-tree-tag "notmuch-tree")
 
+(declare-function widget-field-at "wid-edit")
+(declare-function widget-field-activate "wid-edit")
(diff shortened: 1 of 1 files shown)


milanglacier:
I don't know why the CI on ubuntu yield the following compile error:

```
Compiling /home/runner/work/evil-collection/evil-collection/modes/magit-repos/evil-collection-magit-repos.el... git exited with status 1
```

I didn't modify `evil-collection-magit-repos.el` at all. It seems that the CI failure is just a false positive, and it has nothing related to my PR.

condy0919:
Be patient. I will review this patch after https://github.com/purcell/setup-emacs/issues/49 fixed

PR #830, opened 2024-08-17
dape: Add initial support

svaante:

Dape is an debug adapter client for emacs, like lsp but for debugging.

https://github.com/svaante/dape

### Checklist

- [x] byte-compiles cleanly

### DIFF
--- a/modes/dape/evil-collection-dape.el
+++ b/modes/dape/evil-collection-dape.el
@@ -1,0 +1,54 @@
+;;; evil-collection-dape.el --- Bindings for `dape' -*- lexical-binding: t -*-
+
+;; Copyright (C) 2024 Daniel Pettersson
+
+;; Author: Daniel Pettersson <daniel@dpettersson.net>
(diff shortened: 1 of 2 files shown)


modes/dape/evil-collection-dape.el
```diff
@@ -0,0 +1,54 @@
...
+(require 'dape nil t)
+
+(defconst evil-collection-dape-maps '(dape-info-watch-edit-mode-map
+                                      dape-info-scope-mode-map
+                                      dape-info-watch-mode-map))
+
+;;;###autoload
+(defun evil-collection-dape-setup ()
+  "Set up `evil' bindings for `dape'."
+  (evil-make-overriding-map dape-info-scope-mode-map)
```
jojojames, line 39: There's no reason we couldn't manually bind these right? To avoid using override maps.

jojojames:
Looks good other than the override maps but that's a nit.

PR #832, opened 2024-08-26
remove `(defvar mu4e-mu-version)`

RyanGibb:
Without this change evil-collection fails to pick up the mu4e-mu-version set in the mu4e module.

I'm new to emacs lisp but I think we're overriding the `mu4e-mu-version` variable set in the mu4e module with one set in this module to `nil`.

Any feedback or advice is welcome if this is the wrong solution.

### DIFF
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -69,7 +69,6 @@
 (require 'message)
 (require 'mu4e nil t)
 
-(defvar mu4e-mu-version)
 (declare-function evil-collection-mu4e-set-bindings "evil-collection-mu4e")
 (declare-function evil-collection-mu4e-set-state "evil-collection-mu4e")
 (declare-function mu4e-headers-mark-thread "mu4e-headers")


condy0919:
It's not overriding. `(defvar mu4e-mu-version)` declares the variable.

RyanGibb:
Is it not already declared at https://github.com/djcb/mu/blob/aee76e47a86cb31edcf4ba70c0af9547db220e8c/mu4e/mu4e-config.el.in#L3 ?

evil-collection-mu4e.el does not get the version of mu without this change for me.

RyanGibb:
It looks like this was introduced in https://github.com/emacs-evil/evil-collection/commit/d3b497db3172a34cbe9e51269fb6af57925303b3 to fix the bytecode compilation `Compiling /home/runner/work/evil-collection/evil-collection/modes/mu4e/evil-collection-mu4e.el... reference to free variable ‘mu4e-mu-version’`, and using `defvar` seems to be the correct way to solve this. However, with this defvar evil-collection-mu4e.el fails to pick up the versioned defined in mu4e.

RyanGibb:
Ah, it looks like that I actually needed was https://github.com/emacs-evil/evil-collection/pull/812/commits/149eacce58354f0ee3a55d4c12059148ef4ff953. Sorry for the noise.

PR #833, opened 2024-08-26
fix(view-mode): use `#'View-quit`

gekoke:
view-mode itself recommends using this function to quit the mode in
in a message in its source code:

```
View mode: type \\[help-command] for help, \\[describe-mode] for commands, \\[View-quit] to quit.
```

Using `#'quit-window` does not work as expected when viewing a diff using view-mode for a buffer in a save prompt in magit: `#'quit-window` breaks the prompting sequence, whilst `#'View-quit` doesn't.


condy0919:
Thanks.

PR #849, opened 2024-12-13
citre: Initial support

condy0919:


jojojames:
Think you want to add it https://github.com/emacs-evil/evil-collection/blob/950f7cfe9bf8d863ebc6abf4f90777e1dcbb028f/evil-collection.el#L144 also

condy0919:
oh, I forgot that. Thanks.

PR #853, opened 2024-12-22
fix: the goto-reference in slime-mode only works for global variables.

sakurawald:

Fix the binding of `goto reference` in `slime mode`.
The command `slime-who-references` **is not the proper command** to find the references **on a symbol**.
Actually, the `slime-who-references` only works on **global variables**, it doesn't works for **functions**, **methods** and **class**.
As documented:
```
slime-who-references is an interactive byte-compiled Lisp function in
‘slime.el’.

It is bound to g r, <normal-state> g r, C-c C-w C-r and C-c C-w r.

(slime-who-references SYMBOL)

Show all known referrers of the global variable SYMBOL.

[back]
```

The command `slime-edit-uses` is an all-in-one command, that does combines all the commands to find the references **on a symbol**, including: `a variable`, `a function`, `a method`, `a class`, `a specializer`.

As documented:
```
slime-edit-uses is an interactive byte-compiled Lisp function in
‘slime.el’.

It is bound to M-? and M-_.

(slime-edit-uses SYMBOL)

Lookup all the uses of SYMBOL.

[back]
```

https://github.com/slime/slime

### Checklist

- [x] byte-compiles cleanly

### DIFF
--- a/modes/slime/evil-collection-slime.el
+++ b/modes/slime/evil-collection-slime.el
@@ -163,7 +163,7 @@
 
   (when evil-collection-want-find-usages-bindings
     (evil-collection-define-key 'normal 'slime-mode-map
-      "gr" 'slime-who-references))
+      "gr" 'slime-edit-uses))
 
   (evil-collection-define-key 'normal 'slime-popup-buffer-mode-map
     ;; quit


condy0919:
> >   - [3/4] Installing package-lint (20241127.182[6](https://github.com/emacs-evil/evil-collection/actions/runs/12450929770/job/35176323329#step:6:7))... https://elpa.gnu.org/packages/let-alist-1.0.6.tar: Bad Request

This problem often occurs and can be ignored

PR #858, opened 2025-01-30
gptel: add gptel-context-buffer-mode-map

LemonBreezes:
Adding keybindings for `gptel-context-buffer-mode-map`.


LemonBreezes:
The failing checks related to `magit-forge` are not related to this PR.

PR #860, opened 2025-03-20
fix prefix key collision for mu4e

LoganBarnett:
This reverts commit 05731c551be8cdda40ae6479adfb30b7e9c7fe39.

This fixes an issue that arises when loading `evil-collection-mu4e`, which gives:

```
Key sequence c c starts with non-prefix key c
```

This occurs for both `mu4e-headers-mode-map` and `mu4e-main-mode-map`.

This should probably just be a different key, but for now I'm just trying to get it unbroken.

cc @RyanGibb for visibility.

Thanks for the wonderful library!

### Checklist

- [x] byte-compiles cleanly

### DIFF
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -158,8 +158,7 @@ end of the buffer."
        "f" smtpmail-send-queued-mail
        "m" mu4e--main-toggle-mail-sending-mode
        "s" mu4e-search
-       "q" mu4e-quit
-       "c" mu4e-search-query)
+       "q" mu4e-quit)
 
       (mu4e-headers-mode-map
        "q" mu4e~headers-quit-buffer
@@ -216,8 +215,7 @@ end of the buffer."
        "zd" mu4e-headers-toggle-skip-duplicates
        "gl" mu4e-show-log
        "gv" mu4e-select-other-view
-       "T"  evil-collection-mu4e-mark-thread-as-read
-       "c" mu4e-search-query)
+       "T"  evil-collection-mu4e-mark-thread-as-read)
 
       (mu4e-compose-mode-map
        "gg" mu4e-compose-goto-top


RyanGibb:
Hmm, do you know why this error is occurring? Is it colliding with something in evil-collection or something in your own config? Upstream uses =c= https://github.com/djcb/mu/commit/d5129052df2b1db020465f84ec9c180a3e980405, do you have a suggestion for another key?

paniash:
Hi, I've also noticed this issue with evil-collection.
```
evil-collection: error setting key in mu4e-headers-mode-map (error "Key sequence c c starts with non-prefix key c")
```

LoganBarnett:
@RyanGibb derp, I totally missed the opportunity to state what I saw the collision with.  In both key maps, "cc" (or perhaps better visualized as `c c` is bound to `mu4e-compose-new`.  The first `c` is where the collision occurs.

Since this is about searching, and we already have `s` bound to `mu4e-search`, perhaps we could have `s` become the prefix.  `s s` would be `mu4e-search` and `s q` would be `mu4e-search-query`.  I'm not sure what the difference is between them, and I'm presently phone-bound.

If that sounds amenable to you, I'd be happy to change my PR from a revert to affect the suggested prefix of the two. 

Also of note:  I'm not sure if this would qualify as a "breaking" change for evil-collection, and what should be done to communicate it.  Though having to press one more `s` when we are probably all running around with `which-key` seems fairly innocent in my eyes.

LoganBarnett:
This link should highlight one of the mapping collisions: https://github.com/emacs-evil/evil-collection/pull/860/files#diff-ca4726d29c4a569cbc67fe03272b2342c0298b29daf39eb15c6800dfde5a9adeR160

RyanGibb:
> @RyanGibb derp, I totally missed the opportunity to state what I saw the collision with. In both key maps, "cc" (or perhaps better visualized as `c c` is bound to `mu4e-compose-new`. The first `c` is where the collision occurs.

Right! Strange that I don't get this error when loading the package, but I guess I never noticed this because I use `C` to compose a new mail.

> Since this is about searching, and we already have `s` bound to `mu4e-search`, perhaps we could have `s` become the prefix. `s s` would be `mu4e-search` and `s q` would be `mu4e-search-query`. I'm not sure what the difference is between them, and I'm presently phone-bound.

Given that upstream binds `c` to `mu4e-search-query` and we follow every other binding on the main screen, I wonder if dropping the `cc` binding might be the better solution.

mu4e's [changelog](https://github.com/djcb/mu/blob/master/NEWS.org#other) describes the command as,
```
    - New command ~mu4e-search-query~ (bound to =c=) which lets you pick a query
      (from bookmark / maildir shortcuts) with completion in main / headers /
      view buffers.
```

I should note I don't have commit access to this repository though! I'm just an interested observer.

condy0919:
> Given that upstream binds `c` to `mu4e-search-query` and we follow every other binding on the main screen, I wonder if dropping the `cc` binding might be the better solution.

Yep.

PR #862, opened 2025-04-04
mu4e: add 'c' binding to mu4e-search-query and remove 'cc' binding

RyanGibb:
mu4e-search-query was introduced in mu4e v1.12.0:
- New command ~mu4e-search-query~ (bound to =c=) which lets you pick a query (from bookmark / maildir shortcuts) with completion in main / headers / view buffers.

Bindings are added for the main and headers view.

Upstream binds c to mu4e-search-query, and we follow every other binding on the main screen, so we drop the colliding cc binding

See https://github.com/emacs-evil/evil-collection/pull/844, and https://github.com/emacs-evil/evil-collection/pull/860.

### DIFF
--- a/modes/mu4e/evil-collection-mu4e.el
+++ b/modes/mu4e/evil-collection-mu4e.el
@@ -29,12 +29,12 @@
 ;; keybindings are defined:
 ;;
 ;; General commands:
-;; | Commmand                 | evil-mu4e | Alternative |
-;; |--------------------------+-----------+-------------|
-;; | Jump to maildir          | J         |             |
-;; | Update                   | u         |             |
-;; | Compose message          | cc        | C           |
-;; | Kill update mail process | x         |             |
+;; | Commmand                 | evil-mu4e |
+;; |--------------------------+-----------|
+;; | Jump to maildir          | J         |
+;; | Update                   | u         |
+;; | Compose message          | C         |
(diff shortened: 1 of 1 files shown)


LoganBarnett:
@RyanGibb doesn't this have the same problem in that we have the other prefixes for things like `c e` (compose-edit), `c r` (compose-reply), etc?  We'd have to push for those alternative keybindings as well.

IMO it's kind of a step back - I liken compose to magit's commit.  I hit `c` and if I want to do the default thing (just make a new message), I hit `c` again.  But for the compose alternatives I can see those via `which-key`.  That said, my opinion doesn't mean much, and I would indeed mean retaining a deviation from `mu4e`'s default bindings.

I can always override back to the old behavior, and make `mu4e-search-query` something I can remember (maybe like `Q` or `S`).  So this isn't terribly a big deal to me :)

RyanGibb:
Hmm, that's a good point. What if we bound C to mu4e-search-query?

jojojames:
C is fine if there are no current conflicts

jojojames:
Feel free to reopen if there's a change.

RyanGibb:
I can't seem to reopen but I've opened a PR binding it to `S` here (since `C` collides): https://github.com/emacs-evil/evil-collection/pull/902

PR #873, opened 2025-09-21
feat(reftex): Add binding for accepting selection

jessy836:
Add a new key binding for the "RET" key to accept the current selection in reftex-select-mode. This improves usability by allowing users to confirm their selection easily.


condy0919:
thanks

PR #877, opened 2025-09-28
Fix evil-shift-width for yaml-ts-mode

rynffoll:
Fix setting `evil-shift-width` for `yaml-ts-mode`

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -350,7 +350,7 @@ See `evil-collection-init' and `evil-collection--modes-with-delayed-setup'."
     woman
     xref
     xwidget
-    yaml-mode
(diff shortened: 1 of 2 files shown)


tshu-w:
@condy0919 The `yaml-ts-mode` doesn't have `yaml-indent-offset` by default, which causes 
```
File mode specification error: (void-variable yaml-indent-offset)
```

condy0919:
oops.

condy0919:
Reverted. @rynffoll There is no `yaml-indent-offset` defined in yaml-ts-mode bundled with Emacs. Any thoughts?

rynffoll:
Maybe we could define our own variable if `yaml-ts-mode` doesn't provide any alternatives?

PR #890, opened 2026-02-04
magit: Fix :exclude-newline rewriting

real-or-random:
Only rewrite the EXCLUDE-NEWLINE argument when
evil-collection-magit-in-visual-pre-command is t. Previously the code always rewrote the argument to nil even when evil-collection-magit-in-visual-pre-command was not bound.


real-or-random:
- `(car arglist)` is the original `exclude-newline` arg (it's the first arg of the advised function). The `or` call computes the new `exclude-newline` arg.  
 - `(or (car arglist) ...` ensures that if the original arg is set to non-nil already, it will stay untouched.
 - If the original arg is nil, then we check whether we want to override it to t. This is what the `(and ...)` does.
 - `cons` is used to combine the computed new arg with all further args unmodified. (Currently there are no further args.)

PR #895, opened 2026-03-02
corfu: Add "<return>" to corfu-map when magic-return is enabled

jamescherti:
This ensures the completion candidate is inserted when pressing the Enter key in GUI frames. Previously, only RET was bound, which could cause the key to fall through to the default minibuffer or evil-insert-state binding.


condy0919:
As far as I remember, `<return>` seems to only be triggered in terminal Emacs?

PR #902, opened 2026-04-17
mu4e: add 'Q' binding to mu4e-search-query

RyanGibb:
mu4e-search-query was introduced in mu4e v1.12.0

bindings are added for the main and headers view


modes/mu4e/evil-collection-mu4e.el
jojojames, line 187: "S" here

RyanGibb: Darn, I changed it to "Q" as suggested by @LoganBarnett which I don't think collides?

jojojames: Works for me

PR #903, opened 2026-04-27
Apply the key blacklist to bindings with conflicting prefixes

Stebalien:
This change prevents evil-collection from binding a key that's a strict prefix or extension of a blacklisted key. E.g. Without this, evil-collection can end up binding a key that conflicts with a blacklisted key.

For example, by blacklisting "ab", evil-collection will be unable to bind any of "a", "ab", "abc", etc. However, it will be able to bind "ac", "acd", etc.

**Motivation:** I ran across this because I bind `;` _globally_ as my universal prefix (`C-u`) and add it to `evil-collection-key-blacklist` to prevent evil-collection from conflicting with it. Unfortunately, `evil-collection-dired` still binds `;d`, etc. in `dired-mode-map` which ends up overriding this binding.

### DIFF
--- a/evil-collection.el
+++ b/evil-collection.el
@@ -534,6 +534,26 @@ means all states for `evil-define-key', return nil."
        states)
      evil-collection-state-denylist)))
 
+(defun evil-collection--keys-conflict (key-a key-b)
+  "Return t if KEY-A and KEY-B conflict.
+Keys conflict if they're equal or if one is a prefix of the other."
+  (setq key-a (vconcat key-a)
+        key-b (vconcat key-b))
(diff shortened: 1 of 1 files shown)


evil-collection.el
jojojames, line 582: wonder if we should add it here and then call evil-collection-can-bind-key above

Stebalien: Ah, I missed that case. IMO, we don't want to call `evil-collection-can-bind-key` each time for performance reasons (lots of calls to `kbd`) but I can definitely deduplicate the logic.

PR #910, opened 2026-05-14
eca: Add binding to allow submitting from insert mode

magthe:
### Brief summary of what the change does

Bind C-<return> to submit in eca-mode so one doesn't need to leave insert mode quite as much.

### Direct link to the related package repository

https://github.com/editor-code-assistant/eca-emacs

### Checklist

- [x] byte-compiles cleanly

[eca] Add binding for sending prompt in insert state

### DIFF
--- a/modes/eca/evil-collection-eca.el
+++ b/modes/eca/evil-collection-eca.el
@@ -45,7 +45,8 @@
       (kbd "RET") 'eca-chat--key-pressed-return)
     (evil-collection-define-key newline-state 'eca-chat-mode-map
       (kbd "<return>") 'eca-chat--key-pressed-newline
-      (kbd "RET") 'eca-chat--key-pressed-newline))
+      (kbd "RET") 'eca-chat--key-pressed-newline
+      (kbd "C-<return>") 'eca-chat--key-pressed-return))
 
   (evil-collection-define-key 'insert 'eca-chat-mode-map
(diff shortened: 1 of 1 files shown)


magthe:
This is a reaction to https://github.com/editor-code-assistant/eca-emacs/issues/242.

jojojames:
Let me think about this one first, the last thing I want to do is bind yet another key to another core key-bind and get more bug reports or discussions on the "right" default.

Is setting the default to 'insert insufficient for you? Or you like the current default (e.g. RET in normal), but want a way to submit in insert too? That might inform what the right thing to do is.

magthe:
> Is setting the default to 'insert insufficient for you? Or you like the current default (e.g. RET in normal), but want a way to submit in insert too? That might inform what the right thing to do is.

The latter, I want a way to submit in `insert` too.  But maybe changing `evil-collection-repl-submit-state` is the solution to my problem.

jojojames:
Gotcha, C-return seems ok at first glance but let me double check other modes and what eca has bound to c-return, maybe we can move what's bound there to c-return in normal state if there's an existing binding.

magthe:
I'll try out setting `evil-collection-repl-submit-state` to `'insert` for a while, and close this PR.

jojojames:
Cool feel free to reopen later if you have more thoughts on the topic.

PR #912, opened 2026-05-14
gptel: Bind S-<return> as promised by evil-collection-repl-submit-state

magthe:
### Brief summary of what the changes does

The setup for gptel uses `'evil-collection-repl-submit-state', but doesn't bind 'S-<return>' to 'newline' at all, not in "submit state" nor in "newline state". This change addresses that.

[gptel] Bind S-<return> according to the submit-state promise

The documentation of `evil-collection-repl-submit-state` ends within

    S-RET always inserts a newline regardless of state.

This change makes that true for gptel-mode.


magthe:
I did notice that there are several other modes that use `'evil-collection-repl-submit-state` but don't bind `S-<return>`, e.g. `comint-mode` and `slime-mode`. However I don't dare make the same change for them as I don't use them and therefore don't know if that binding is used already.

PR #921, opened 2026-06-03
eat: bind p/P to eat-yank in normal state

jam1015:
Brief summary of what the package does

  eat (Emulate A Terminal) is a pure-Elisp terminal emulator with first-class eshell integration. This PR doesn't add a new mode —
  it extends the existing evil-collection-eat so that p and P in normal state send the kill-ring into the terminal via eat-yank,
  matching what evil-collection-term and evil-collection-vterm already do.

  Direct link to the package repository

  https://codeberg.org/akib/emacs-eat

  Checklist

  - byte-compiles cleanly
  - M-x checkdoc is happy
  - evil-collection-eat-setup is defined with defun (pre-existing; extended here)
  - evil-collection-eat-maps is defined with defconst (pre-existing; extended here with the four input sub-mode maps)
  - All functions start with evil-collection-eat- (no new functions added — the new bindings call eat-yank directly)

  ---
  Description

Written with help of AI, I understand if its not ok, but I've tested it

  Brings eat in line with term and vterm. evil-collection-term binds p to term-paste (line 149), and evil-collection-vterm binds p
  and P to send the kill-ring through the terminal (lines 283-284). eat had no equivalent — p in normal state just fell through to
  evil-paste-after which doesn't do anything sensible in a terminal buffer.

  Both p and P go to eat-yank. In a terminal there's no meaningful "before vs after" since the running program controls the cursor,
   so splitting them would be cosmetic at best. Binding both stops them falling through to evil-paste-before/-after.

  Insert state isn't touched — p self-inserts there, and C-y is already routed to eat-self-input by the existing setup.

  The bindings are placed on the four sub-mode input maps (eat-semi-char-mode-map, eat-char-mode-map,
  eat-eshell-semi-char-mode-map, eat-eshell-char-mode-map) rather than on eat-mode-map. Those maps are active when the user is
  actually interacting with a running terminal, which is when "send the kill-ring as terminal input" is the right semantics.
  eat-line-mode-map is deliberately excluded so line mode keeps evil's default paste — it's a buffer-edit mode, not a
  terminal-interaction mode. The four maps are also added to evil-collection-eat-maps for visibility to consumers of
  evil-collection-setup-hook.

  The defvars for the new map symbols aren't strictly required (the symbols only appear inside quoted lists, so the byte-compiler
  doesn't warn), but I added them to match the existing (defvar eat-line-mode-map) and keep other linters happy.


jojojames:
@condy0919 Do you use eat? I don't use it so not sure if this is the right semantics. Looks ok though.

> Written with help of AI, I understand if its not ok, but I've tested it

No issue with it, in fact, I think this package is a good one to use AI to add more modes automatically.

Feel free to merge @condy0919 .

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
+name: Sync Upstream
```
jojojames: What is this for?

condy0919:
> @condy0919 Do you use eat? I don't use it so not sure if this is the right semantics. Looks ok though.
> 
> > Written with help of AI, I understand if its not ok, but I've tested it
> 
> No issue with it, in fact, I think this package is a good one to use AI to add more modes automatically.
> 
> Feel free to merge @condy0919 .

The keybindings seem fine, but why added a new GitHub action?

condy0919:
I thought the `main.yaml` is for you only. For gh:evil/evil-collection, no need to sync from itself. So drop the `main.yaml`.

.github/workflows/main.yml
jam1015: oh no sorry this shouldn't be here, in the future, its for my own use, in the future i'll always make PRs from dedicated branches

jam1015:
so sorry about my workflow file polluting this PR, deleted it, will always use a feature  branch for PRs in the future

jam1015:
starting new PR on feature branch

PR #933, opened 2026-06-22
telega: Fix void binding and unshadow leaked native button keys

0WD0:
`telega-msg-pin-toggle` was renamed to the transient prefix `telega-transient-msg-pin-toggle` upstream, so "P" on a message button raised a void-function error.  Bind the transient prefix instead.

Also unbind native keys that leak through button text-property keymaps, which take precedence over Evil state maps and therefore fire instead of the intended g/Z/ds-prefixed bindings:

- user "i" (telega-describe-user)           -> use "g?"
- chat "h" (telega-describe-chat)           -> use "g?"
- msg "B" (telega-msg-ban-sender)           -> use "ds"
- msg "L" (telega-msg-redisplay)            -> use "ZL"
- msg "T" (telega-msg-open-thread-or-topic) -> use "gr"

### DIFF
--- a/modes/telega/evil-collection-telega.el
+++ b/modes/telega/evil-collection-telega.el
@@ -142,6 +142,7 @@
     "B" nil
     "K" nil
     (kbd "DEL") nil
+    "i" nil
 
     "g?" 'telega-describe-user
     "m" 'telega-user-chat-with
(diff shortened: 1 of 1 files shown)


jojojames:
Is it worth adding a backwards compatible check for the old function (e.g. fboundp for new/old function). Up to you.

Otherwise, can you rebase. I made a lot of changes recently.

PR #937, opened 2026-07-10
Add bindings for `agent-shell-manager`

juboba:

Evil bindings for `agent-shell-manager`, a `tabulated-list` buffer manager for `agent-shell`. Adds mapping for `RET` to visit entries and all manager actions onto convenient keys.

https://github.com/jethrokuan/agent-shell-manager

### Checklist

- [X] byte-compiles cleanly
- [X] `M-x checkdoc` is happy. Don't manually write `(provide 'evil-collection-mpc)`, `M-x checkdoc` can do it automatically for you
- [X] define `evil-collection-mpc-setup` with `defun`
- [X] define `evil-collection-mpc-mode-maps` with `defconst`
- [X] All functions should start with `evil-collection-mpc-`

feat(agent-shell-manager): Add evil bindings for agent-shell-manager

### DIFF
--- a/modes/agent-shell-manager/evil-collection-agent-shell-manager.el
+++ b/modes/agent-shell-manager/evil-collection-agent-shell-manager.el
@@ -70,7 +70,7 @@
       "M" 'agent-shell-manager-set-model
       "x" 'agent-shell-manager-interrupt
       "t" 'agent-shell-manager-view-traffic
-      "l" 'agent-shell-manager-toggle-logging)))
+      "L" 'agent-shell-manager-toggle-logging)))
 
 (provide 'evil-collection-agent-shell-manager)
 ;;; evil-collection-agent-shell-manager.el ends here
--- a/modes/agent-shell-manager/evil-collection-agent-shell-manager.el
+++ b/modes/agent-shell-manager/evil-collection-agent-shell-manager.el
@@ -1,0 +1,75 @@
+;;; evil-collection-agent-shell-manager.el --- Bindings for agent-shell-manager -*- lexical-binding: t -*-
+
+;; Copyright (C) 2026 Julio Borja Barra
+
+;; Author: Julio Borja Barra
+;; Maintainer: Julio Borja Barra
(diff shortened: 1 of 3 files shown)


condy0919:
Personally, I prefer key bindings that associate actions with uppercase letters. For example, "K" corresponds to `agent-shell-manager-kill`. BTW, you shouldn't bind `l`.

modes/agent-shell-manager/evil-collection-agent-shell-manager.el
```diff
@@ -0,0 +1,75 @@
...
+        "gm" 'agent-shell-manager-set-mode
+        "gv" 'agent-shell-manager-set-model
+        "gx" 'agent-shell-manager-interrupt
+        "gt" 'agent-shell-manager-view-traffic
+        "gl" 'agent-shell-manager-toggle-logging)
+    (evil-collection-define-key 'normal 'agent-shell-manager-mode-map
+      "K" 'agent-shell-manager-kill
+      "c" 'agent-shell-manager-new
+      "R" 'agent-shell-manager-restart
+      "d" 'agent-shell-manager-delete-killed
```
jojojames: (delete :enabled t :state normal :key "d")
    (delete-2 :enabled t :state normal :key "x")

Should we use?

juboba: thanks @jojojames I made the change for "d", but "x" is interrupt which is semantically different from "delete-2", right?

juboba:
@condy0919 changed to "L" (instead of "l"), but I don't see a problem with the rest, considering this is a read-only buffer.

modes/agent-shell-manager/evil-collection-agent-shell-manager.el
jojojames: Yeah I wonder if we should have an interrupt type binding mapped to x or something similar. (probably the c-c c-c should go to one.) Will defer on that topic for this pr though.

PR #944, opened 2026-09-05
Add configuration for grep-edit-mode

hiecaq:

grep-edit-mode is Emacs 31's built-in alternative to wgrep.  Wgrep is more feature-complete,  so I bind it as a fallback when wgrep is not available.

https://github.com/emacs-mirror/emacs/blob/emacs-31/lisp/progmodes/grep.el#L1125

### Checklist

- [x] byte-compiles cleanly

### DIFF
--- a/modes/grep/evil-collection-grep.el
+++ b/modes/grep/evil-collection-grep.el
@@ -30,7 +30,10 @@
 (require 'evil-collection)
 (require 'grep)
 
-(defconst evil-collection-grep-maps '(grep-mode-map))
+(defvar grep-edit-mode-map) ;; might be missing if it's older Emacs
+
+(defconst evil-collection-grep-maps '(grep-mode-map
+                                      grep-edit-mode-map))
 
 ;;;###autoload
 (defun evil-collection-grep-setup ()
@@ -43,7 +46,13 @@
   ;; `wgrep' integration
   (when (fboundp 'wgrep-setup)
     (evil-collection-define-key 'normal 'grep-mode-map
-      "i" 'wgrep-change-to-wgrep-mode)))
+      "i" 'wgrep-change-to-wgrep-mode))
+
+  (when (>= emacs-major-version 31)
+    (evil-collection-define-key 'normal 'grep-mode-map
+      "I" 'grep-change-to-grep-edit-mode)
+    (evil-collection-bind 'grep-edit-mode-map
+                          'quit-save 'grep-edit-save-changes)))
 
 (provide 'evil-collection-grep)
 ;;; evil-collection-grep.el ends here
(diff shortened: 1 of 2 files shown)


modes/grep/evil-collection-grep.el
```diff
@@ -40,10 +43,17 @@
     "\C-j" 'next-error-no-select
     "\C-k" 'previous-error-no-select)
 
-  ;; `wgrep' integration
-  (when (fboundp 'wgrep-setup)
+  (cond
```
jojojames, line 46: Is it possible to use both wgrep and grep-edit-mode in the same emacs session?

hiecaq: Technically yes, they implement a similar grep-mode <-> editable grep-mode behavior.  But I believe the reason most people installed wgrep is for this editing grep result feature, which precedes grep-edit-mode by far (as it implements atomic updates,  could be reverted, etc).  The most benefit of grep-edit-mode is that it's built-in, so I consider it as a fallback as for now.

jojojames: Lets split them then. I use wgrep most with dired, not grep.

condy0919: > Lets split them then. I use wgrep most with dired, not grep.

Maybe you're using `wdired-mode`

jojojames: Yeah, that's what I was thinking of. These should still be split though.

hiecaq: I re-pushed a commit that binds the `grep-change-to-grep-edit-mode` into `I`, so that it doesn't conflict with wgrep.

jojojames: I took a closer look, maybe 1. we should a defcustom and let the user pick which one they prefer (we can use the fallback logic for the case where one doesn't exist as the fallback to the defcustom)

or 2. we can keep it simple and leave it as is on "I", or 3. we can fallback to your initial change of keeping both on "i" and having one override the other.

Up to you, I don't have a strong opinion on any of the 3 options. The reason I bring it up is I'm guessing it's annoying to have to remember "I" for grep-edit versus the clearer "i" on wgrep.

hiecaq: Maybe let's wait for another week (or two?) to see if people have more comments on this.

I personally prefer option 1.

jojojames: Sounds good, re-ping when you decide.

paniash: @hiecaq Thanks for the PR. I personally prefer option 1 as well.

